PR #10, opened 2017-03-09
Add header-only INTERFACE library if no source files

davidbeckingsale:


davidbeckingsale:
@gzagaris does this address #5 for you?

gzagaris:
@davidbeckingsale thanks for rapidly addressing this. I am using an older version of BLT and I had to backport your fix to try it out. I wasn't able to get it to work due to the way we were dealing with things like MPI etc. in the old version. For example, once the `add_library( foo INTERFACE )` is added we were making calls to target_include_directories() and for an INTERFACE library, we would have to propagate the INTERFACE property to the downstream calls.

Your changes seem reasonable to me. I'll have to hold on a bit to try it out though until we move to a newer version of BLT, which I suspect will happen soon.

cyrush:
using INTERFACE seems like the right way to go assuming we can iron out the kinks. It looks like we need to provide abs paths.

davidbeckingsale:
Yup, I just noticed that, I think I can fix it 😄

PR #22, opened 2017-03-27
rework treatment of extra compiler flags

cyrush:
Use BLT_<LANG>_FLAGS{\_<DEBUG|RELEASE>}

### DIFF
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -134,6 +134,20 @@ if(ENABLE_FORTRAN AND BLT_FORTRAN_FLAGS)
     set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${BLT_FORTRAN_FLAGS}")
 endif()
 
+################################################
+# Support Extra Flags for the CUDA compiler.
+# N.B. must be ; delimited.
+################################################
+if(ENABLE_CUDA AND BLT_CUDA_FLAGS)
+    set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS};${BLT_CUDA_FLAGS}")
+    # Replace spaces with semicolons in the CUDA flags
+    foreach(flag_var CUDA_NVCC_FLAGS)
+      if(${flag_var} MATCHES " ")
+        string(REGEX REPLACE " " ";" ${flag_var} "${${flag_var}}")
+      endif(${flag_var} MATCHES " ")
+    endforeach(flag_var)
+endif()
+
 
 ###############################################################
 # Support Extra Flags based on CMake Config Type
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -96,6 +96,8 @@ endif()
 # Support for extra compiler flags and defines
 ################################################
 
+message(STATUS "Adding optional BLT definitions and compiler flags")
+
 ####################################################
 # create relocatable static libs by default
 ####################################################
@@ -105,7 +107,8 @@ set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 # Support Extra Definitions for all targets
 ##############################################
 if(BLT_DEFINES)
-    add_defintions(${BLT_DEFINES})
+    add_definitions(${BLT_DEFINES})
+    message(STATUS "Added \"${BLT_DEFINES}\" to definitions")
 endif()
 
 ##########################################
@@ -118,6 +121,7 @@ endif()
 ##########################################
 if(BLT_C_FLAGS)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${BLT_C_FLAGS}")
+    message(STATUS "Updated CMAKE_C_FLAGS to \"${CMAKE_C_FLAGS}\"")
 endif()
 
 #############################################
@@ -125,6 +129,7 @@ endif()
 #############################################
 if(BLT_CXX_FLAGS)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${BLT_CXX_FLAGS}")
+     message(STATUS "Updated CMAKE_CXX_FLAGS to \"${CMAKE_CXX_FLAGS}\"")
 endif()
 
 ################################################
@@ -132,6 +137,7 @@ endif()
 ################################################
 if(ENABLE_FORTRAN AND BLT_FORTRAN_FLAGS)
     set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${BLT_FORTRAN_FLAGS}")
+     message(STATUS "Updated CMAKE_Fortran_FLAGS to \"${CMAKE_Fortran_FLAGS}\"")
 endif()
 
 ################################################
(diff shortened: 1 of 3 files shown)


davidbeckingsale:
:+1: looks good to me!

cyrush:
One subtle issue folks may miss:

CMake uses "Fortran" vs "FORTRAN for its settings vars used to configure Fortran (including compiler flags)

For example:
`CMAKE_Fotran_FLAGS` vs `CMAKE_FORTRAN_FLAGS`

For BLT, I think its good to be consistent with using all caps for our variable settings, so I used:

`BLT_FORTRAN_FLAGS`
`BLT_FORTRAN_FLAGS_DEBUG`
`BLT_FORTRAN_FLAGS_RELEASE`

davidbeckingsale:
I like that, however, people still need to remember to enable `Fortran` as a language, rather than all caps?

cyrush:
@davidbeckingsale  we have "ENABLE_FORTRAN" as our option, however they do need to set CMAKE_Fortran_COMPILER ...

davidbeckingsale:
That's fine, I think. I was more getting at using `project(blah LANGUAGES Fortran)` to tell CMake that the project needs Fortran. Do you think it's worth adding the corresponding `enable_language` call in SetupCompilerOptions in case the user hasn't done it already?

cyrush:
Yes, I'll add something like this:

```
# Invoke CMake Fortran setup
# if ENABLE_FORTRAN == ON
if(ENABLE_FORTRAN)
    enable_language(Fortran)
endif()
```

I thought it was already there, but I was confusing with another cmake setup.

cyrush:
Hadn't considered it, but purpose of ENABLE_FORTRAN allows us to optionally enable fortran -- so we can guard fortran features when the compiler doesn't exist.

If you have a fortran only project, I think you can avoid C and CXX with your own project statement:

`project(myname Fortran)`

However you will sill have to set `ENABLE_FORTRAN = ON` to trigger the right blt paths. 

Don't know of a use case where we optionally want C and CXX? Maybe we can revisit ...

cmake/SetupCompilerOptions.cmake
```diff
@@ -97,49 +97,130 @@ endif()
...
-        add_compile_options("${EXTRA_C_FLAGS_RELEASE}")
     endif()
 
-    #############################################
-    # Support Extra Flags for the C++ compiler.
-    #############################################
-    if(EXTRA_CXX_FLAGS)
-        add_compile_options("${EXTRA_CXX_FLAGS}")
+    # Extra Flags for release builds
+    if(CMAKE_BUILD_TYPE MATCHES RELEASE)
```
kennyweiss: Does capitalization matter here?  I think the build types are ``Debug`` and ``Release``  (note the capitalization)

cmake/SetupCompilerOptions.cmake
```diff
@@ -97,49 +97,130 @@ endif()
...
-      )
-        add_compile_options("${EXTRA_CXX_FLAGS_DEBUG}")
-    endif()
-
-    # Extra Flags for the release builds with the C++ compiler.
-    if(EXTRA_CXX_FLAGS_RELEASE AND CMAKE_BUILD_TYPE MATCHES RELEASE)
-        add_compile_options("${EXTRA_CXX_FLAGS_RELEASE}")
+    
+    # Extra Flags for release w/ debug info builds
+    if(CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
```
kennyweiss: Is this a standard thing to assume that ``RelWithDebInfo`` builds would have the flags from ``Release`` and form ``Debug``?

kennyweiss:
We were just dealing with a situation recently, where, depending on the platform, we might need some additional linker flags.

I think we should also have user-supplied ``BLT_LINKER_FLAGS`` -- I would think that these need to be language dependent (e.g. ``BLT_C_LINKER_FLAGS``, ``BLT_CXX_LINKER_FLAGS``, ... ).


Should this be part of the same PR, or in a new PR?

cmake/SetupCompilerOptions.cmake
cyrush: This follows what we were doing before, there wasn't much vetting of the prior solution -- so we can discuss alternatives (ex: req sep flags for RelWithDebug)

cmake/SetupCompilerOptions.cmake
cyrush: good catch, old code was likely wrong -- we need to fix this.

cmake/SetupCompilerOptions.cmake
davidbeckingsale: I see (as a user) `RelWithDebInfo` as meaning the `Release` flags, plus whatever is needed to generate information for debuggers (e.g. `-g` for a host compiler, `-G` for CUDA).

Combining the `Release` and `Debug` flags might have consequences a user wouldn't expect.

I would propose that we set the flags to `Release` + `-g` by default, and have a new variable so that users can add any extra flags they want.

cyrush: two things:
- sounds like we are being too clever, should we just provide paths for the std cmake options 
- I think Release + -g is handled already by simply setting RelWithDebInfo, these things are for extra flags.

cyrush:
@kennyweiss  can we tackle linker stuff in a new pr?

cmake/SetupCompilerOptions.cmake
kennyweiss: Yes -- ``RelWithDebInfo`` adds in the appropriate debug symbols flag (e.g. ``-g``) for you.

I think the ``RelWithDebInfo`` flags should be explicitly set when necessary in the same manner as the ``Release`` and ``Debug`` flags.

kennyweiss:
Sure -- let's handle the linker flags separately.

davidbeckingsale:
Do we need the conditionals checking for the current build type for every set of flags?

I think we could always add `BLT_C_FLAGS_DEBUG` to `CMAKE_C_FLAGS_DEBUG` (for example) regardless of build type, since they will only be used based on the type of the current build.

cyrush:
@davidbeckingsale . yes -- good catch, we only needed those when we were doing logic to combine debug and release flags for the RelWithDebInfo case. 

I however keep the guard for (NOT CMAKE_CONFIGURATION_TYPES) to avoid doing this logic in visual studio

davidbeckingsale:
Sounds good!

cmake/SetupCompilerOptions.cmake
```diff
@@ -92,54 +92,145 @@ elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "Intel")
...
-    ##########################################
-    if(EXTRA_C_FLAGS)
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${EXTRA_C_FLAGS}")
+
+    ####################################
+    # Extra Flags for debug builds 
+    ####################################
+    
+    # debug flags for the C compiler
+    if(BLT_C_FLAGS_DEBUG)
```
kennyweiss: Could we clean this up with a loop over the different build types?
e.g. 
```make
set(bltTypes DEBUG RELEASE RELWITHDEBINFO ...)

foreach(bld in ${bldTypes)
    if(BLT_C_FLAGS_${bld})
       set(CMAKE_C_FLAGS_${bld} "${CMAKE_C_FLAGS_${bld}} ${BLT_C_FLAGS_${bld}}")
    endif()

    if(BLT_CXX_FLAGS_${bld})
       set(CMAKE_CXX_FLAGS_${bld} "${CMAKE_CXX_FLAGS_${bld}} ${BLT_CXX_FLAGS_${bld}}")
    endif()

   if(ENABLE_FORTRAN AND BLT_FORTRAN_FLAGS_${bld})
     set(CMAKE_Fortran_FLAGS_${bld}  "${CMAKE_Fortran_FLAGS_${bld}} ${BLT_FORTRAN_FLAGS_${bld}}")
  endif()
endforeach()
```

cyrush: It would be shorter ... we could loop over cfg_types

cyrush:
@kennyweiss  updated to use a loop over the types as you suggested

cyrush:
@kennyweiss @davidbeckingsale  can you give a quick review again? I tested the paths a bit manually, but another set of eyes would be good.

kennyweiss:
Looks good (based on a visual inspection)

davidbeckingsale:
Yeah, looks good to me!

cyrush:
thanks!

Issue #12, opened 2017-03-09
non-existent include dir warning from fortran


cyrush:
When testing blt, the "include" dir is never created, yet we add it to the target_includes.
gcc and clang don't seem to care, but gfortran issues a warning:

```
f951: Warning: Nonexistent include directory ‘blt_work/github/build-debug/include’ [-Wmissing-include-dirs]
[ 68%] Linking Fortran static library ../../../lib/libfruit.a

```
 

Here is where we set it:
```
# Defines the layout of the build directory. Namely,
# it indicates the location where the various header files should go,
# where to store libraries (static or shared), the location of the
# bin directory for all executables and the location for fortran modules.

# Set the path where all the headers will be stored
set(HEADER_INCLUDES_DIRECTORY
    ${PROJECT_BINARY_DIR}/include/
    CACHE PATH
    "Directory where all headers will go in the build tree"
    )
include_directories(${HEADER_INCLUDES_DIRECTORY})
```

This is happening b/c we have a bare bones project, but I think if folks aren't using _copy_headers_ this could also happen.  It's very unlikely that this will happen in user code that uses _copy_headers_.

Note: _copy_headers_ is off by default.

white238:
This is now guarded with ENABLE_COPY_HEADERS and is no longer an issue.

PR #23, opened 2017-03-29
Enable DEPENDS_ON for blt_register_library

white238:
* Add DEPENDS_ON to blt_register_library
* Stop a global include directory
* Turn off a variable if it's not being used

### DIFF
--- a/SetupBLT.cmake
+++ b/SetupBLT.cmake
@@ -94,12 +94,14 @@ include(${BLT_ROOT_DIR}/cmake/SetupCodeChecks.cmake)
 # bin directory for all executables and the location for fortran modules.
 
 # Set the path where all the headers will be stored
-set(HEADER_INCLUDES_DIRECTORY
-    ${PROJECT_BINARY_DIR}/include/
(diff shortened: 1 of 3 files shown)


SetupBLT.cmake
```diff
@@ -94,12 +94,14 @@ include(${BLT_ROOT_DIR}/cmake/SetupCodeChecks.cmake)
...
-    )
-include_directories(${HEADER_INCLUDES_DIRECTORY})
+if ( ENABLE_COPY_HEADERS )
+    set(HEADER_INCLUDES_DIRECTORY
+        ${PROJECT_BINARY_DIR}/include/
+        CACHE PATH
+        "Directory where all headers will go in the build tree"
+        )
+    include_directories(${HEADER_INCLUDES_DIRECTORY})
+endif()
```
cyrush: i think this helps with #12 -- doesn't completely eliminate the issue, but makes it very unlikely to be encountered outside of basic testing.

davidbeckingsale:
LGTM 👍

PR #28, opened 2017-04-11
Don't try and pass Fortran files to NVCC

davidbeckingsale:


davidbeckingsale:
@cyrush it looks like the original approach to regexing the Fortran files (in FindCUDA) got removed. This should fix it.

cyrush:
@davidbeckingsale -- so the regex was wrong, or setting the properties simply doesn't work?

davidbeckingsale:
I think it might be the properties, but I didn't debug it very well because I forgot that the REGEX was there, so I just added the check in FindCUDA. I think we could use the FindCUDA check in BLTPrivateMacros, but in some ways, it's "safer" to have it in FindCUDA since that's the lowest level.

cyrush:
Ok, just wanted to have a better understanding of what went wrong.

davidbeckingsale:
Yeah no worries, I'll check on it some more before I merge it.

kennyweiss:
Looks good.

Assuming that we intend to support Cuda Fortran files, would they get compiled another way?
Do we have a convention for naming them (e.g. *.cuf)?

white238:
Whats the status of this pull request? This bit my project and we came up with a hack around it but this seems like a better solution.

davidbeckingsale:
It can be safely merged, but I think that this functionality is somewhat duplicated in our handling of cuda as a dependency. My preference is probably to remove that and just handle it like we do here.

cmake/BLTPrivateMacros.cmake
```diff
@@ -216,7 +216,9 @@ macro(blt_setup_cuda_source_properties)
 
 
     foreach (_file ${arg_TARGET_SOURCES})
-        if (${_file} MATCHES "\\.(f|F)\\*")
+      get_source_file_property(_lang ${_file} LANGUAGE)
```
davidbeckingsale: This is the alternative location where we were trying to find Fortran files

@cyrush @white238 I think it's probably better to have this here and revert the FindCUDA changes...

cyrush: yes, that looks like a better approach!

robinson96: I think it's appropriate to exclude C in addition to Fortran here. CUDA is a C++ extension, and you'll likely get linking issues for mixed C and C++ code if you auto identify C files as CUDA files.

davidbeckingsale: Thanks @robinson96 - I'll extend the list, get find of the FindCUDA changes, then we can merge this and #82

cmake/BLTPrivateMacros.cmake
```diff
@@ -216,7 +216,9 @@ macro(blt_setup_cuda_source_properties)
...
+
+      if (${_lang} MATCHES Fortran OR ${_lang} MATCHES C)
```
davidbeckingsale: @robinson96 can you give this a shot?

cmake/BLTPrivateMacros.cmake
cyrush: To avoid surprises for C cases -- we could also try to change the CMake linker lang to CXX in the case that we are using CUDA and a we have a source file that CMake initially thinks requires C linkage.

We may still have a problem if the project doesn't specify CXX as a language.

jdahm: I just ran this over a set of C++ and Fortran files for SAMRAI, and the `LANGUAGE` property was very inconsistently set. For the majority of the files scanned, it looks like it sets `_lang` to `NOTFOUND`, even for "official" fortran and C++ extensions `.f` and `.C`, respectively and with C, CXX, and Fortran languages enabled. This was using CMake 3.9.2 on an LLNL machine. I'm investigating workarounds.

cyrush: @jdahm, is SAMRAI passing LANGUAGES CXX and Fortran as part of its project commands?

jdahm: Yes, it's using `project(SAMRAI C CXX Fortran)`

cyrush: @jdahm  thats a bummer, thats the only simple issue I can think of off the top of my head

zbeekman: @jdahm It sounds like a bug report to Kitware may be in order. A collaborator of mine has a support contract with them if LLNL doesn't, and we're looking at using BLT in a project together for NRC, so we may be able to justify commissioning a speedy bugfix. (Obviously it won't help for older versions of CMake though.)

@jdahm SAMRAI is up on GH, but I don't see any CMake files. If you are willing to share I'd be happy to take a look and submit a bug report to Kitware, or if you prefer, you can do it at https://gitlab.kitware.com/cmake/cmake/issues

jdahm: @zbeekman We're building in the feature/uvm branch of SAMRAI, but it requires a few fixes and dependencies, so it's not the easiest target for this. We fixed it there by manually setting the language property of the files.

I just verified and together with the workaround there this works, so I think it's OK to merge.

zbeekman: @jdahm Curious what the work around was. 

Just double checked the docs for LANGUAGE property:

> A property that __*can*__ be set to indicate what programming language the source file is. If it is not set the language is determined based on the file extension.

Sounds like this is *NOT* set by CMake most of the time, but you can set it yourself. @davidbeckingsale this makes me think that an alternate solution for the diff being discussed may be required.

white238:
This PR is no longer needed.

PR #30, opened 2017-04-21
Update Google Benchmark

davidbeckingsale:


cyrush:
yikes, it looks like we have a linking issue with gbenchmark + gbenchmark is adding a bunch of tests that we don't want added :-(

davidbeckingsale:
Yeah, I think the previous version included had some custom changes to the build. I'll try and work this out.

cyrush:
From the linker errors:
http://stackoverflow.com/questions/2418157/ubuntu-linux-c-error-undefined-reference-to-clock-gettime-and-clock-settim

It looks like new version of gbenchmark also depends on librt if we are on linux.

Also, shared vs static is always touchy on Windows.  We may need to force gbenchmark to mirror the build mode of BLT.

thirdparty_builtin/CMakeLists.txt
```diff
@@ -68,11 +68,16 @@ if(ENABLE_TESTS)
 
     if(ENABLE_BENCHMARKS)
         ## google benchmark support
-        add_subdirectory(gbenchmark)
+        add_subdirectory(gbenchmark-release-1.1.0)
+
+        if (NOT WIN32)
+          find_library(RT_LIBRARIES rt)
+          set(rt_libraries rt)
```
cyrush: should we use `RT_LIBRARIES` directly? It seems we are simply overriding to use the string "rt"

I don't think we need rt on osx, here is what I used when I needed for conduit :

https://github.com/LLNL/conduit/blob/master/src/libs/relay/CMakeLists.txt

davidbeckingsale:
@cyrush do you have any idea what's happening with the Windows build when `BUILD_SHARED_LIBS=On`. I don't have access to a Windows box locally so it's a pain to debug.

cyrush:
@davidbeckingsale I think we are close -- I can try to debug this weekend

cyrush:
For some reason on windows when we build shared -- benchmark.lib isn't generated (although the dll is created). I'll keep digging.

cyrush:
looked at the benchmark code, and I don't see any export statements in their interfaces -- so I don't think it is setup to work on windows for shared libs.  Missing .libs is a symptom when interfaces aren't exported:

https://cmake.org/pipermail/cmake/2009-March/027986.html

Adding the exports would be quite a bit of work --  for now I think we should to disable benchmark when shared is enabled on windows. We should also ping the benchmark folks.

davidbeckingsale:
Okay, that sounds good. Disabling benchmarks is a good idea. Should we do this in the appveyor config, or actually handle it inside BLT?

cyrush:
I think we should throw FATAL_ERROR in blt, if `BUILD_SHARED_LIBS` is on and `ENABLE_BENCHMARKS` is on for win32. 

Then in appyeyor, we can turn off `ENABLE_BENCHMARKS` for the shared case.

davidbeckingsale:
Okay, I'll do that.

cyrush:
Thanks, looks good!

davidbeckingsale:
This update should fix using Benchmark with Clang on PowerPC.

cyrush:
thanks for tackling this :-)

PR #32, opened 2017-06-15
Improve flags for better support of xlC and nvcc combinations

davidbeckingsale:
Ran into some issues building with xlC and nvcc - these changes fix that, by separating host and device flags, and adding support for C++ standards when xlC is used.

### DIFF
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -229,8 +229,18 @@ if( BLT_CXX_STD STREQUAL c++98 )
     set(CMAKE_CXX_STANDARD 98)
 elseif( BLT_CXX_STD STREQUAL c++11 )
     set(CMAKE_CXX_STANDARD 11)
+    if (COMPILER_FAMILY_IS_XL)
+      blt_append_custom_compiler_flag(
+        FLAGS_VAR CMAKE_CXX_FLAGS
+        DEFAULT "-std=c++11")
+    endif ()
 elseif( BLT_CXX_STD STREQUAL c++14 )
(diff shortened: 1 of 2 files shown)


cmake/SetupCompilerOptions.cmake
```diff
@@ -229,8 +229,18 @@ if( BLT_CXX_STD STREQUAL c++98 )
     set(CMAKE_CXX_STANDARD 98)
 elseif( BLT_CXX_STD STREQUAL c++11 )
     set(CMAKE_CXX_STANDARD 11)
+    if (COMPILER_FAMILY_IS_XL)
```
davidbeckingsale: I'm all for suggestions on a more universal way to handle this. I believe that we would run into this problem with Intel also on older CMake versions.

cyrush: This looks fine, it's clear you are adding flags for xl for the c++11 & c++14 case. 
If we need a fix for other cmakes thats ok as well, but our goal is to work with a few blessed versions of cmake -- currently 3.3.x and 3.8.x.

davidbeckingsale: I am running into this issue with Intel also. CMAKE_CXX_STANDARD does not add the flags. I've not checked this with 3.8.x yet

white238: You need to go to CMake 3.4 i think

davidbeckingsale: I'm using 3.4.1 - it's not supported. We have to work around this in RAJA also.

PR #37, opened 2017-07-05
Adds support for ensuring that all tests are invoked through mpiexec.

kennyweiss:
Some platforms, like IBM's BG/Q, require all executables to be run
on 'back end' compute nodes through a call to mpiexec.  This commit gives
the blt_add_test() macro the ability to ensure that all tests are wrapped with
a call to ${MPIEXEC} with at least one mpi rank.

To enable this feature, users must configure their code with
the cmake cache variable BLT_ALWAYS_WRAP_TESTS_WITH_MPIEXEC  set to TRUE.
E.g.
    set(BLT_ALWAYS_WRAP_TESTS_WITH_MPIEXEC TRUE CACHE BOOL
        "Ensures that tests will be invoked through mpiexec")

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -567,6 +567,12 @@ macro(blt_add_test)
     endif()
     set(test_command ${runtime_output_directory}/${arg_COMMAND} )
 
+    # If configuration option ENABLE_WRAP_ALL_TESTS_WITH_MPIEXEC is set, 
+    # ensure NUM_PROCS is at least one. This invokes the test through MPIEXEC.
+    if ( ENABLE_WRAP_ALL_TESTS_WITH_MPIEXEC AND NOT arg_NUM_PROCS )
+        set( arg_NUM_PROCS 1 )
+    endif()
+
     # Handle mpi
     if ( ${arg_NUM_PROCS} )
         set(test_command ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${arg_NUM_PROCS} ${test_command} )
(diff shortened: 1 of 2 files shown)


davidbeckingsale:
This is a nice solution. Should the `BLT_ALWAYS_WRAP_TESTS_WITH_MPIEXEC` variable be an option?

kennyweiss:
Thanks @davidbeckingsale -- That was my original idea, but it was deemed as an extra complication for users who would likely (almost) never need this feature.  We will have to document this as an advanced feature.

kennyweiss:
Note: this solution was discussed internally with help from @gzagaris and @cyrush

kennyweiss:
Note: This PR will resolve issue #37

kennyweiss:
@cyrush -- Just spoke with @gzagaris about this, and the ``BLT_ALWAYS_WRAP_TESTS_WITH_MPIEXEC`` should be an option in ``BLTOptions.cmake`` that is marked as advanced.

This will reduce the magic in our system, and allow users to see this in ``ccmake`` or ``cmake-gui``.   In the current system, it is a variable that cmake knows about that users have no way of seeing/verifying/modifying.

This is also more inline with @davidbeckingsale's intuition.

kennyweiss:
I updated the PR, and the option is called ``ENABLE_WRAP_ALL_TESTS_WITH_MPIEXEC`` and marked it as advanced.

This feels like the right solution here.  Now that it is an option, I was able to remove a lot of the awkwardness about how this was described in the Note on ``blt_add_test``, which I completely removed.

Issue #39, opened 2017-07-06
Add a patch to disable gtest death tests on unsupported platforms


kennyweiss:
gtest's death tests allows writing tests that verify that a program will exit under certain conditions, but they require forking and joining threads, and are not supported on all platforms.  gtest gets around through the ``{ASSERT,EXPECT}_DEATH_IF_SUPPORTED`` macros, which are no-ops when death tests are not supported.  They are controlled by the ``GTEST_HAS_DEATH_TEST`` compiler define.

Unfortunately, the code that determines whether death tests should be supported does not cover all platforms that we'd like to run code on, including IBM's BG/Q platform and the "Bash for Windows" environment. In addition, the platform setup code (in  gtest's``/include/gtest/internal/gtest-port.h``) does not allow one to override this variable.

To resolve this, we need to add a simple patch to only redefine the ``GTEST_HAS_DEATH_TEST`` compiler define if it is not already defined.

PR #40, opened 2017-07-06
Feature/patch gtest death test

kennyweiss:
Relates to issue #39


thirdparty_builtin/patches/gtest-1.8.0-override-GTEST_HAS_DEATH_TEST.patch
```diff
@@ -0,0 +1,39 @@
+From 84b7421d3c7f112c9501e564a86bd1bca4da4fc5 Mon Sep 17 00:00:00 2001
```
kennyweiss: The patch file is to help us track our patches to the third_party code.

kennyweiss:
It looks like appveyor's windows environment does not support death tests:

See line 1076 in https://ci.appveyor.com/project/cyrush/blt/build/1.0.99/job/dj3f1365yn72vq3m#L1076
```
4/6 Test #4: blt_gtest_death_test_smoke .......***Failed    0.03 sec
  Running main() from gtest_main.cc
  [==========] Running 1 test from 1 test case.
  [----------] Global test environment set-up.
  [----------] 1 test from blt_gtest_smoke
  [ RUN      ] blt_gtest_smoke.death_test
  Running main() from gtest_main.cc
C:\projects\blt-test\blt\tests\blt_gtest_death_test_smoke.cpp(57): error : Death test: ((void)0) [C:\projects\build\RUN_TESTS.vcxproj]
      Result: failed to die.
EXEC : error msg:  [C:\projects\build\RUN_TESTS.vcxproj]
  [  DEATH   ] 
  [  FAILED  ] blt_gtest_smoke.death_test (7 ms)
  [----------] 1 test from blt_gtest_smoke (7 ms total)
  
  [----------] Global test environment tear-down
  [==========] 1 test from 1 test case ran. (7 ms total)
  [  PASSED  ] 0 tests.
  [  FAILED  ] 1 test, listed below:
  [  FAILED  ] blt_gtest_smoke.death_test
  
   1 FAILED TEST
```

kennyweiss:
I should be able to fix the appveyor builds by adding the following to the configuration line (line 26 in ``appveyor.yml``)

```
cmake ...  -DBLT_C_FLAGS="-DGTEST_HAS_DEATH_TEST=0"  -DBLT_CXX_FLAGS="-DGTEST_HAS_DEATH_TEST=0"  ...
```

kennyweiss:
@white238 -- In my test configuration, I disabled death tests using a global ``BLT_CXX_FLAGS`` variable.  I.e. 
```
set(BLT_CXX_FLAGS -DGTEST_HAS_DEATH_TEST=0 CACHE PATH "")
```   
Now that we have the transitive dependencies working properly, is there a better way to only add the flags to the google test targets?

Note: We do not (currently) have a direct blt option for this, so it would have to be something that  can be set in the appropriate host-config file, or passed directly to cmake when configuring a build.

davidbeckingsale:
I think we should handle this better - if the death test is unsupported by the compiler, it should be disabled without having to pass a flag...

cyrush:
Since death tests have issues multiple platforms -- I think the most conservative solution is to disable gtest death tests by default, and have folks pass a flag if they want to opt into them.  

` -DGTEST_HAS_DEATH_TEST=1` 

We aren't sure what other platforms are going to have issues with death tests.  Out of the several hundred tests in the toolkit I think we have a handful of that are death tests. It would be better to not use death tests than to try to support for such a little payoff.

tests/CMakeLists.txt
```diff
@@ -55,6 +55,14 @@ blt_add_executable(NAME blt_gtest_smoke
...
 
+blt_add_executable(NAME blt_gtest_death_test_smoke
+                   SOURCES blt_gtest_death_test_smoke.cpp
+                   OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
+                   DEPENDS_ON gtest)
+
+blt_add_test( NAME blt_gtest_death_test_smoke
+              COMMAND blt_gtest_death_test_smoke)
+
 ################
```
cyrush: since this use case is a bit obscure, I don't think it rises to the level of its own blt smoke test.  If we want to test it should be folded into the gtest test (per haps if-defed, etc)

davidbeckingsale: Agree.

kennyweiss:
@cyrush and @davidbeckingsale -- based on your comments, and the way that gtest-port.h is set up, it sounds like we need a BLT option to enable death tests.

I will add ``BLT_ENABLE_DEATH_TESTS`` as an option that is off by default to ``BLTOptions.cmake`` and mark is as advanced.

cyrush:
@kennyweiss following our discussion -- I am on board with the option (and the one for the mpi stuff as well)

davidbeckingsale:
If we know which systems death tests definitely fail with, perhaps we can check for them if the option is enabled, and fail at CMake time with an error?

kennyweiss:
@davidbeckingsale -- I think the safest and simplest thing is to disable the death tests by default, and only enable them when we know they are valid.  This should lead to the fewest surprises.

gtest has some code to check for the platform, but the granularity is not high enough for us.  E.g. bg/q (and bash for windows) are identified as linux (where death tests work), but we want to disable it there.  Similarly, appveyor is marked as windows (where death tests work), but doesn't appear to support death tests either.

tests/CMakeLists.txt
kennyweiss: Good call.  I'll move this into the ``gtest_smoke`` test

davidbeckingsale:
Okay, that makes sense to me.

cmake/BLTOptions.cmake
```diff
@@ -77,3 +77,17 @@ if( NOT ( ( BLT_CXX_STD STREQUAL "c++98" )
     message(FATAL_ERROR "${BLT_CXX_STD} is an invalid entry for BLT_CXX_STD.
 Valid Options are ( c++98, c++11, c++14 )")
 endif()
+
+
+################################
+# Advanced configuration options
+################################
+
+option(ENABLE_GTEST_DEATH_TESTS
```
kennyweiss: I named the option ``ENABLE_GTEST_DEATH_TESTS`` and marked it as advanced.  
For consistency with the other options, I did not prefix it with ``BLT``.

Does that work for everyone?

tests/CMakeLists.txt
kennyweiss: Done.

cmake/BLTOptions.cmake
davidbeckingsale: Good with me.

PR #43, opened 2017-07-09
remove blt_add_component macro

### LINKED ISSUE #34: remove blt support for "components"
We have a small set of logic that supports the LLNL's Axom toolkit's "components" concept. 

This is a bit special case, it should exist in Axom's CMake logic instead of BLT.

cyrush:
This will resolve #34.

Since the components logic is special purpose, we are removing it from BTL and supporting it direclty in Axom.


davidbeckingsale:
Was the `add_component` macro ever used or tested inside of BLT?

kennyweiss:
@davidbeckingsale -- No, it was only used in axom.  That's why @cyrush is pulling it out.

PR #53, opened 2017-07-15
Feature/weiss27/bugfix win flags

kennyweiss:
Some extra flags for windows. 

Developed using Visual Studio 2010 ( MSVC 16.0.40219.1)

### DIFF
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -112,11 +112,6 @@ if(BLT_DEFINES)
 endif()
 
 if(COMPILER_FAMILY_IS_MSVC)
-    # Mark 32 bit executables large address aware so they can use > 2GB address space
-    if(CMAKE_SIZEOF_VOID_P MATCHES 4)
-        SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /LARGEADDRESSAWARE")
-    endif()
-
     # Visual studio can give a warning that /bigobj is required due to the size of some object files
     set( BLT_CXX_FLAGS "${BLT_CXX_FLAGS} /bigobj" )
     set( BLT_C_FLAGS   "${BLT_C_FLAGS} /bigobj" )
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -143,7 +143,7 @@ endif()
 #############################################
 if(BLT_CXX_FLAGS)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${BLT_CXX_FLAGS}")
-     message(STATUS "Updated CMAKE_CXX_FLAGS to \"${CMAKE_CXX_FLAGS}\"")
+    message(STATUS "Updated CMAKE_CXX_FLAGS to \"${CMAKE_CXX_FLAGS}\"")
 endif()
 
 ################################################
@@ -322,8 +322,7 @@ if (ENABLE_WARNINGS_AS_ERRORS)
     endforeach()
 endif()
 
-
-foreach(flagVar ${langFlags})   
+foreach(flagVar ${langFlags} "CMAKE_Fortran_FLAGS" "CMAKE_EXE_LINKER_FLAGS" )
     message(STATUS "${flagVar} flags are:  ${${flagVar}}")
 endforeach()
 
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -112,8 +112,6 @@ if(BLT_DEFINES)
 endif()
 
 if(COMPILER_FAMILY_IS_MSVC)
-    add_definitions(-D_USE_MATH_DEFINES)
-
     # Mark 32 bit executables large address aware so they can use > 2GB address space
     if(CMAKE_SIZEOF_VOID_P MATCHES 4)
         SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /LARGEADDRESSAWARE")
@@ -122,7 +120,6 @@ if(COMPILER_FAMILY_IS_MSVC)
     # Visual studio can give a warning that /bigobj is required due to the size of some object files
     set( BLT_CXX_FLAGS "${BLT_CXX_FLAGS} /bigobj" )
     set( BLT_C_FLAGS   "${BLT_C_FLAGS} /bigobj" )
-
 endif()
 
 ##########################################
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -314,11 +314,6 @@ if (ENABLE_WARNINGS_AS_ERRORS)
     endforeach()
 endif()
 
-foreach(flagVar ${langFlags} "CMAKE_Fortran_FLAGS" "CMAKE_EXE_LINKER_FLAGS" )
-    message(STATUS "${flagVar} flags are:  ${${flagVar}}")
-endforeach()
-
-
 ################################
 # Enable Fortran
 ################################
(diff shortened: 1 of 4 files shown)


cmake/SetupCompilerOptions.cmake
```diff
@@ -111,6 +111,20 @@ if(BLT_DEFINES)
     message(STATUS "Added \"${BLT_DEFINES}\" to definitions")
 endif()
 
+if(COMPILER_FAMILY_IS_MSVC)
+    add_definitions(-D_USE_MATH_DEFINES)
+
+    # Mark 32 bit executables large address aware so they can use > 2GB address space
```
kennyweiss: I had these from a previous project.  They are for 32-bit systems, so might be outdated.
(But my copy of Visual Studio still generates 32-bit executables).

cmake/SetupCompilerOptions.cmake
```diff
@@ -275,18 +289,22 @@ blt_append_custom_compiler_flag(
...
-        if(${flag_var} MATCHES "/MD")
+        
+        # Avoid static linking runtime issues on windows (adapted from RAJA)
+        if(NOT BUILD_SHARED_LIBS AND ${flag_var} MATCHES "/MD")
             string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-        endif(${flag_var} MATCHES "/MD")
+        endif()
+
+        # Clear runtime check flags -- incompatible with optimization flags
+        if(${flag_var} MATCHES "/RTC")
```
kennyweiss: This one might be related to an improper configuration setup.

The debug build was getting flags ``\O2`` (optimized) and ``\RTC1`` (run time check).
Visual studio gave an error about how they were incompatible.

If anyone has a better solution -- i.e. to get cmake to give the proper flags -- I'd appreciate it.

cmake/SetupCompilerOptions.cmake
```diff
@@ -111,6 +111,20 @@ if(BLT_DEFINES)
     message(STATUS "Added \"${BLT_DEFINES}\" to definitions")
 endif()
 
+if(COMPILER_FAMILY_IS_MSVC)
+    add_definitions(-D_USE_MATH_DEFINES)
```
cyrush: Was this necessary for Axom?  Wondering if it should be placed in our axom_config, instead of here. 

Seems like it will be needed by client code. 

https://msdn.microsoft.com/en-us/library/4hwaceh6.aspx

cmake/SetupCompilerOptions.cmake
cyrush: I haven't run into these issues with VS 2015 or 2017, but I have been doing 64-bit builds.

Appveyor is using 2015, and seems ok with these.  Could be harmless, what do folks think?

Kathleen Biagas from the VisIt team may have some wisdom -- I believe we have stopped building 32-bit versions of VisIt not to long ago.

For context, on VisIt, we use VS 2012 and VS 2013.
(here are the lib sets and projects for our tpls http://visit.ilight.com/svn/visit/trunk/windowsbuild/)

cmake/SetupCompilerOptions.cmake
cyrush: was \O2 getting added to a Debug build?
(https://stackoverflow.com/questions/37007939/command-line-error-d8016-o2-and-rtc1-command-line-options-are-incompatibl)

cmake/SetupCompilerOptions.cmake
kennyweiss: I'd appreciate feedback from Kathleen (or others) on this.

Unfortunately, the only version of Visual Studio that I had access to is pretty old -- VS 2010, and doesn't have 64-bit or ``C++11`` capabilities.

I think I can get an express version working the next time I have time to look into this.

However, (I think) the free version is missing some features like folders support, that I was hoping to play with.

cmake/SetupCompilerOptions.cmake
kennyweiss: That is a good question.  

I don't think we are properly setting up the different configurations for Visual Studio.  This might be related to cmake's generator expression, but I don't have much experience with this.

The o2/rtc fix was the only thing I could figure out to get rid of the error.  

Would definitely appreciate a better solution to this one, if anyone has more experience.

cmake/SetupCompilerOptions.cmake
kennyweiss: Yes. 

So, you're suggesting that this should go into the ``config.hpp`` of user projects?

cyrush: Yes -- since you need this define if you are using the math headers, I think it makes sense for projects using the math headers to provide a solution instead of blt always adding it.
For axom, that would be in our generated config.hpp.

kennyweiss:
I removed the line ``add_definitions(-D_USE_MATH_DEFINES) `` from Windows configurations. 
This is something that user codes will have to add.

cmake/SetupCompilerOptions.cmake
```diff
@@ -111,6 +111,17 @@ if(BLT_DEFINES)
     message(STATUS "Added \"${BLT_DEFINES}\" to definitions")
 endif()
 
+if(COMPILER_FAMILY_IS_MSVC)
+    # Mark 32 bit executables large address aware so they can use > 2GB address space
+    if(CMAKE_SIZEOF_VOID_P MATCHES 4)
+        SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /LARGEADDRESSAWARE")
+    endif()
+
```
cyrush: Still not sure if we need these fixes, adding some refs to the discussion:

https://stackoverflow.com/questions/2288728/drawbacks-of-using-largeaddressaware-for-32-bit-windows-executables

cmake/SetupCompilerOptions.cmake
```diff
@@ -111,6 +111,17 @@ if(BLT_DEFINES)
...
+    # Visual studio can give a warning that /bigobj is required due to the size of some object files
+    set( BLT_CXX_FLAGS "${BLT_CXX_FLAGS} /bigobj" )
+    set( BLT_C_FLAGS   "${BLT_C_FLAGS} /bigobj" )
+endif()
```
cyrush: This one seems less perilous:

https://msdn.microsoft.com/en-us/library/ms173499.aspx

cmake/SetupCompilerOptions.cmake
cyrush: I think we should try to hunt this one down, `\RTC` should only be used w/o optimization, see **Remarks** in:

https://msdn.microsoft.com/en-us/library/8wtf2dfz.aspx

I think we want rtc on for debug builds, it can help us discover output bounds issues, etc. We just need to uncover why `\O2` is also being added to debug builds on Windows. 

I'll try a debug build with blt-test on windows and see if I get any more hints.

kennyweiss: I was able to track this one down and it was an error on our end.  This workaround for BLT is not necessary, and I will remove it.

The problem was due to a flag in one of our tests for warnings about aliasing where we were explicitly adding the ``O2`` flag (even in debug configurations).  There was a note in our CMake setup code that this would likely need a workaround for msvc (and it does).  

I will take care of it.

kennyweiss:
Thanks @cyrush for the comments about the flags.
I removed the commit with the ``\RTC`` override, as well as the ``\LARGEADDRESSAWARE`` flag.

I think this PR should be ready at this point.

kennyweiss:
I updated the ``blt_add_test`` macro to support multi-configuration generators (e.g. Visual Studio).

@cyrush, @davidbeckingsale and @white238 -- could you please check it out and review it again, when you get a chance so we can merge this PR?

cyrush:
@kennyweiss where these changes just to support MPI tests?

I haven't tested MPI on widows for conduit, but I have been using BTL for my other tests in Visual Studio w/o issues.

kennyweiss:
@cyrush -- Yes.  After installing MS-MPI on my system and enabling MPI,  the ``blt_mpi_smoke`` test (and other MPI tests) failed.  

The problem is that we were patching the path to our executables based on the ``runtime_executable_path``, but in multi-configuration generators, the actual path has the build type  (e.g. ``build/tests/Debug/blt_smoke_test``).

kennyweiss:
After several false starts, I was able to get MPI to work with Appveyor.
Please let me know if there are any remaining issues, I think this PR should be ready to merge.

Issue #50, opened 2017-07-14
gtest and gmock emit MACOSX_RPATH warnings


cyrush:
```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   gmock
   gmock_main
   gtest
   gtest_main

This warning is for project developers.  Use -Wno-dev to suppress it.
```

cyrush:
resolved in #59

Issue #55, opened 2017-07-15
blt_gmock_smoke crashes with gcc 4.9.3 on surface


cyrush:
It's crashing when cleaning up global objects:

`6  0x00002aaaab1f1713 in __do_global_dtors_aux () from build-debug/lib/libgmock_main.so`

this may be an issue with our smoke test, or an issue with gmock

Here are some (likely) related threads:

- https://groups.google.com/forum/#!topic/googlemock/5W-zJKZFtE0
- (this issue is refed from above thread: https://github.com/google/googlemock/issues/54)
- https://github.com/google/googletest/issues/461

cyrush:
Our example doesn't include  any global instances of objects, so I think this probably bad luck with global destruction of something in gmock.

white238:
GMock has been updated multiple times since then.  If this didn't get fixed then someone will create a new issue.

PR #56, opened 2017-07-17
enh: add ENABLE_GMOCK option

cyrush:
Adds a option to enable google mock support (default = OFF)

### DIFF
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -61,7 +61,9 @@ option(ENABLE_MPI          "Enables MPI support" OFF)
 option(ENABLE_OPENMP       "Enables OpenMP compiler support" OFF)
 option(ENABLE_CUDA         "Enable CUDA support" OFF)
 
-option(ENABLE_GMOCK        "Enable Google Mock testing support" OFF)
+# Controls if Google Mock is built and available for use. 
+# If ENABLE_TESTS=OFF, no testing support is built and this option is ignored.
+option(ENABLE_GMOCK        "Enable Google Mock testing support (if ENABLE_TESTS=ON)" OFF)
 
 ################################
 # Compiler Options
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -61,6 +61,7 @@ option(ENABLE_MPI          "Enables MPI support" OFF)
 option(ENABLE_OPENMP       "Enables OpenMP compiler support" OFF)
 option(ENABLE_CUDA         "Enable CUDA support" OFF)
 
+option(ENABLE_GMOCK        "Enable Google Mock testing support" OFF)
 
 ################################
 # Compiler Options
--- a/thirdparty_builtin/CMakeLists.txt
+++ b/thirdparty_builtin/CMakeLists.txt
@@ -64,11 +64,15 @@ if(ENABLE_TESTS)
                          INCLUDES ${gtest_SOURCE_DIR}/include
                          LIBRARIES gtest_main gtest
                          DEFINES  ${gtest_defines})
-
-    blt_register_library(NAME gmock
-                         INCLUDES ${gmock_SOURCE_DIR}/include
-                         LIBRARIES gmock_main gmock
-                         DEFINES  ${gtest_defines})
+    
+    message(STATUS "Google Mock Support is ${ENABLE_GMOCK}")
+    
+    if(ENABLE_GMOCK)
+        blt_register_library(NAME gmock
+                             INCLUDES ${gmock_SOURCE_DIR}/include
+                             LIBRARIES gmock_main gmock
+                             DEFINES  ${gtest_defines})
+    endif()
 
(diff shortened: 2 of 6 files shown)


cyrush:
To resolve #50.

We can discuss if the default should be `OFF` or `ON`, I choose `OFF` b/c of #55

cyrush:
For symmetry, we talked about also adding `ENABLE_GTEST` and `ENABLE_FRUIT`.
If possible, I would like to do this as part of another PR.

cmake/BLTOptions.cmake
```diff
@@ -61,6 +61,7 @@ option(ENABLE_MPI          "Enables MPI support" OFF)
 option(ENABLE_OPENMP       "Enables OpenMP compiler support" OFF)
 option(ENABLE_CUDA         "Enable CUDA support" OFF)
 
+option(ENABLE_GMOCK        "Enable Google Mock testing support" OFF)
```
cyrush: It mirrors `ENABLE_MPI`, `ENABLE_OPENMP`, and `ENABLE_CUDA`.
It's also grouped with them in `BLTOptions.cmake`

I thought about adding a more descriptive comment that acknowledges the fact that this also screened by `ENABLE_TESTS` -- would that suffice?

thirdparty_builtin/googletest-release-1.8.0/CMakeLists.txt
```diff
@@ -16,13 +16,21 @@ project( googletest-distribution )
 
 enable_testing()
 
-option(BUILD_GTEST "Builds the googletest subproject" OFF)
+
+##########################################################
```
davidbeckingsale: Yeah, I think it makes sense to do this at a higher level.

cmake/BLTOptions.cmake
kennyweiss: I think so.

It would be good to think about a clean way to set up sub-options for a given TPL/component (e.g. to include the name of the TPL/component that is is related to).

kennyweiss: E.g. it we had 
* ``TESTS_ENABLE_GMOCK`` 
* ``TESTS_ENABLE_GTEST``
* ``TESTS_ENABLE_FRUIT`` 
* ...

Then these would show up in the ``TESTS`` group in ``ccmake``/``cmake-gui`` rather than the ``ENABLE`` group.  
Pro: This would declutter the ``ENABLE`` group
Con: But, it might make the options less discoverable.

cyrush: `TESTS_ENABLE_GMOCK ` to me implies "enable the gmock tests" vs build gmock

kennyweiss: Yeah, it is definitely not perfect, and I don't think this should hold up the current PR.

But I was aiming for something clean along these lines.

thirdparty_builtin/googletest-release-1.8.0/CMakeLists.txt
cyrush: I'll look into it -- Note: We already have a BLT specific fix in this file on master, we should try to move that as well.

cyrush:
I moved our `ENABLE_GMOCK` logic into one of our `CMakeLists.txt` files, and added more details in `BLTOptions.cmake` related to the fact that `ENABLE_GMOCK` is also guarded by `ENABLE_TESTS`.

I also tried to move the policy command related to `project(zzz VERSION "x.y.z)` out of the master googletest `CMakeLists.txt` file, but that change has to stay there  to take effect -- I believe b/c of the use of the `cmake_minimum_version` command.

cyrush:
@kennyweiss are these changes sufficient to merge?

thirdparty_builtin/CMakeLists.txt
```diff
@@ -57,18 +57,36 @@ if(ENABLE_TESTS)
         list(APPEND gtest_defines "-DGTEST_HAS_DEATH_TEST=0")
     endif()
 
+    #
+    # Use BLT's ENABLE_GMOCK option to set the build options for 
+    # gtest's cmake project
+    #
+    
+    if(ENABLE_GMOCK)
```
white238: This off/on and on/off logic doesn't make sense to me. Can you explain? Why does enabling gmock disable building gtest

cyrush: It's the way that googletest's cmake project works.

First, I don't belive you can build gmock w/o gtest. 

You either build gmock (using their `BUILD_GMOCK=ON` which also builds gtest), or you build gtest (and don't build gmock)

I haven't checked, so I don't know if you can actually set both to OFF (for a future case where we would also want to support our own`ENABLE_GTEST` -- but we could always simply not include the projects as our own guard.

cyrush: Check out how they are used in:

https://github.com/LLNL/blt/blob/master/thirdparty_builtin/googletest-release-1.8.0/CMakeLists.txt

While keeping in mind that building gmock also takes care of building gtest.

PR #61, opened 2017-07-21
Adding pthreads to gtest as a dependent library. This fixes undefined…

mclarsen:
… symbols while using intel compilers

This pull request fixes undefined symbols on surface with the intel 16 compilers:

[  6%] Linking CXX executable ../../tests/blt_gtest_smoke
../../lib/libgtest.so: undefined reference to `pthread_key_create'
../../lib/libgtest.so: undefined reference to `pthread_getspecific'
../../lib/libgtest.so: undefined reference to `pthread_key_delete'
../../lib/libgtest.so: undefined reference to `pthread_setspecific'

nm libgtest.so | grep thread
                 w pthread_cancel
                 U pthread_getspecific
                 U pthread_key_create
                 U pthread_key_delete
                 U pthread_mutex_destroy@@GLIBC_2.2.5
                 U pthread_mutex_init@@GLIBC_2.2.5
                 U pthread_mutex_lock@@GLIBC_2.2.5
                 U pthread_mutex_unlock@@GLIBC_2.2.5
                 U pthread_self@@GLIBC_2.2.5
                 U pthread_setspecific


cyrush:
I think this may be a safer solution, adapted from my use of thread linking in civetweb for conduit:

```
if(UNIX AND NOT APPLE)
    find_package( Threads )
    set(gtest_extra_libs ${CMAKE_THREAD_LIBS_INIT})
endif()
```

mclarsen:
What about intel compilers on osx? I don't have the compilers to test it on my machine. Perhaps just if Unix?

cyrush:
@mclarsen -- we can keep if UNIX and fix if an issue pops up

(the logic I pulled that from was also guarding librt, which doesn't work on osx)

mclarsen:
Hmm. CMake 3.3 didn't find threads.

mclarsen:
`Looking for include file pthread.h`
`CMake Error at /home/travis/build/LLNL/tpls/cmake/share/cmake-3.3/Modules/CheckIncludeFiles.cmake:79 (try_compile):`
`Unknown extension ".c" for file`
` /home/travis/build/LLNL/travis-debug-build/CMakeFiles/CMakeTmp/CheckIncludeFiles.c`

davidbeckingsale:
You need to have `C` as a project language when looking for threads. We had to do this in RAJA.

cyrush:
So we need to add C in addition to CXX in the project command in the test project:
https://github.com/LLNL/blt/blob/master/cmake/blt-test/CMakeLists.txt

davidbeckingsale:
The problem is that anyone using BLT and with `ENABLE_TESTING` would also need to have C enabled.

cyrush:
Added #71, which includes an option to enable / disable gtest and fruit. 

If we guard the check for pthreads w/ ENABLE_GTEST, 
By disabling gtest, allows fortran folks to use BLT with out requiring their cmake project includes C

cyrush:
@mclarsen, i merged #71 -- if you re-work you changes from master (really just shifting down your current logic into the ENABLE_GTEST block) -- we should be able to merge this PR. 

(Not sure if I can checkout your fork to try to make the changes)

mclarsen:
@cyrush , I think this is good to go.

cyrush:
@davidbeckingsale  FYI:

since @mclarsen's changes were on a fork, the appveyor reporting for the branch test was hung up. This was also the problem with @gonsie's pr #87.

I had to change the appveyor branch check to not be required to avoid this.

The hint was here:

http://help.appveyor.com/discussions/problems/3612-github-showing-multiple-checks#comment_38488858

Issue #29, opened 2017-04-13
Change LCOV_PATH to LCOV_EXECUTABLE,  GENHTML_PATH to GENHTML_EXECUTABLE


cyrush:
The name path is ambiguous in this case, we want use ZZZ_DIR for the case where we are looking for some standard directory layout, and ZZZ_EXECTUABLE when we point directly to a tool.

cyrush:
Adam Kunen was exploring some clang tools for coverage that generate reports

cyrush:
resolved in #62

PR #62, opened 2017-07-25
refactor code-coverage support

cyrush:
Breaks up code coverage flags and code report generation logic into separate files. 

changes `LCOV_PATH`, `GENHTML_PATH`, and `GCOV_PATH` to:

`LCOV_EXECUTABLE`, `GENHTML_EXECUTABLE`, `GCOV_EXECUTABLE`

Makes code coverage reports optional on if `LCOV_EXECUTABLE`, `GENHTML_EXECUTABLE`, `GCOV_EXECUTABLE` are specified.


cyrush:
The primary goal of this is to keep the coverage reporting features from undermining builds that just want coverage flags. 

This will also resolve #29 

Note: Since the lcov doesn't build on OSX, I can't really test the reporting.

cmake/SetupCodeCoverageReports.cmake
```diff
@@ -0,0 +1,158 @@
...
+# - Refactored to only include report gen logic, not coverage flags
+#
+###############################################################################
+
+set(BLT_CODE_COVERAGE_REPORTS ON)
+
+# Check for lcov
+if(NOT EXISTS ${LCOV_EXECUTABLE})
+    message(STATUS "Code coverage: Unable to find lcov, disabling code coverage reports.")
+endif()
```
white238: I think you missed the set(BLT_CODE_COVERAGE_REPORTS OFF) here

cmake/SetupCodeMetrics.cmake
```diff
@@ -49,7 +49,28 @@
...
-    include(blt/cmake/CodeCoverage.cmake)
+    ##########################################################################
+    # Setup coverage compiler flags 
+    ##########################################################################
+    # Set the actual flags for coverage in the COVERAGE_FLAGS variable 
+    # Note: '--coverage' is equivalent to '-fprofile-arcs -ftest-coverage' for compiling and '-lgcov' for linking
+    # Additional flags that might be useful: 
+    #       " -fno-inline -fno-inline-small-functions -fno-default-inline"
+    blt_append_custom_compiler_flag(FLAGS_VAR   COVERAGE_FLAGS 
+                                    DEFAULT     "--coverage")
```
kennyweiss: Is ``--coverage`` valid on all compilers?  Or only on gcc / clang ?
The previous version had a check that ensured that coverage was only supported on gcc. 
Otherwise, you'll need to set flags for the other compilers.

cmake/SetupCodeMetrics.cmake
```diff
@@ -49,7 +49,28 @@
...
+        
+    SET( CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} ${COVERAGE_FLAGS}" )
+    SET( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${COVERAGE_FLAGS}" )
```
kennyweiss: The previous version set the linker flags as well, and differentiated between static and dynamic linking flags.  
Is this not necessary anymore?

kennyweiss:
> Note: Since the lcov doesn't build on OSX, I can't really test the reporting.

Did you test this to make sure the reporting works on linux?

cmake/SetupCodeMetrics.cmake
cyrush: `--coverage` works for me on clang and gcc, it may not work on other compilers. 

The old logic turned it off for clang, but i think that may have been tangled up with the reporting (gcov)

Since its optional,  I am comfortable with folks to turn it on when they know it will work. 

We can improve as we get more info about what works where.

cmake/SetupCodeMetrics.cmake
cyrush: They where actually empty:

```
SET(CMAKE_SHARED_LINKER_FLAGS_COVERAGE
    ""
    CACHE STRING "Flags used by the shared libraries linker during coverage builds."
    FORCE )
```

cyrush: Note this is what I use for conduit:

```
# Coverage Flags
if(ENABLE_COVERAGE)
    message(STATUS "Building using coverage flags (ENABLE_COVERAGE == ON)")
    set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} --coverage")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --coverage")
else()
    message(STATUS "Building without coverage flags (ENABLE_COVERAGE == OFF)")
endif()

```

cmake/CodeCoverage.cmake
```diff
@@ -1,240 +0,0 @@
...
-    
-if(ENABLE_FORTRAN)
-    SET(CMAKE_Fortran_FLAGS_COVERAGE
-        "-g -O0 ${COVERAGE_FLAGS}"
-        CACHE STRING "Flags used by the Fortran compiler during coverage builds."
-        FORCE )
-    MARK_AS_ADVANCED(CMAKE_Fortran_FLAGS_COVERAGE)
-endif()
-
-SET(CMAKE_EXE_LINKER_FLAGS_COVERAGE
```
kennyweiss: This linker flag wasn't empty.  I think it was important.

cyrush: Gotcha, I'll add it back since you recall it being useful.

cmake/SetupCodeMetrics.cmake
kennyweiss: The ``blt_append_custom_compiler_flags`` macro allows you to set one flag per supported compiler family.  You should avoid setting a flag for a compiler that is not supported.  E.g.  if ``--coverage`` does not work for intel, it could give a compiler error.  setting ``INTEL " "`` should fix this.

cmake/CodeCoverage.cmake
kennyweiss: https://stackoverflow.com/questions/566472/where-are-the-gcov-symbols

cyrush: perfect thanks for the ref!

cyrush:
> Did you test this to make sure the reporting works on linux?

No I did not, I don't when when this has last been tested. 

My main goal is to stop this from breaking cases where folks don't need the coverage reporting, but need coverage flags.

cmake/SetupCodeMetrics.cmake
cyrush: Ideally, it would be good to add the proper cases for the known compilers.

I have only done coverage with gcc and clang, so those are all I can vouch for. 

I am all for adding the correct path for intel in the future. 

Since this is screened by `ENABLE_COVERAGE`,  I think the default is ok until we have more info.

kennyweiss: I was suggesting that we only set the flag on the compilers that we know are valid, rather than setting it for all compilers.  We can add the proper flags when we need them.

cyrush: The old logic simply error when it wasn't gcc and used the `DEFAULT` path.

I can make that change.

It looks like we use DEFAULT, for gcc?

What else should I add?  I see cases like this for XL, are there existing examples for `INTEL`

cyrush: Changed `DEFAULT` to " " and added `--coverage` for `GNU` and `CLANG`

PR #67, opened 2017-07-31
Add clang-cuda support

davidbeckingsale:
Add support for clang's native CUDA compilation.

### DIFF
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -60,6 +60,9 @@ option(ENABLE_SHARED_LIBS  "Enables shared libraries." OFF)
 option(ENABLE_MPI          "Enables MPI support" OFF)
 option(ENABLE_OPENMP       "Enables OpenMP compiler support" OFF)
 option(ENABLE_CUDA         "Enable CUDA support" OFF)
+option(ENABLE_CLANG_CUDA   "Enable Clang's native CUDA support" OFF)
+mark_as_advanced(ENABLE_CLANG_CUDA)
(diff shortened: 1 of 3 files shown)


cmake/BLTOptions.cmake
```diff
@@ -60,6 +60,9 @@ option(ENABLE_SHARED_LIBS  "Enables shared libraries." OFF)
 option(ENABLE_MPI          "Enables MPI support" OFF)
 option(ENABLE_OPENMP       "Enables OpenMP compiler support" OFF)
 option(ENABLE_CUDA         "Enable CUDA support" OFF)
+option(ENABLE_CLANG_CUDA   "Enable Clang's native CUDA support" OFF)
+mark_as_advanced(ENABLE_CLANG_CUDA)
+set(BLT_CUDA_ARCH "sm_30" CACHE STRING "Compute architecture to use when generating CUDA code")
```
davidbeckingsale: I'm not sure the best way to handle this argument in normal CUDA builds - do we want to force it into `CUDA_NVCC_FLAGS`?

davidbeckingsale:
@cyrush @white238 any thoughts on this?

cmake/BLTOptions.cmake
cyrush: To be consistent, we should use this for both cases (cuda-clang + nvcc), adding to NVCC FLAGS seems ok.

Or maybe make it optional, and have a default for cuda-clang?

davidbeckingsale:
@cyrush - I updated this to make the "arch" argument specific to clang-cuda. I think this is the best approach for now.

PR #71, opened 2017-08-10
add blt options for gtest and fruit

cyrush:
Adds:

 `ENABLE_GTEST` (default = ON)
 `ENABLE_FRUIT` (default = ON (also guarded by `ENABLE_FORTRAN` = ON)

Like `ENABLE_GMOCK`, these are guarded by `ENABLE_TESTS`.

gmock is also guarded by `ENABLE_GTEST`, 
I think this makes sense b/c building gmock always builds gtest.

### DIFF
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -61,9 +61,13 @@ option(ENABLE_MPI          "Enables MPI support" OFF)
 option(ENABLE_OPENMP       "Enables OpenMP compiler support" OFF)
 option(ENABLE_CUDA         "Enable CUDA support" OFF)
 
-# Controls if Google Mock is built and available for use. 
-# If ENABLE_TESTS=OFF, no testing support is built and this option is ignored.
+# Options that control if Google Test, Google Mock, and Fruit are built 
+# and available for use. 
+# If ENABLE_TESTS=OFF, no testing support is built and these option are ignored.
+option(ENABLE_GTEST        "Enable Google Test testing support (if ENABLE_TESTS=ON)" ON)
 option(ENABLE_GMOCK        "Enable Google Mock testing support (if ENABLE_TESTS=ON)" OFF)
+option(ENABLE_FRUIT        "Enable Fruit testing support (if ENABLE_TESTS=ON and ENABLE_FORTRAN=ON)" ON)
+    
 
 ################################
 # Compiler Options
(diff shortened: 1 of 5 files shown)


cyrush:
This it should help with #61.

This is a WIP (need to fix tests in example),

cyrush:
resolve testing issues, let me know if you have any suggestions

thirdparty_builtin/CMakeLists.txt
```diff
@@ -70,27 +70,36 @@ if(ENABLE_TESTS)
...
+    
+    if(ENABLE_GTEST)
+        # Enable builtin google test 
+        add_subdirectory(googletest-release-1.8.0)
+
+        blt_register_library(NAME gtest
+                             INCLUDES ${gtest_SOURCE_DIR}/include
+                             LIBRARIES gtest_main gtest
                              DEFINES  ${gtest_defines})
+         if(ENABLE_GMOCK)
```
davidbeckingsale: Should `ENABLE_GMOCK` just force `ENABLE_GTEST`?

cyrush: Yeah, that sounds good -- either that or we should add an error message.

cyrush: @davidbeckingsale  I added logic to force ENABLE_GTEST to ON, if ENABLE_GMOCK is ON

cyrush:
@white238, what do you think?

cmake/BLTOptions.cmake
```diff
@@ -61,9 +61,17 @@ option(ENABLE_MPI          "Enables MPI support" OFF)
...
+# Options that control if Google Test, Google Mock, and Fruit are built 
+# and available for use. 
+#
+# If ENABLE_TESTS=OFF, no testing support is built and these option are ignored.
+#
+# Google Mock requires and always builds Google Test, so ENABLE_GMOCK=ON implies ENABLE_GTEST=ON.
+#
+option(ENABLE_GTEST        "Enable Google Test testing support (if ENABLE_TESTS=ON)" ON)
 option(ENABLE_GMOCK        "Enable Google Mock testing support (if ENABLE_TESTS=ON)" OFF)
+option(ENABLE_FRUIT        "Enable Fruit testing support (if ENABLE_TESTS=ON and ENABLE_FORTRAN=ON)" ON)
```
white238: Do you check if they enable fruit but don't enable fortran?  Same with enable gtest/gmock/fruit without enabling test... I would rather error and exit then ignore an option that they probably got wrong.

cyrush: reviewed with @white238, the fact that these are guarded in a nested manner ( fruit support requires ENABLE_FORTRAN and ENABLE_FRUIT) means we shouldn't hit any strange cases.

PR #73, opened 2017-08-25
[RFC] Dibs

DavidPoliakoff:
This is a prototype for some components I've developed with @becker33 for integrating Spack into CMake builds. It needs a lot of mechanical work, but I'm curious whether this idea is of interest to the group, and if so, what we want the implementation to look like.

I am really frustrated with TPL management in CMake. I am really happy with TPL management in Spack. The idea in Dibs is to allow users to replace most of the clunky ExternalProject_Add or FindTPL.cmake files they have with calls out to Spack.

To get a feel for the desired end interface, check [the example](https://github.com/DavidPoliakoff/blt/blob/db0207066d4c91353a20d47a22096da2c6d6a59d/cmake/dibs/dibs_test/CMakeLists.txt)

I typically invoke this as follows (from within a build directory)

> dibs gcc@4.9.3 -DCMAKE_INSTALL_PREFIX=$HOME ..
> source .envrc
> make

Mechanically, what is happening is me calling Spack to install packages or query things about them. I use the direct modulecmd, the thing underlying the module alias, to create an environment file which reflects the contents of the module Spack creates for those packages. This can be used with [direnv](https://github.com/direnv/direnv) or manually sourced.

From there, when a user types make, they get a valid build. This is getting wall-of-texty, but please let me know if you have questions

I've talked with a few people about this PR, I'll tag them here (@tgamblin , @trws, @ajkunen)

Dire issues to resolve

- [ ] Productization work
- [ ] Allowing overrides if the user knows a package might already be installed somewhere _not_ with Spack
- [ ] Allowing groups to share installs, so that not every member of PhysicsTeam has to install Caliper with Intel 16.bleh 
- [ ] Deployment: the root of BLT has no CMakeLists.txt. I can see a few models for deploying this, but we should pick one
- [ ] Do we install Spack if users don't have it?
- [ ] The shell components of this are assuming Bash. They shouldn't


DavidPoliakoff:
Also pinging @cyrush because we've talked about this, @davidbeckingsale for technical advice, and @alalazo because I think he'd find this amusing.

cmake/dibs/dibs_test/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+project(dibs_test)
+cmake_minimum_required(VERSION 2.8)
+include(../build/dibs.cmake)
```
DavidPoliakoff: TODO: make this a little better productized

DavidPoliakoff:
Also adding @rountree, as we talked briefly about this yesterday (day before?)

cyrush:
Hi @DavidPoliakoff this looks interesting, but but I would like to hold of on tightly coupling spack into BLT. 

The projects I am aware of using BLT are opinionated about strategies for including TPLS. 

Also, we have folks using BLT to support their builds on Windows, so relying on the spack integration wont help their FIndTPL woes there (If they do use this, they will need two solutions, which we want to avoid)

I do think this is a good idea, but it seems like a separate project from BLT.

ajkunen:
@cyrush I would really love support for this, especially if:
1) it allows the user to override with the normal FindXXX command line arguments, and point to a prebuilt library
2) on windows it just devolves to 1, so that it just acts like FindXXX

DavidPoliakoff:
Closing, will probably one day (is not currently) be its own project at [LLNL/Dibs](https://www.github.com/LLNL/dibs)

PR #76, opened 2017-08-29
System includes for external libraries

white238:
This is to enable using system includes (as long as the compiler supports it) to suppress warnings from external header files.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -147,7 +147,8 @@ endmacro(blt_add_target_link_flags)
 ##------------------------------------------------------------------------------
 ## blt_register_library( NAME <libname>
 ##                       DEPENDS_ON [dep1 [dep2 ...]]
-##                       INCLUDES [include1 [include2 ...]] 
+##                       INCLUDES [include1 [include2 ...]]
+##                       USE_SYSTEM_INCLUDES [ON] 
 ##                       FORTRAN_MODULES [ path1 [ path2 ..]]
 ##                       LIBRARIES [lib1 [lib2 ...]]
 ##                       COMPILE_FLAGS [ flag1 [ flag2 ..]]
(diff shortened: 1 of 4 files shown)


cmake/BLTMacros.cmake
```diff
@@ -161,12 +162,17 @@ endmacro(blt_add_target_link_flags)
 ## the DEPENDS_ON in your blt_add_executable call and it will add the INCLUDES
 ## and LIBRARIES to that executable.
 ##
+## TREAT_INCLUDES_AS_SYSTEM marks the utilizes the compiler's system include path
```
kennyweiss: Please clarify this sentence (``..marks the utilizes..``).

cmake/BLTMacros.cmake
```diff
@@ -147,7 +147,8 @@ endmacro(blt_add_target_link_flags)
 ##------------------------------------------------------------------------------
 ## blt_register_library( NAME <libname>
 ##                       DEPENDS_ON [dep1 [dep2 ...]]
-##                       INCLUDES [include1 [include2 ...]] 
+##                       INCLUDES [include1 [include2 ...]]
+##                       TREAT_INCLUDES_AS_SYSTEM [ON]
```
kennyweiss: In this case, does it make sense to have ``TREAT_INCLUDES_AS_SYSTEM`` be a cmake ``option``, since users are manually setting it, and the only supported case in for this parameter to be set to ``ON``?

(I know that macro options can be problematic, in general)

cmake/BLTMacros.cmake
```diff
@@ -161,12 +162,17 @@ endmacro(blt_add_target_link_flags)
...
+## setting for this libraries include directories.  This is useful if the headers
+## generate warnings you want to not have them reported in your build.
+##
 ## This does not actually build the library.  This is strictly to ease use after
 ## discovering it on your system or building it yourself inside your project.
 ##
 ## Output variables (name = "foo"):
 ##  BLT_FOO_DEPENDS_ON
 ##  BLT_FOO_INCLUDES
+##  BLT_FOO_TREAT_INCLUDES_AS_SYSTEM
```
kennyweiss: Should the documentation reflect the fact that this one is only optionally generated?

cmake/BLTMacros.cmake
white238: im trying to avoid options at this point..

cmake/BLTMacros.cmake
white238: im debating whether we should even expose users to what is going on in the background.

cmake/BLTMacros.cmake
kennyweiss: Here is the relevant code:

```cmake
  if( arg_TREAT_INCLUDES_AS_SYSTEM )
    set(BLT_${uppercase_name}_TREAT_INCLUDES_AS_SYSTEM ON CACHE BOOL "" FORCE)
    mark_as_advanced(BLT_${uppercase_name}_TREAT_INCLUDES_AS_SYSTEM)
  endif()
```
We are currently treating this ``singleValueArg`` as an ``option`` -- if it is there, regardless of its value, we are marking the includes as system.

cyrush: so we need to swtich to 
```cmake
  if( ${arg_TREAT_INCLUDES_AS_SYSTEM} )
  ...
  endif()
```

?

kennyweiss: Yes, I think so.

PR #82, opened 2017-09-05
Bugfix/probinso/cuda fixes

robinson96:
Addresses three CUDA build configuration fixes: 

- Filters CUDA / Non Cuda selctions by file extension. 
- Fixes a duplicate library target issue when use separable compilation. 
- Fixes an issue that made NVCC_COMPILE_DEFINITIONS incompatible with defines that had quotes in them.


cmake/BLTPrivateMacros.cmake
```diff
@@ -214,12 +214,14 @@ macro(blt_setup_cuda_source_properties)
         message( FATAL_ERROR "Must provide TARGET_SOURCES to the 'blt_setup_cuda_source_properties' macro")
     endif()
 
-
+    set ( _non_cuda_extensions .f .F .f90 .F90 .c .h )
```
davidbeckingsale: I think that we should switch away from extension-based "non-cuda" stuff... take a look at #28. We could do something similar for C files also.

cmake/thirdparty/FindCUDA/run_nvcc.cmake
```diff
@@ -74,7 +74,7 @@ set(CUDA_NVCC_FLAGS @CUDA_NVCC_FLAGS@ ;; @CUDA_WRAP_OPTION_NVCC_FLAGS@) # list
 @CUDA_NVCC_FLAGS_CONFIG@
 set(nvcc_flags @nvcc_flags@) # list
 set(CUDA_NVCC_INCLUDE_DIRS "@CUDA_NVCC_INCLUDE_DIRS@") # list (needs to be in quotes to handle spaces properly).
-set(CUDA_NVCC_COMPILE_DEFINITIONS "@CUDA_NVCC_COMPILE_DEFINITIONS@") # list (needs to be in quotes to handle spaces properly).
+set(CUDA_NVCC_COMPILE_DEFINITIONS @CUDA_NVCC_COMPILE_DEFINITIONS@) # list
```
davidbeckingsale: Should this be here?

robinson96: Yup. Quoting CUDA_NVCC_COMPILE_DEFINITIONS breaks defines that include quotes.  I think it's a copy-paste error from the line above, where it makes sense to quote the INCLUDE_DIRS to allow for spaces in paths.

davidbeckingsale: Sounds good 😄

cmake/BLTPrivateMacros.cmake
robinson96: Fair enough. Reverted.

davidbeckingsale:
@robinson96 I'd appreciate your feedback on #28 I think we want to make the non-cuda languages selectable, with defaults being Fortran and C.

Issue #79, opened 2017-08-30
windows: lots of gtest warnings on windows when building shared libs


cyrush:
Example:
`C:\projects\conduit\src\blt\thirdparty_builtin\googletest-release-1.8.0\googletest\include\gtest/internal/gtest-internal.h(589): warning C4251: 'testing::internal::TypedTestCasePState::registered_tests_': class 'std::map<std::string,testing::internal::CodeLocation,std::less<_Kty>,std::allocator<std::pair<const _Kty,_Ty>>>' needs to have dll-interface to be used by clients of class
`

For more examples, see:
https://ci.appveyor.com/project/cyrush/conduit/build/1.0.643

I believe these didn't happen with gtest 1.7.

Seems like this issue:
https://github.com/google/googletest/issues/860

This PR may provide a fix:
https://github.com/google/googletest/pull/862

(But this PR hasn't been merged into gtest)

cyrush:
In the interim, It may be possible to mute these issues by adding extra compiler flags or defines when we use `blt_register_library` for gtest and gmock when on windows.

cyrush:
Adding `/wd4251 /wd4251` to the compile flags in `blt_reg_lib` for gtest and gmock squashes these.

This isn't likely as targeted as the solution in the pr for gtest, but I think its a good temp solution. I'll work on a pr.

cyrush:
closing, resolved in PR #83

Issue #81, opened 2017-08-30
Support for additional compilers in the  blt_append_custom_compiler_flag macro


kennyweiss:
The ``blt_custom_compiler_flag`` macro allows users to define an appropriate compiler-specific value for a compiler flag and currently allows setting different values for: ``gcc``, ``clang``, ``ibm``, ``intel`` and ``msvc``.  

BLT determines the appropriate compiler based on the ``${CMAKE_BUILD_TOOL}`` variable for ``msvc`` and on ``${CMAKE_CXX_COMPILER_ID}`` for the other compilers, see ``cmake/SetupCompilerOptions.cmake``

The current logic only supports the visual studio compiler on windows.
This ticket is a feature request to support the intel compiler on windows.

kennyweiss:
Resolved in #112

PR #83, opened 2017-09-05
add flags to avoid gtest 1.8 dll-iface warnings on windows

cyrush:
Avoid warnings described in #79.

WIP since I need to test in conduit on windows to double check it works as expected.

### DIFF
--- a/thirdparty_builtin/CMakeLists.txt
+++ b/thirdparty_builtin/CMakeLists.txt
@@ -97,7 +97,7 @@ if(ENABLE_TESTS)
         
         blt_append_custom_compiler_flag( FLAGS_VAR gtest_extra_flags
                                          DEFAULT " " 
-                                         MSVC  "/wd4251 /wd4251" )
+                                         MSVC  "/wd4251 /wd4275" )
                 
         # Enable builtin google test 
         add_subdirectory(googletest-release-1.8.0)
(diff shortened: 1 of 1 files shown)


thirdparty_builtin/CMakeLists.txt
```diff
@@ -86,17 +86,33 @@ if(ENABLE_TESTS)
     #
     
     if(ENABLE_GTEST)
+        
+        set(gtest_extra_flags "")
+        
+        if(COMPILER_FAMILY_IS_MSVC)
```
kennyweiss: You can use the ``blt_append_custom_compiler_flag`` macro.

E.g. 
```cmake
    FLAGS_VAR gtest_extra_flags
    DEFAULT " " 
    MSVC  "/wd4251 /wd4251"
    )
```

cyrush: except, we only want it to be associated with gtest + gmock -- so we need to have it as part of the info passed to btl register library

kennyweiss: The ``blt_append_custom_compiler_flag`` macro only appends the given flags into the provided variable, and does so in a way that hides the compiler specific cmake logic.  You can then use that variable in the same way, by passing it to the ``blt_register_library`` for gmock and gtest. There are no other side effects.

This has the added benefit of centralizing this logic, so that fixes can be more easily incorporated throughout our build system.  E.g. see issue #81, where we need to add support for the intel compiler through MSVC.

cyrush: Sorry, I misunderstood - I will definitely move to use that.

cyrush:
tested in conduit and this squashed the gtest warnings, ready for review

PR #84, opened 2017-09-06
Use two argument form of add_subdirectory ...

robinson96:
so blt is not required to be  part of the source of the project using blt.

### DIFF
--- a/thirdparty_builtin/CMakeLists.txt
+++ b/thirdparty_builtin/CMakeLists.txt
@@ -87,7 +87,7 @@ if(ENABLE_TESTS)
     
     if(ENABLE_GTEST)
         # Enable builtin google test 
-        add_subdirectory(googletest-release-1.8.0)
+        add_subdirectory(googletest-release-1.8.0 ${PROJECT_BINARY_DIR}/blt/thirdparty_builtin/googletest-release-1.8.0)
 
         blt_register_library(NAME gtest
                              INCLUDES ${gtest_SOURCE_DIR}/include
@@ -105,7 +105,7 @@ if(ENABLE_TESTS)
     if (ENABLE_FORTRAN)
         message(STATUS "Fruit Support is ${ENABLE_FRUIT}")
         if(ENABLE_FRUIT)
-            add_subdirectory(fruit-3.4.1)
+            add_subdirectory(fruit-3.4.1 ${PROJECT_BINARY_DIR}/blt/thirdparty_builtin/fruit-3.4.1)
         endif()
     endif()
 
@@ -115,7 +115,7 @@ if(ENABLE_TESTS)
         endif()
 
         ## google benchmark support
-        add_subdirectory(gbenchmark-master-2017-05-19)
+        add_subdirectory(gbenchmark-master-2017-05-19 ${PROJECT_BINARY_DIR}/blt/thirdparty_builtin/gbenchmark-master-2017-05-19)
 
         if (UNIX AND NOT APPLE)
           find_library(RT_LIBRARIES rt)
--- a/thirdparty_builtin/CMakeLists.txt
+++ b/thirdparty_builtin/CMakeLists.txt
@@ -100,7 +100,7 @@ if(ENABLE_TESTS)
                                          MSVC  "/wd4251 /wd4275" )
                 
         # Enable builtin google test 
-        add_subdirectory(googletest-release-1.8.0 ${PROJECT_BINARY_DIR}/blt/thirdparty_builtin/googletest-release-1.8.0)
+        add_subdirectory(googletest-release-1.8.0 ${BLT_BUILD_DIR}/thirdparty_builtin/googletest-release-1.8.0)
 
         blt_register_library(NAME gtest
                              INCLUDES ${gtest_SOURCE_DIR}/include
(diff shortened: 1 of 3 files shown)


thirdparty_builtin/CMakeLists.txt
```diff
@@ -87,7 +87,7 @@ if(ENABLE_TESTS)
     
     if(ENABLE_GTEST)
         # Enable builtin google test 
-        add_subdirectory(googletest-release-1.8.0)
+        add_subdirectory(googletest-release-1.8.0 ${PROJECT_BINARY_DIR}/blt/thirdparty_builtin/googletest-release-1.8.0)
```
robinson96: Does anyone know a more clever way of identifying the path that I want here other than hardcoding it?

cyrush: @robinson96 can you say a little more what scenario this is needed for? Are there two instances of BLT?

davidbeckingsale:
Two instances of BLT should already work...

cyrush:
Ok i see this is to control where gtest (etc) are built, still knowing  the dir structure ( where blt is relative to source for Peter's case ) would help me ponder potential solutions

cyrush:
1)
For when blt is not in the source dir, we will have to pick a place to anchor where we build things ( ${PROJECT_BINARY_DIR}/blt  (as you have it) is a good choice )

2) 
For when blt is in the source dir, it would be nice to have the relative path it was included as determine blt's build dir. With the current logic the build dir will always be simply `blt `at the root. 
That said, accomplishing this might require more cmake string and path logic than is worth to maintain. 


For  @robinson96's  original question, we could try to wield some cmake string and path stuff to create the right names for the binary dir -- but honesty that may be more confusing and brittle to maintain  than hard coding as you have it now

kennyweiss:
It seems like the ``"blt"``-way would be for us to have a default path encoded into a variable 
(e.g.  ``set(BLT_DIR "." CACHE PATH "")`` ), and user codes can override this with a cache variable in configuration line or host-config.

Then all the ``add_subdirectories()`` commands would have a path like ``${BLT_DIR}/googletest-release-1.8.0``

Incidentally, this might be able to help us with the weird way our blt testing framework is set up, where we add a symbolic link on linux and have to copy the directories on windows.

cyrush:
sounds good, I would call it `BLT_BUILD_DIR` to avoid confusing with BLT_ROOT

robinson96:
Hey all, 
Sorry for the radio silence. I've been out sick. (Still ESL / WAH). 
I think you've figured out my use case from the comments, but let me be clear. 
In my project, which consists of a main project with many TPLs, we are planning on adding a BLT build system for each TPL that does not have one yet. (We've decided we'd rather maintain a BLT build system as a team than have a single individual maintain a Visual Studio build system). 
We don't want each TPL to have its own instance of BLT, but would like them to share a single instance of BLT, e.g., the one in the host project. 
So, directory structure looks something like this: 

hostcode/cmake/blt/...
hostcode/imports/tpl1/CMakeLists.txt
hostcode/imports/tpl2/CMakeLists.txt

 As things are now, when I try to write a tpl/CMakeLists.txt that includes the absolute path to the hostcode/cmake/blt/SetupBLT, all add_subdirectory calls that are of the one argument variety barf because CMAKE insists that the subdirectory be inside the host project dir. 
Giving the second argument resolves this issue. 

I will update to use BLT_BUILD_DIR as suggested by @kennyweiss  and @cyrush .

cyrush:
@robinson96 -- we needed this as part of our testing for docs, 

https://github.com/LLNL/blt/pull/85/commits/4bb93ca793ef8e1419dda8234a6f73b60d1ed3e8

thirdparty_builtin/CMakeLists.txt
```diff
@@ -100,7 +100,7 @@ if(ENABLE_TESTS)
                                          MSVC  "/wd4251 /wd4275" )
                 
         # Enable builtin google test 
-        add_subdirectory(googletest-release-1.8.0)
+        add_subdirectory(googletest-release-1.8.0 ${BLT_BUILD_DIR}/thirdparty_builtin/googletest-release-1.8.0)
```
kennyweiss: Since the gtest and fruit directories have the release information in them, does it makes sense to extract that out into a variable rather than duplicating the path?
E.g.
```cmake 
set(gtest_dir googletest-release-1.8.0)
add_subdirectory(${gtest_dir} ${BLT_BUILD_DIR}/thirdparty_builtin/${gtest_dir}) 
```
... and similarly for fruit and gbenchmark.

thirdparty_builtin/gbenchmark-master-2017-05-19/CMakeLists.txt
```diff
@@ -194,9 +194,9 @@ find_package(Threads REQUIRED)
 include_directories(${PROJECT_SOURCE_DIR}/include)
 
 # Build the targets
-add_subdirectory(src)
+add_subdirectory(src ${BLT_BUILD_DIR}/thirdparty_builtin/gbenchmark-master-2017-05-19/src)
```
kennyweiss: This is a file that is part of the gbenchmark repo.

Is there any way to avoid modifying the CMakeLists file, so we don't have to patch up gbenchmark when we upgrade?  E.g. since you already added the ``add_subdirectory`` call in the parent directory, is it possible that these two calls will work without the second argument?

cyrush:
@robinson96  -- we merged the docs branch to master, which includes the similar fix.  can you take it for a spin?

robinson96:
new changes in master does fix this issue. Closing this PR.

Issue #65, opened 2017-07-26
current lua and python support are Axom specifc


cyrush:
Should be moved into Axom.

cyrush:
@white238  is looking into this

Issue #51, opened 2017-07-14
add_sphinx_target needs to be reworked to follow BLT standards


white238:
* should be prefixed with blt_
* ${deps} is referenced but not set (should be an argument DEPENDS_ON)
* change sphinx_target_name to "NAME foo"

PR #85, opened 2017-09-10
add tutorial documentation

cyrush:
add blt tutorial documentation and examples


cyrush:
@white238 @kennyweiss creating a PR, so we can use it to collab for the tutorial docs. 

A few notes:

1.  I added a cuda example, however it requires a gpu w/ compute_60 or greater. Basically a pascal card. This won't run on surface.  I have a work around that may work, but surface's cards may still be too old.

2. I took a pass at modifying peter's solution for an out-of-source blt, to support a `BLT_BUILD_DIR`.  With my changes, I was also able to use `BLT_ROOT_DIR` to include blt via a path passed on the cmake command line

      In the main CMakeLists.txt, I used: 
      `include(${BLT_ROOT_DIR}/SetupBLT.cmake)`

      And when calling cmake via the command line, I used:
       `-DBLT_ROOT_DIR=/Users/harrison37/Work/blt_work/github/blt/`

      It's great for testing, will also make our ci setup much easier. 

      I haven't committed this yet b/c it will conflict with peter's solution.
      Also hung up on if we should change `BLT_ROOT_DIR` to `BLT_SOURCE_DIR`

3. You can symlink blt into the docs/tutorial examples -- if run into problems, I would double check the symlink is to an abs path.

cyrush:
Update:

for the calc_pi example, I aded:
- doxygen and sphinx examples
- a (potential) fix for the `atomicAdd` for older cuda archs

I also committed the fix for using an out-of-source blt, it's peter's solution + an optional `BLT_BUILD_DIR`. 

I think we are very close to being done with the actual example code, but we still need to integrate into the tutorial docs

SetupBLT.cmake
```diff
@@ -45,6 +45,11 @@ if (NOT BLT_LOADED)
   mark_as_advanced(BLT_LOADED)
 
   set( BLT_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR} CACHE PATH "" FORCE )
+  
+  # if an explicit build dir was not specified, set a default.
+  if( NOT BLT_BUILD_DIR )
```
kennyweiss: #84 writes this more simply as:
```cmake
 set( BLT_BUILD_DIR ${PROJECT_BINARY_DIR}/blt CACHE PATH "" ) 
```
Note: The ``if()`` check is missing, as is the ``FORCE`` keyword.
I checked the documentation to make sure this is correct.

from https://cmake.org/cmake/help/v3.8/command/set.html:
> Set the given cache <variable> (cache entry). Since cache entries are meant to provide user-settable values this does not overwrite existing cache entries by default. Use the FORCE option to overwrite existing entries.

robinson96:
Should I abandon efforts on #84? If so, you might want to bring some of Kenny's suggestions regarding patch files over to this branch.

SetupBLT.cmake
cyrush: Yes #84 is simpler, but it doesn't allow you to actually pass in your own BLT_BUILD_DIR.

I thought we wanted to support a user settable build loc?

cyrush:
@robinson96, yes I think so. Sorry for the hassle.

I wanted to work to make the changes on #84 -- but the fork made it hard for me to do so directly.

SetupBLT.cmake
kennyweiss: Yes -- both versions should allow you to pass in a user-settable parameter.  @robinson96's version is shorter. 

Not sure if I was clear above.  According to the docs, 
```cmake
 set( BLT_BUILD_DIR ${PROJECT_BINARY_DIR}/blt CACHE PATH "" ) 
```
should do the following:
* if ``BLT_BUILD_DIR`` is not already a cache variable, it will add it to the cache and set its value to ``${PROJECT_BINARY_DIR}/blt``
* if it is already a cache variable, the value of ``BLT_BUILD_DIR`` will not be modified by this call (since the statement is missing the ``FORCE`` keyword)

It should have the same effect as your three liner.

kennyweiss: Here is a simple example to corfirm

```cmake
# -- cache.cmake
set(FOO "foo-cache-value" CACHE PATH "")
set(BAR "bar-cache-value" CACHE PATH "")
set(BAZ "baz-cache-value" CACHE PATH "")

#-- CMakeLists.txt
project(test-var)

foreach(v FOO BAR BAZ)
  message(STATUS "${v} before -- ${${v}}")
endforeach()

set(FOO "foo-in-cmake-lists" CACHE PATH "" )
set(BAR "bar-in-cmake-lists" CACHE PATH "" FORCE)
set(BAZ "baz-in-cmake-lists" )

foreach(v FOO BAR BAZ)
  message(STATUS "${v} after -- ${${v}}")
endforeach()
```
Running ``cmake -C ../cache.cmake`` yields the following
```bash
-- FOO after -- foo-cache-value
-- BAR after -- bar-in-cmake-lists    # overridden due to CACHE FORCE
-- BAZ after -- baz-in-cmake-lists    # overridden due to missing CACHE
```

robinson96:
Hmm I had to fork because I don't have write permissions to this repo.  Should I / can I get them?

SetupBLT.cmake
cyrush: got it - I did not realize that was the case.  I guess you can tell I don't override things using that mechanism.  Do we do this anywhere else in BLT?

The behavior for the 3 line form seems much clearer to me - what do other folks think?

Also, here is some more info about what happens when things are already set:

>  It is possible for the cache entry to exist prior to the call but have no type set if it was created on the cmake(1) command line by a user through the -D<var>=<value> option without specifying a type. In this case the set command will add the type. Furthermore, if the <type> is PATH or FILEPATH and the <value> provided on the command line is a relative path, then the set command will treat the path as relative to the current working directory and convert it to an absolute path.

I have been always using as paths, but is that the logic we want for relative paths? 
Sounds confusing ...

cyrush:
@robinson96, sorry about that -- @white238 is the gatekeeper.

white238:
@robinson96 you should have access now.

cyrush:
@white238 , should we try to address #51  

or at least change -- `add_sphinx_target`, to `blt_add_sphinx_target` to future proof the docs some what?

white238:
@cyrush changed the names to atleast be blt_*.

cyrush:
I also @white238 changed the `NUM_PROCS` arg for `blt_add_test` to `NUM_MPI_TASKS`,
to address part of #2

white238:
I like the NUM_MPI_TASKS change..

white238:
@kennyweiss did you test the removing of `pwd` from the BLT_SOURCE_DIR? I got errors until i added that specifically.

kennyweiss:
@white238 

Yes -- it works for me and is how I always do this.
I was trying to simplify the writeup and didn't realize it was an issue.  

Should I revert?

Here is what I typed:

```bash
>pwd
.../blt/docs/tutorial/calc_pi
>mkdir build
>cd build
>cmake -DBLT_SOURCE_DIR=../../.. ..
-- The C compiler identification is GNU 4.9.3
-- The CXX compiler identification is GNU 4.9.3
...
-- Configuring done
-- Generating done
-- Build files have been written to: .../blt/docs/tutorial/calc_pi/build
```

kennyweiss:
@cyrush and @white238 -- How do you feel about adding a host-config and instructions for a blue-os system?  I have a working host-config.

cyrush:
@kennyweiss, yes lets to ray?

kennyweiss:
@cyrush -- great -- adding it.

cyrush:
@kennyweiss -- improvements look awesome, thanks!

cyrush:
Final updates, I also removed unused lua and python support in the add lib macro to address #65

PR #86, opened 2017-09-14
Remove option to macro

white238:
This causes problems when you want to wrap the macro and pass the arguments.  There is no way to do this with option arguments.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -258,7 +258,7 @@ endmacro(blt_register_library)
 ##                  OUTPUT_NAME [name]
 ##                  OUTPUT_DIR [dir]
 ##                  HEADERS_OUTPUT_SUBDIR [dir]
-##                  SHARED
+##                  SHARED [TRUE | FALSE]
 ##                 )
 ##
 ## Adds a library to the project composed by the given source files.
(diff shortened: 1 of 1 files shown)


cmake/BLTMacros.cmake
```diff
@@ -258,15 +258,15 @@ endmacro(blt_register_library)
...
+##                  SHARED [TRUE | FALSE]
 ##                 )
 ##
 ## Adds a library to the project composed by the given source files.
 ##
 ## Adds a library target, called <libname>, to be built from the given sources.
 ## This macro internally checks if the global option "ENABLE_SHARED_LIBS" is
 ## ON, in which case, it will create a shared library. By default, a static
-## library is generated unless the SHARED option is added.
+## library is generated unless the SHARED to TRUE.
```
kennyweiss: I think this writeup can be clarified a bit.  
It implies that the default buiild type is static unless ``SHARED`` is added.
A more accurate description might be that the default build type (shared|static) is  based on the value of ``ENABLE_SHARED_LIBS``, and that the ``SHARED`` argument can be used to override this default.  

How about replacing the last three lines with something like this:

> This macro uses the ``ENABLE_SHARED_LIBS`` to determine whether the library will be build as shared or static.  The optional boolean ``SHARED`` argument can be used to override this choice.

PR #88, opened 2017-09-15
General Macro clean up

white238:
This is a pull request for some general things that i have been putting off doing:

* Remove copy headers,
* Add inherited INCLUDES and DEFINES to blt_add_library
* Add more functionality needed for uncrustify targets


white238:
After doing these changes I realized we lack some testing on a few things.  I am going to attempt to add some more rigorous testing early next week before I merge this.

cmake/BLTMacros.cmake
```diff
@@ -48,13 +48,15 @@ include(${BLT_ROOT_DIR}/cmake/BLTPrivateMacros.cmake)
...
 ##
-## Adds pre-processor definitions to a particular. This macro provides very
+## Adds pre-processor definitions to a particular target. This macro provides very
 ## similar functionality to cmake's native "add_definitions" command, but,
 ## it provides more fine-grained scoping for the compile definitions on a
 ## per target basis. Given a list of definitions, e.g., FOO and BAR, this macro
 ## adds compiler definitions to the compiler command for the given target, i.e.,
 ## it will pass -DFOO and -DBAR.
 ##
+## These compile definitions are not inherited to any dependent targets.
```
kennyweiss: Does it make sense to add an option (or single value arg) to allow the flags to propagate?

cmake/BLTMacros.cmake
```diff
@@ -290,7 +287,7 @@ endmacro(blt_register_library)
 macro(blt_add_library)
 
     set(singleValueArgs NAME OUTPUT_NAME OUTPUT_DIR HEADERS_OUTPUT_SUBDIR SHARED)
-    set(multiValueArgs SOURCES HEADERS DEPENDS_ON)
+    set(multiValueArgs SOURCES HEADERS INCLUDES DEFINES DEPENDS_ON)
```
kennyweiss: Does it make sense to pull out some of this logic into a private macro, e.g. if it duplicates logic that is in ``blt_register_library``?

cmake/BLTOptions.cmake
```diff
@@ -53,7 +53,6 @@ option(ENABLE_COVERAGE   "Enables code coverage support" OFF)
 ################################
 # Build Options
 ################################
-option(ENABLE_COPY_HEADERS "Copy headers to build directory" OFF)
```
kennyweiss: I appreciate the simplicity that comes from ripping this feature out, but do we have workarounds for this feature in all codes that were using ``COPY_HEADERS``?

Should we check that these workarounds work (e.g. by disabling ``COPY_HEADERS``) before removing the feature from blt?

cmake/SetupCodeChecks.cmake
```diff
@@ -58,15 +58,23 @@ endif()
...
+    set(multiValueArgs)
+    
+    cmake_parse_arguments(arg
+        "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN} )
 
     if(UNCRUSTIFY_FOUND)
         # Only run uncrustify on C and C++ files
         # Note, we can later extend this by passing in a list of valid types to the macro
-        set(_fileTypes ".cpp" ".hpp" ".c" ".h")
+        set(_fileTypes ".cpp" ".hpp" ".cxx" ".hxx" ".cc" ".c" ".h")
```
kennyweiss: If we're adding all these, we should also add ``.hh``

cmake/SetupCodeChecks.cmake
```diff
@@ -75,67 +83,93 @@ macro(blt_add_code_check_targets cfg_file)
...
     cmake_parse_arguments(arg
         "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN} )
 
+    if(${arg_WORKING_DIRECTORY})
+        set(_wd ${arg_WORKING_DIRECTORY})
+    else()
+        set(_wd ${CMAKE_CURRENT_SOURCE_DIR})
+    endif()
+
     add_custom_target("uncrustify_check_${PROJECT_NAME}"
```
kennyweiss: Using ``PROJECT_NAME`` means that users can only set up one uncrustify target per ``project()``. 
Are there build systems where users only use the ``project()`` command once, but would want more than one uncrustify target?

cmake/SetupCodeChecks.cmake
```diff
@@ -75,67 +83,93 @@ macro(blt_add_code_check_targets cfg_file)
...
+##                           WORKING_DIRECTORY <working_directory>
+##                           REDIRECT <redirection_commands>
 ##                           SRC_FILES <list_of_src_files_to_uncrustify> )
 ##
 ##------------------------------------------------------------------------------
 macro(blt_add_uncrustify_check)
-    
-    MESSAGE(STATUS "Creating uncrustify check target: uncrustify_check_${PROJECT_NAME}")
+        
+    message(STATUS "Creating uncrustify check target: uncrustify_check_${PROJECT_NAME}")
```
kennyweiss: I think we should pull the target name out into its own variable e.g.
```cmake
set(uncrustify_target "uncrustify_check_${PROJECT_NAME}")
```

cmake/BLTMacros.cmake
white238: I think these three macros need some discussion.

cmake/BLTMacros.cmake
white238: My idea was to get blt_add_library fully flushed out then make blt_register_library only for external libraries when I can get the imported targets working.  The mixing of the two idioms is will not work in the future I would expect.  Not to mention the confusion I think it creates.  Maybe the name needs to change.  Like blt_import_target

cmake/BLTOptions.cmake
white238: AXOM is the only code I know that uses copy headers.

cmake/SetupCodeChecks.cmake
white238: Yea that's not a very general purpose thing to do.  Maybe default to project if one isn't given

cmake/BLTOptions.cmake
kennyweiss: I meant that we should have a working solution there with ``ENABLE_COPY_HEADERS`` disabled before we remove it from blt.

cmake/SetupCodeChecks.cmake
kennyweiss: In the same spirit, this macro should also take a list of source files as an input, rather than implicitly using ``${PROJECT_NAME}_ALL_SOURCES}``

cmake/BLTOptions.cmake
cyrush: @kennyweiss, I think we agree any `COPY_HEADERS` solution will be in Axom proper, so I am not sure that should hold up work on BLT master.

cmake/BLTMacros.cmake
```diff
@@ -266,11 +260,14 @@ endmacro(blt_register_library)
...
-## 
+## The INCLUDES argument allows you to define what include directories are
+## needed by any target that is dependent on this library.  These will
+## be inherited by CMake's target dependency rules.
+##
+## The DEFINES argument allows you to add needed compiler definitions that are
+## needed by any target that is dependent on this library.  These will
+## be inherited by CMake's target dependency rules.
+##
 ## If given a DEPENDS_ON argument, it will add the necessary includes and
```
cyrush: Could was also add `COMPILE_FLAGS`, so we can smooth out CUDA support?

white238:
Closing this pull request due to all changes, except ripping out copy headers, are now moved to other pull requests.

PR #89, opened 2017-09-15
add ci testing of tutorial example projects

cyrush:


cyrush:
@white238 @kennyweiss added testing of the two example projects to travis and appveyor, ready for review.

cyrush:
yes @kennyweiss we definitely want the original tests, they will test much more than what we will expose as part of the tutorial. 

We can move to use the BLT_SOURCE_DIR, but some copying is necessary (due to the git tests). I didn't touch it as part of this PR b/c I didn't want too many moving parts while things were in a broken state.

.travis.yml
```diff
@@ -125,13 +125,12 @@ install:
...
   - git config --global user.email "blt-dev@llnl.gov" 
-  - git config --global user.name  "BLT Robot"    
+  - git config --global user.name  "BLT Robot"
+  - BLT_SOURCE_DIR=${TRAVIS_BUILD_DIR}
   - cmake --version
   - cd $TRAVIS_BUILD_DIR/..
   # copy our internal testing project
   - cp -r blt/cmake/blt-test .
-  # symlink blt into this project
-  - ln -s ${TRAVIS_BUILD_DIR} blt-test/blt
```
kennyweiss: Thanks -- the symbolic link with recursive directories was awkward.

cyrush: no problem, glad we can stop & avoid fun with sphinx infinity traversing the recursive symlink.

PR #91, opened 2017-09-18
NVCC flag support in blt_append_custom_compiler_flag

robinson96:
If the NVCC argument is defined and CUDA is enabled, blt_append_custom_compiler_flag will now append to the CUDA_NVCC_FLAGS list.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -629,7 +629,7 @@ endmacro(blt_add_benchmark)
 macro(blt_append_custom_compiler_flag)
 
    set(options)
-   set(singleValueArgs FLAGS_VAR DEFAULT GNU CLANG INTEL XL MSVC)
+   set(singleValueArgs FLAGS_VAR DEFAULT GNU CLANG INTEL XL MSVC NVCC)
    set(multiValueArgs)
 
    # Parse the arguments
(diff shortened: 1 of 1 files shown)


cmake/BLTMacros.cmake
```diff
@@ -656,5 +656,10 @@ macro(blt_append_custom_compiler_flag)
       set (${arg_FLAGS_VAR} "${${arg_FLAGS_VAR}} ${arg_DEFAULT} ")
    endif()   
 
+   # append any NVCC flags to CUDA_NVCC_FLAGS if CUDA is enabled
+   if ( DEFINED arg_NVCC AND ENABLE_CUDA)
+       list (APPEND CUDA_NVCC_FLAGS ${arg_NVCC})
```
davidbeckingsale: I'm not sure this will work exactly how you think. `FindCUDA` sets up the flags as an empty variable in the cache, which causes all kinds of problems.

cyrush:
Setting the specific NVCC cmake var doesn't really fit here. 

This macro is supposed to append the flags to a CMake variable passed as input to the macro. Then that flag can be used as input to other things. 

While it could be used to set special cmake vars (CMAKE_CXX_FLAGS, or CUDA_NVCC_FLAGS), that would have to be done by the user by providing them as input. 

I think we are going to add COMPILE_FLAGS to blt_add_target to help with the case you are trying to tackle. 

FindCUDA copies global settings from CMake vars when setting up each target, so you have to do set / unset global vars dance to get things propagated.

robinson96:
Hey guys. Points taken. Closing this, refactoring my use case to be limited to my build system.

PR #92, opened 2017-09-19
Add DEPENDS_ON to multiValue args for parsing in blt_register_library

davidbeckingsale:
DEPENDS_ON was not parsed out of the macro arguments, hence was always empty.


white238:
The joys of CMake's "language"

PR #93, opened 2017-09-20
only apply cmake rpath settings when creating shared libs

cyrush:
avoid setting cmake rpath settings when building shared libs.

Avoids install errors like the following:

```
CMake Error at libs/relay/cmake_install.cmake:62 (file):
  file RPATH_CHANGE could not write new RPATH:

    /project/projectdirs/.../conduit/install/lib

  to the file:

    /project/projectdirs/.../conduit/install/bin/conduit_relay_node_viewer

  No valid ELF RPATH or RUNPATH entry exists in the file;
  libs/cmake_install.cmake:38 (include)
  cmake_install.cmake:60 (include)
```

### DIFF
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -205,7 +205,7 @@ endif()
 # RPath Settings
 ################################
 # only apply rpath settings for builds using shared libs
-if(BUILD_SHARED_LIBS)
+if(ENABLE_SHARED_LIBS)
     # use, i.e. don't skip the full RPATH for the build tree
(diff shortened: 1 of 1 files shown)


white238:
Hopefully this doesn't cause problems when one of your libraries is shared.. It doesn't look like it should but i will know when i update my code.

cyrush:
@white238, hope so as well.  If so, we will have to think of away to deal with using both shared and static -- it would be nice if CMake didn't complain when there is no rpath b/c all deps are static. 

On the flip side, (using shared w/o rpath settings)  I think it would simply not use rpath, which might be a problem for using an installed binary -- but i don't think it would cause an error during the install process.

PR #96, opened 2017-09-26
add page that describes recommended cmake practices  

cyrush:
work in progress, wanted to capture a few I as I was recalling our approach for blt and other projects


cyrush:
Adds some more docs that encompass some of our CMake practices, ready for review.

PR #98, opened 2017-10-04
Add option for turning off Find mpi

white238:
This is useful on systems where find_package(MPI) just doesn't seem to return the correct information.  This allows you to compile with the just the mpi wrapper.

### DIFF
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -58,6 +58,8 @@ option(ENABLE_FORTRAN      "Enables Fortran compiler support" OFF)
 option(ENABLE_SHARED_LIBS  "Enables shared libraries." OFF)
 
 option(ENABLE_MPI          "Enables MPI support" OFF)
+option(ENABLE_FIND_MPI     "Enables CMake's Find MPI support (Turn off when compiling with the mpi wrapper directly)" ON)
+
 option(ENABLE_OPENMP       "Enables OpenMP compiler support" OFF)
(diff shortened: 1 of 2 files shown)


cmake/BLTOptions.cmake
```diff
@@ -58,6 +58,8 @@ option(ENABLE_FORTRAN      "Enables Fortran compiler support" OFF)
 option(ENABLE_SHARED_LIBS  "Enables shared libraries." OFF)
 
 option(ENABLE_MPI          "Enables MPI support" OFF)
+option(ENABLE_FIND_MPI     "Enables CMake's Find MPI support (Turn off when compiling with the mpi wrapper directly)" ON)
```
kennyweiss: This should be an advanced option since most users will not have to care about this.

kennyweiss: Side note: We should think about how we name sub-options for how we use enabled packages, since we are starting to get a lot of options.

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -44,7 +44,10 @@
 # MPI
 ################################
 
-find_package(MPI REQUIRED)
+if (ENABLE_FIND_MPI)
+    find_package(MPI REQUIRED)
+endif()
+
 message(STATUS "MPI C Compile Flags:  ${MPI_C_COMPILE_FLAGS}")
```
white238: Should the compile, include, link, and libraries message lines be in the if guard as well?

kennyweiss:
Looks good

PR #102, opened 2017-10-13
Added blt_find_libraries in order to search for multiple libraries

rrsettgast:
The cmake macro/function "find_library" only returns the first library found when specifying multiple libraries to search for. This macro returns a list of all found libraries, and generates an error if one of the libraries is not found.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -667,7 +667,7 @@ endmacro(blt_append_custom_compiler_flag)
 
 
 ##------------------------------------------------------------------------------
-## blt_find_libraries( <FOUND_LIBS>
+## blt_find_libraries( FOUND_LIBS <FOUND_LIBS>
 ##                     NAMES [libname1 [libname2 ...]]
 ##                     PATHS [path1 [path2 ...]] )
 ##
@@ -681,7 +681,10 @@ endmacro(blt_append_custom_compiler_flag)
 ##
 ##
 ##------------------------------------------------------------------------------
-macro(blt_find_libraries FOUND_LIBS)
+macro(blt_find_libraries)
+
+    set(singleValueArgs FOUND_LIBS )
+
 
     set(multiValueArgs NAMES
                        PATHS )
@@ -713,6 +716,6 @@ macro(blt_find_libraries FOUND_LIBS)
             message(FATAL_ERROR "NAMES entry ${lib} not found. These are not the libs you are looking for.")
         endif()
     endforeach()
-    set( ${FOUND_LIBS} ${TEMP_LIBS} )
+    set( ${arg_FOUND_LIBS} ${TEMP_LIBS} )
 
 endmacro(blt_find_libraries)
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -668,14 +668,17 @@ endmacro(blt_append_custom_compiler_flag)
 
 ##------------------------------------------------------------------------------
 ## blt_find_libraries( FOUND_LIBS <FOUND_LIBS>
-##                     NAMES [libname1 [libname2 ...]]
+##                     REQUIRED_NAMES [libname1 [libname2 ...]]
+##                     OPTIONAL_NAMES [libname1 [libname2 ...]]
 ##                     PATHS [path1 [path2 ...]] )
 ##
 ## This command is used to find a list of libraries. A cache entry named by <FOUND_LIBS> 
 ## is created to store the result of this command. If the libraries are found the
 ## results are stored in FOUND_LIBS.
 ##
-## NAMES lists the names of the libraries to search for.
+## REQUIRED_NAMES lists the names of the libraries that are required to be found.
+##
+## OPTIONAL_NAMES lists the names of the optional libraries to be found.
 ##
 ## PATH lists the paths in which to search for NAMES.
 ##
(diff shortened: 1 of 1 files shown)


cmake/BLTMacros.cmake
```diff
@@ -663,3 +663,56 @@ macro(blt_append_custom_compiler_flag)
...
+        message(FATAL_ERROR "blt_find_libraries requires that NAMES specify the library names you are searching for")
+    endif()
+
+    if ( NOT DEFINED arg_PATHS )
+        message(FATAL_ERROR "blt_find_libraries requires that PATHS specify the paths to search for NAMES")
+    endif()
+
+    foreach( lib ${arg_NAMES} )
+        unset( temp CACHE )
+        find_library( temp NAMES ${lib}
```
white238: It would be nice if this macro also handled foo and libfoo since that is used so often.

cmake/BLTMacros.cmake
```diff
@@ -663,3 +663,56 @@ macro(blt_append_custom_compiler_flag)
...
+## is created to store the result of this command. If the libraries are found the
+## results are stored in VAR.
+##
+## NAMES lists the names of the libraries to search for.
+##
+## PATH lists the paths in which to search for NAMES.
+##
+##
+##------------------------------------------------------------------------------
+macro(blt_find_libraries VAR)
```
white238: While CMake does this on a regular basis, having unnamed vars causes problems with wrapping macros.  Please name the VAR variable, like OUTPUT or whatever you like.

white238:
This will clean up a lot of my code. Definitely one of those functions of things I should have thought of....

cmake/BLTMacros.cmake
rrsettgast: so this I assume the best way to do this is to check to see if there is a "lib" in front of "foo" and strip it out prior to the call do camke::find_library()? Is it safe to just strip the characters from the name? i.e. is there something that might be called "liblibfoo" and it is entered as "libfoo" then the lib will get stripped off and it won't be found.

white238: Cyrus claims it does the correct thing already.  This might need to be tested.  I will do so now

I was more saying that it should add both the base name and the lib+base name to this find_library call.  Having extra names that don't match wont cause problems..

 For example:

set(_prependedlib "lib${lib}")
find_library( temp NAMES ${lib} ${_prependedlib}

white238: I apparently don't know what I am talking about. I could have sworn you needed both on the function.  But I just tested and if you add "foo libfoo" it works cuz it finds "foo" but if you do just "libfoo" it fails to find it.  So this is a non issue apparently.

rrsettgast: OK. I am going to leave it as is then.

cmake/BLTMacros.cmake
rrsettgast: I renamed this to FOUND_LIBS.

cmake/BLTMacros.cmake
```diff
@@ -663,3 +663,56 @@ macro(blt_append_custom_compiler_flag)
...
+##
+## NAMES lists the names of the libraries to search for.
+##
+## PATH lists the paths in which to search for NAMES.
+##
+##
+##------------------------------------------------------------------------------
+macro(blt_find_libraries FOUND_LIBS)
+
+    set(multiValueArgs NAMES
```
kennyweiss: Does it make sense to have a set of required ``NAMES`` and a set of ``OPTIONAL_NAMES``?  The way this is written, if the library is not found, the configuration will fail.

Alternatively, does it make sense to add a parameter to make the results optional, similar to how the ``REQUIRED`` keyword works in cmake`s ``find(FOO REQUIRED)``?

cmake/BLTMacros.cmake
```diff
@@ -663,3 +663,56 @@ macro(blt_append_custom_compiler_flag)
...
+        unset( temp CACHE )
+        find_library( temp NAMES ${lib}
+                      PATHS ${arg_PATHS}
+                      NO_DEFAULT_PATH
+                      NO_CMAKE_ENVIRONMENT_PATH
+                      NO_CMAKE_PATH
+                      NO_SYSTEM_ENVIRONMENT_PATH
+                      NO_CMAKE_SYSTEM_PATH)
+        if( temp )
+            set( TEMP_LIBS ${TEMP_LIBS} ${temp} )
```
kennyweiss: This could be changed to ``list(APPEND TEMP_LIBS ${temp})``

cmake/BLTMacros.cmake
```diff
@@ -663,3 +663,56 @@ macro(blt_append_custom_compiler_flag)
...
+## is created to store the result of this command. If the libraries are found the
+## results are stored in FOUND_LIBS.
+##
+## NAMES lists the names of the libraries to search for.
+##
+## PATH lists the paths in which to search for NAMES.
+##
+##
+##------------------------------------------------------------------------------
+macro(blt_find_libraries FOUND_LIBS)
```
kennyweiss: Could you please add an example of the usage of this to the blt tests?
I suppose it should be for a system library that should be available on all systems (or guarded if it is only for unix-based systems).

cmake/BLTMacros.cmake
kennyweiss: A different useful feature, somewhat in line with @white238's request for foo and libfoo would be if you want to find at least one library in the list, 

A (possible bad) example would be if you want to find ``foo`` or ``foo.so.6``.

cmake/BLTMacros.cmake
```diff
@@ -663,3 +663,56 @@ macro(blt_append_custom_compiler_flag)
 
 endmacro(blt_append_custom_compiler_flag)
 
+
+
+
+##------------------------------------------------------------------------------
+## blt_find_libraries( <FOUND_LIBS>
```
white238: Sorry I wasn't more clear.  This needs to be changed to:

blt_find_libraries(FOUND_LIBS <FOUND_LIBS>

Unnamed variables are a large hassle to parse if someone wants to wrap this macro.

cmake/BLTMacros.cmake
white238: I agree an optional list would be very useful.  I would on the other hand not do a required keyword.

cmake/BLTMacros.cmake
rrsettgast: should it be changed?

cmake/BLTMacros.cmake
rrsettgast: I can do this, but I am wondering what would be the advantage of this option? For instance, I am using this to find specific libraries, and if they are not found then cmake gives an error. If they are found, things link fine. If there is an optional library you won't know you missed it until you get a link error? I would think this is worse? Perhaps I am missing something here?

rrsettgast: I changed NAMES to REQUIRED_NAMES and OPTIONAL_NAMES. Failure to find any of the REQUIRED_NAMES results in a FATAL_ERROR. Failure to find any of the OPTIONAL_NAMES prints a warning.

PR #104, opened 2017-10-25
Slight macro clean up and update developers

white238:


cmake/BLTMacros.cmake
```diff
@@ -664,65 +664,51 @@ macro(blt_append_custom_compiler_flag)
...
 
-
-
 ##------------------------------------------------------------------------------
-## blt_find_libraries( FOUND_LIBS <FOUND_LIBS>
-##                     REQUIRED_NAMES [libname1 [libname2 ...]]
-##                     OPTIONAL_NAMES [libname1 [libname2 ...]]
+## blt_find_libraries( FOUND_LIBS <FOUND_LIBS variable name>
+##                     NAMES [libname1 [libname2 ...]]
+##                     OPTIONAL [TRUE | FALSE ]
```
kennyweiss: Can we make the keyword be ``REQUIRED`` instead of ``OPTIONAL``, with a default of ``TRUE`` ?
This seems like a better fit with cmake's ``find_package`` -- https://cmake.org/cmake/help/v3.8/command/find_package.html -- although, in that case, it is an option, and the default is OFF.

cmake/SetupCodeChecks.cmake
```diff
@@ -99,15 +104,22 @@ macro(blt_add_uncrustify_check)
...
     cmake_parse_arguments(arg
         "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN} )
 
+    if(${arg_WORKING_DIRECTORY})
+        set(_wd ${arg_WORKING_DIRECTORY})
+    else()
+        set(_wd ${CMAKE_CURRENT_SOURCE_DIR})
+    endif()
+
     add_custom_target("uncrustify_check_${PROJECT_NAME}"
```
kennyweiss: It would be nice if there were an option to set the name of this target rather than using the cmake project.

cmake/SetupCodeChecks.cmake
```diff
@@ -99,15 +104,22 @@ macro(blt_add_uncrustify_check)
...
-            ${UNCRUSTIFY_EXECUTABLE}
-            -c ${CMAKE_CURRENT_SOURCE_DIR}/${arg_CFG_FILE} --check ${arg_SRC_FILES}
-             COMMENT "Running uncrustify source code formatting checks.")
+            ${UNCRUSTIFY_EXECUTABLE} ${arg_FLAGS}
+            -c ${arg_CFG_FILE} --check ${arg_SRC_FILES} ${arg_REDIRECT}
```
kennyweiss: Can you please a line to the documentation about what ``REDIRECT`` does?

cmake/BLTMacros.cmake
```diff
@@ -676,16 +676,16 @@ endmacro(blt_append_custom_compiler_flag)
 ##
 ## NAMES lists the names of the libraries that will be searched for in the given PATHS.
 ##
-## If OPTIONAL is set to TRUE, BLT will not produce an error message if any of the
-## given libraries that are not found.  The default value is FALSE.
+## If REQUIRED is set to TRUE, BLT will produce an error message if any of the
+## given libraries that are not found.  The default value is TRUE.
```
kennyweiss: typo:  
> ...if any of the given libraries \*are\* not found. ...

cmake/SetupCodeChecks.cmake
```diff
@@ -45,122 +45,93 @@
...
-##                           REDIRECT <redirection_commands>
-##                           SRC_FILES <list_of_src_files_to_uncrustify> )
+## FLAGS are additional flags given to uncrustify.
+##
+## COMMENT is prepended to the commented outputted by CMake.
+##
+## WORKING_DIRECTORY is the directory that uncrustify will be ran.  It defaults to the directory
+## where this macro is called.
+##
+## REDIRECT is appended to the uncrustify command and allows you to redirect output to a file.
```
kennyweiss: Thanks for adding this.  Should ``REDIRECT`` just be ``EXTRA_COMMANDS``  or ``ADDL_OPTIONS`` or something like that?  You're basically just adding it to the uncrustify invocation.

white238:
@rrsettgast I know we discussed this and asked you to make that change but after further discussion we went with this.  Mainly because you run into ordering problems if the two lists are dependent on each other and this simplifies the normal case which if you are asking for a library you already know its required.

rrsettgast:
@white238 I hadn't thought about the ordering issue between the two lists. Good preemptive catch.

Issue #107, opened 2017-10-29
Question/Proposed enhancement


zbeekman:
Is there a reason you're not using the [`CMakeDependentOption`](https://cmake.org/cmake/help/v3.0/module/CMakeDependentOption.html) CMake module for handling option dependencies?

i.e., `ENABLE_GMOCK` depends on `ENABLE_GTEST`,  `ENABLE_FRUIT` depends on `ENABLE_FORTRAN` etc.

It would be easy to work up a PR to include this capability. CMake support goes as far back as at least 3.0 if not earlier.

kennyweiss:
Thanks for the suggestion.  I wasn't aware of this macro.  
This seems like it could clean up some of our internal logic.

zbeekman:
Let me know if a contribution here is welcome or if you'd prefer to look at this on your own.

kennyweiss:
If you're interested, we'd definitely welcome the contribution.

Issue #99, opened 2017-10-12
Would be nice if blt used BUILD_SHARED_LIBS


tepperly:
It would be nice if blt supported [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/v3.4/variable/BUILD_SHARED_LIBS.html?highlight=build_shared_libs]) instead of requiring -DENABLE_SHARED_LIBS. Or perhaps, the default for ENABLE_SHARED_LIBS should come from BUILD_SHARED_LIBS instead of being false.

cyrush:
@tepperly, I agree BUILD_SHARED_LIBS seems to be more widely used. 

For conduit, I work around this with the following logic before including blt:

```
option(BUILD_SHARED_LIBS  "Build shared libraries"      ON)
# forward conduit's build shared libs option to blt
set(ENABLE_SHARED_LIBS ${BUILD_SHARED_LIBS} CACHE BOOL "")

```

zbeekman:
I'm going to submit a PR for this. I won't be offended if you mark as won't fix or want to handle it some other way.

davidbeckingsale:
👍

PR #108, opened 2017-10-29
Fortran fixes

### LINKED ISSUE #105: BLT breaks simple Fortran w/ CMake 3.8
When I include blt in my CMakeLists.txt I get the following error:

```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_CXX_COMPILE_OBJECT
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_CXX_LINK_EXECUTABLE
```

I cloned blt from master at 33b5245039eb5c74b58e25ff96c0171a514585e9 before vendoring it in my project.

Without blt, there are no issues. Any thoughts? Should I try the release version?

### LINKED ISSUE #106: Why are you adding -ffree-form (GFortran)
I did not ask for `-ffree-form` to be added. This breaks my build. Either let the user control all compiler flags with your wonderful abstraction or at least detect from the file name (i.e. f77 ends in `.f` or `.F`). Letting users manage it would probably be better though. This is a deal breaker for us, unless there is a publicized API in BLT for removing compiler flags and/or setting defaults. The docs don't mention anything though.

zbeekman:
- Fixes #105 
  - CXX must be enabled earlier if `ENABLE_GTEST` is set, otherwise a CMake error occurs
- Fixes #106 
  - An option was added to allow users to specify `FIXED`, `FREE` or `INFER` for `FORTRAN_FORMAT`. Defaults to `FREE`: This is the old behavior for backwards compatibility. Setting `FREE` or `FIXED` cause `CMAKE_Fortran_SOURCE` to be set accordingly, setting `INFER` causes `CMAKE_Fortran_SOURCE` to *NOT* get set, so that the compiler handles detection of `FREE` vs `FIXED` source. Most compilers do this via the filename extension:
    - `.f` or `.F`: `FIXED` source formatting is assumed
    - `.f90` or `.F90`: `FREE` source formatting is assumed
 - Allows user to not need to specify `ENABLE_FORTRAN` because this will default to `TRUE` IFF Fortran is detected already by CMake. (Query `ENABLED_LANGUAGES` global property.) Users will enable Fortran via `project()` command by default most of the time, and having to do this again for BLT is redundant and annoying, and unconventional/un-idiomatic.


zbeekman:
Let me know if you want me to cleanup the commits so that my "bug fix" commits are in the commits that introduced the issue. I'm happy to perform an interactive rebase to squash the commits so that each one is an atomic set of self-consistent changes (which they are save for removing `IN_LIST` (eef788c49317982b0341512a9276adbf4925e194) and fixing `set_property()` (8fda9696d323df2f25f5cfd2f730c2a6587b837a).

zbeekman:
I'm also thinking of stealing #99 and including the fix here and also setting up some dependent option logic (#107)

cmake/BLTOptions.cmake
```diff
@@ -53,8 +53,14 @@ option(ENABLE_COVERAGE   "Enables code coverage support" OFF)
 ################################
 # Build Options
 ################################
+get_property(_languages GLOBAL PROPERTY ENABLED_LANGUAGES)
+if(_languages MATCHES "Fortran")
```
kennyweiss: I think ``ENABLED_LANGUAGES`` is a list -- https://cmake.org/cmake/help/v3.8/prop_gbl/ENABLED_LANGUAGES.html#prop_gbl:ENABLED_LANGUAGES

Does this only work if Fortran is the only enabled language?  
What if multiple languages are enabled?

SetupBLT.cmake
```diff
@@ -82,6 +82,13 @@ if (NOT BLT_LOADED)
   endif()
 
   ################################
+  # Enable CXX required by GTEST
+  ################################
+  if(ENABLE_GTEST)
```
kennyweiss: Not sure about this one.

I think we'd probably want the logic to go the other way.
I.e. rather than enabling ``CXX`` when ``GTEST`` is on, it should be a config error if ``GTEST`` is on when ``CXX`` is not available, and there should be an appropriate error message.
(Although, perhaps this could be valid for C projects that want to use gtest?)

Can others please weigh in here?

cmake/BLTOptions.cmake
kennyweiss: Since one of our core macros already uses ``if( ... IN_LIST ...)``  (see the ``blt_setup_target`` macro), I think it is safe to say that blt already requires cmake 3.3 or higher.  

We should probably note this in our documentation, add a check in ``SetupBLT`` and set policy ``CMP0057`` appropriately. 

See: https://stackoverflow.com/questions/23323147/best-way-to-check-with-cmake-whether-list-containts-a-specific-entry  for some other options for checking if an item is in a list.

cmake/BLTOptions.cmake
```diff
@@ -93,6 +99,10 @@ if( NOT ( ( BLT_CXX_STD STREQUAL "c++98" )
 Valid Options are ( c++98, c++11, c++14 )")
 endif()
 
+set(FORTRAN_FORMAT FREE CACHE STRING
+  "Fortran source formatting: FREE (default), FIXED, or INFER")
+set_property(CACHE FORTRAN_FORMAT PROPERTY STRINGS FREE FIXED INFER)
```
kennyweiss: How often do we expect users to want to explicitly set this option?
Should this be marked as an advanced cache variable?

SetupBLT.cmake
zbeekman: I will note that testing C with GTest is perfectly valid/reasonable. I was tempted to implement it the way you were suggesting, but I didn't want to cause unexpected behavior (i.e. this will break backwards compatibility if people rely on GTest for C codes and haven't been explicitly enabling it)

cmake/BLTOptions.cmake
zbeekman: Yes, `ENABLED_LANGUAGES` IS a list.

Fortunately `MATCHES` takes a regex so this check just looks for `Fortran` any where in the "list" (i.e. string with semicolons in it)

I noticed you were using IN_LIST, but I didn't see any calls to deal with CMP0057. I think it would be wise/safe to set CMP0057 to NEW.

I agree, IN_LIST is a much better solution.

cmake/BLTOptions.cmake
zbeekman: I was surprised and confused that you were setting FREE for everything, and others may have a similar experience. Marking as advanced may delay the resolution of peoples issues... just food for thought. Personally I think INFER should be the default, but it's your project, so I'll defer to your preferences.

cmake/SetupCompilerOptions.cmake
```diff
@@ -327,8 +327,15 @@ if(ENABLE_FORTRAN)
...
 
-    # default property to free form
-    set(CMAKE_Fortran_FORMAT FREE)
+    if(FORTRAN_FORMAT MATCHES "^FREE$|^FIXED$")
+        set(CMAKE_Fortran_FORMAT ${FORTRAN_FORMAT})
+        message(STATUS "Defaulting to ${FORTRAN_FORMAT} formatted Fortran source files")
+    elseif(FORTRAN_FORMAT MATCHES INFER)
+        message(STATUS "Fortran source formatting not set---the compiler may infer this from each source file")
+    else()
+        message(ERROR
```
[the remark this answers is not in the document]
zbeekman: You're totally right, my apologies.

cmake/BLTOptions.cmake
zbeekman: To further elaborate, IN_LIST was causing CI failures because, AFAICT, CMP0057 was not set anywhere. My preference would be to set `CMP0057` to `NEW` and switch back to `IN_LIST` (revert eef788c49317982b0341512a9276adbf4925e194)

zbeekman:
@kennyweiss Since you're waiting on input from colleagues I won't do any major rebasing to cleanup the commits.

I'll fix the [`message(ERROR ...` problem](https://github.com/LLNL/blt/pull/108/files/8fda9696d323df2f25f5cfd2f730c2a6587b837a#diff-593cec9ab98299a1f902c71dba565433) and wait for further direction.

cmake/BLTOptions.cmake
kennyweiss: I created issue #109 about the required version of cmake and the ``CMP0057`` issue.

zbeekman:
I'm going to force push some cleaned up commits to my fork. This will update some of the diffs here and cause some of the commentary to go away. If you want me to undo that, just let me know, I've saved a local branch to restore it from.

zbeekman:
Please let me know if there's anything else I can do to help get this PR to a state where you would be willing to incorporate some or all of the proposed changes. If there are issues with a smaller set of the changes I can factor the PR into smaller one. Your wish is my command! (At least as far as this PR goes)

cmake/BLTOptions.cmake
kennyweiss: After an internal (offline) discussion, we're leaning towards making ``INFER`` the default, and projects that want to change this default can explicitly set the desired behavior.

This seems likely to lead to fewer surprises for fortran projects.

If the default is ``INFER``, would it make sense for ``FORTRAN_FORMAT`` to be marked as advanced?

SetupBLT.cmake
white238: Was this brought on by an error you encountered?

zbeekman:
I think it could go either way.

I personally tend to reserve advanced CMake options for ones targeting the
project’s developers most of the time, or ones that are blatantly
experimental or dangerous (like in source builds). Basically anything where
a particular user requested some non standard functionality etc.

In this vein I’m inclined to default any options a project provided as
non-advanced. In my mind advanced says “you’re on your own if you mess with
this and I can’t really envision a justifiable reason for doing so.” In
practice, however, I find most of the time users don’t even know about
`ccmake` or `cmake-gui` and therefore advanced vs normal is of very little
impact.

SetupBLT.cmake
kennyweiss: I think it was this: #105

zbeekman: @white238 yes #105. If you create a C or Fortran only minimal project CMake (at least the latest the latest stable version) throws an error. It’s possible it could be a CMake bug. I didn’t localize it beyond enabling gtest

white238: Sorry coming in to this at the end due to a vacation (Disneyland was fun as always). This looks great.

white238:
This looks good, I'll admit I am not great with Fortran so any improvements from a person more knowledgeable than me is always welcome.

cmake/BLTOptions.cmake
zbeekman: I think it could go either way.

I personally tend to reserve advanced CMake options for ones targeting the
project’s developers most of the time, or ones that are blatantly
experimental or dangerous (like in source builds). Basically anything where
a particular user requested some non standard functionality etc.

In this vein I’m inclined to default any options that a project provides to
non-advanced. In my mind, advanced says “you’re on your own if you mess with
this and we cannot envision a justifiable reason for doing so.” In
practice, however, I find most of the time users don’t even know about
`ccmake` or `cmake-gui` and therefore advanced vs normal is of very little
impact.

zbeekman:
Here is a summary of issues that need to be decided on to merge this PR. If you are happy with the way things are you can check the checkbox, or edit the comment and add your GH name saying so for the purposes of internal discussions. If you check the checkbox I will make the appropriate change and push up to this PR

 - [ ] [Default to enabling CXX if GTEST enabled](https://github.com/LLNL/blt/pull/108/files/065dc44e12f94afcc514dbbc06dd1105a7b620b4#diff-5d7cb2c8be5f80418f33fe762b558b09)
    - Alternate solution: disable GTEST by default UNLESS CXX enabled in project
 - [ ] Leave [`_languages MATCHES "Fortran"`](https://github.com/LLNL/blt/pull/108/files/065dc44e12f94afcc514dbbc06dd1105a7b620b4#diff-6611ff752bc7a19fe63f12190de9f1b9)
    - vs using less portable `IN_LIST` (requires CMP0057 NEW and CMake >= 3.3, see #109)
 - [ ] [Default to letting compiler `INFER` Fortran source formatting](https://github.com/LLNL/blt/pull/108/files/065dc44e12f94afcc514dbbc06dd1105a7b620b4#diff-6611ff752bc7a19fe63f12190de9f1b9), [__*not*__ making `FORTRAN_FORMAT` "advanced"](https://github.com/LLNL/blt/pull/108#discussion_r147868737)
    - vs keeping current behavior defaulting to forced `FREE` formatting for all Fortran sources and/or marking the `FORTRAN_FORMAT` option advanced.

I'm happy to help out a little bit since resolving this will let me use BLT in a project that I am working on that requires porting from Makefiles/autotools. I can let it sit too, but I want to make sure that anything I can address gets addressed as quickly as it is feasible to do so.

cmake/BLTOptions.cmake
cyrush: After reviewing -- I still feel we should simply remove all of the logic related to this and let folks using 
BLT set `CMAKE_Fortran_FORMAT`  (or the related source properties) as they see fit in their project. 

Adding the option is another thing to explain and support vs standard CMake, which in this case is  very straight forward. 

Again, sorry for the extra effort, but not having it means we save on docs, maintenance, testing, etc in the long run. 

@kennyweiss @white238, what do you think?

cmake/BLTOptions.cmake
```diff
@@ -53,8 +53,14 @@ option(ENABLE_COVERAGE   "Enables code coverage support" OFF)
...
+get_property(_languages GLOBAL PROPERTY ENABLED_LANGUAGES)
+if(_languages MATCHES "Fortran")
+    set(_fortran_already_enabled TRUE)
+else()
+    set(_fortran_already_enabled FALSE)
+endif()
 option(ENABLE_COPY_HEADERS "Copy headers to build directory" OFF)
-option(ENABLE_FORTRAN      "Enables Fortran compiler support" OFF)
+option(ENABLE_FORTRAN      "Enables Fortran compiler support" ${_fortran_already_enabled})
 option(ENABLE_SHARED_LIBS  "Enables shared libraries." OFF)
```
cyrush: +1 for `ENABLE_FORTRAN` defaulting to on when folks already enabled fortran in the project stmt - Thanks for this change!

SetupBLT.cmake
cyrush: Like @kennyweiss  -- I would prefer we add an error message if  ENABLE_GTEST is on, but CXX isn't enabled. 

(one that gives folks a chance to know whats wrong vs any cmake error they will encounter otherwise) 

Even though you can use gtest for testing c, it's a c++ framework and folks are going to have to use a c++ compiler to compile tests.

white238:
I agree with @cyrush on the Fortran format...

cmake/BLTOptions.cmake
kennyweiss: The benefit of keeping it is that the code is already written and it would make it easy for people setting up a fortran code to know about this distinction.  

The downside is having to describe/maintain it, and the added complexity of additional configuration options.

I'm ok with removing this feature.

kennyweiss:
Thanks @cyrush 
Ok, so, to summarize:

- [ ] It is a configuration error when ENABLE_GTEST = ON and CXX is not one of the project languages. We need a clear check / error message for this case.
- [ ] Leave ``_languages MATCHES "Fortran"``.  We will upgrade to ``if( ... IN_LIST ... )`` when dealing with #109 
- [ ] Remove FREE_FORM option from BLT.  User codes that want this feature will have to explicity set this outside BLT

Does this work for everyone?

cyrush:
@kennyweiss , yes sounds good to me.

zbeekman:
FWIW, sounds good to me.

kennyweiss:
@zbeekman -- are you planning to work on the requested changes, or would you like us to take care of it?

zbeekman:
Yes gonna do it in a sec. busy prepping for SC

cyrush:
@zbeekman, want to sync up at SC (I'll be there Sun - Wed)?

zbeekman:
Sure, would be great to meet you! My email is ibeekman@paratools.com, let's
coordinate that way. I'll be at the PGAS booth a bit too, on the
OpenCoarrays poster there.

cyrush:
@zbeekman  sounds good, I'll ping you via email!

kennyweiss:
Thanks @zbeekman -- no rush at all, just wanted to follow up.

zbeekman:
It looks like the FRUIT smoke tests fail w/ GFortran due to the `.f` file extension. I'll include a fix.

zbeekman:
Can I just rename `blt_fruit_smoke.f` to `blt_fruit_smoke.f90`?

I think that should work for 90% of compilers. This would be the more canonical Fortran convention, rather than explicitly setting free-form source properties in CMake.

kennyweiss:
@zbeekman -- I am fine with the file extension change for this test.

A different strategy would be to set the free-form flag for the fruit smoke test
(with a one-line comment in the cmake file about why you're doing that).

zbeekman:
@kennyweiss yes, I know, but IMO, there are a set of four file extensions allowable for Fortran files (other than include files...that's less clear):

- `.f`: Fixed form, no preprocessing macros
- `.F`: Fixed form, preprocessing required
- `.f90`: Any free form Fortran source file, no preprocessing macros
- `.F90`: Any free form Fortran, preprocessing required

zbeekman:
gah, my gtest fix is broken, stand by...

cyrush:
@zbeekman renaming sounds like the best solution

PR #110, opened 2017-10-30
`BUILD_SHARED_LIBS` as `ENABLE_SHA...` default

zbeekman:
- If defined, use `BUILD_SHARED_LIBS` to populate the default value of
   `ENABLE_SHARED_LIBS`
 - Fixes LLNL/blt#99
 - Shamelessly stole solution from @cyrush<sup>[1]</sup>

[1]: https://github.com/LLNL/blt/issues/99#issuecomment-338068664


cyrush:
@zbeekman looks good!

Interestingly, when I setup appveyor for windows testing, I mistakenly already used  BUILD_SHARED_LIBS.

Which means we were not testing the shared libs case, and of course it failed. 
We will need to resolve, not sure when i will have time to take a pass.

zbeekman:
I rebased on latest master, and did some brief hunting for the issue on appveyor. It looks like the visual studio project is trying to link `example.exe` against `libexample.lib` instead of `libexample.dll`, which is the name of the library that will be built if dynamic libraries are enabled. It seems odd to me that this is happening at all, unless the logic to deposit the dll in the right directory has something broken in it. Otherwise, I would guess that CMake should be able to resolve the name of the library created, since `target_link_libraries` is just passed a target name. I looked through the logic and didn't see any place where the custom dependency resolution logic or target creation logic would be renaming a .dll to a .lib. 

__TL;DR;__ I'm a windows newb, and I don't see anything obviously wrong in BLT so @cyrush or someone else my proficient with Windows will likely have to debug this.

cyrush:
This sounds familiar. I think this is a simple issue, but painful to debug in appveyor. 

I'll try to carve out some time to investigate on a windows machine.

zbeekman:
Resolved the merge conflicts. Still somewhat clueless about fixing the appveyor failures

cyrush:
going to try to poke at it on my windows machine soon, sorry for the delay

zbeekman:
No worries, just wanted to make your job merging this as easy as I can... If it doesn't get merged over the next few weeks, my windows + CMake proficiency may level up to the point where I can take care of this myself 😝

cyrush:
@zbeekman  found the issue, a classic windows shared libs issue.

We need to export the symbols in the example lib or else the `.lib` file wont generate, and we can't link (even though the dll exists). I'll work up a solution soon.

cmake/blt-test/CMakeLists.txt
```diff
@@ -59,6 +59,10 @@ endif()
 
 include(${BLT_SOURCE_DIR}/SetupBLT.cmake)
 
+if(WIN32 AND ENABLE_SHARED_LIBS)
+    add_definitions("WIN32_SHARED_LIBS")
```
kennyweiss: This should be an ``add_target_definitions``, right?

cmake/blt-test/src/Example_Exports.h
```diff
@@ -0,0 +1,73 @@
...
+// 
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+#ifndef _EXAMPLE_EXPORTS_H_
+#define _EXAMPLE_EXPORTS_H_
+
+//-----------------------------------------------------------------------------
+// -- define proper lib exports for various platforms -- 
+//-----------------------------------------------------------------------------
+#if defined(_WIN32)
```
kennyweiss: @cyrush -- It's difficult to see the nesting of the conditional. 
Can you please indented these for clarity?  E.g. 

```cpp
#if defined(_WIN32)
#  if defined(WIN32_SHARED_LIBS)
#    if defined(EXAMPLE_EXPORTS) || defined(example_EXPORTS)
#      define EXAMPLE_API __declspec(dllexport)
#    else
#      define EXAMPLE_API __declspec(dllimport)
#  endif
...
#endif

```

cmake/blt-test/CMakeLists.txt
cyrush: perhaps, but trying in a way that applies to all things on windows.

cmake/blt-test/src/Example_Exports.h
cyrush: sure, I can fix that

cyrush:
this is still a work in progress, I need to iterate on an actual windows machine vs fighting appveyor.

cyrush:
Alright, got things working!

I tried to make minimal changes to the tutorial source, since it's pulled into the docs (in some cases by line number) Some line numbers did shift,  I updated the docs and checked -- I think we are ok. 

Not super happy with the use of the windows define to select the proper export type on windows.

More complex projects would be better served using cmake to configure (generate) a header tied to static vs shared on widows, but for our testing projects and the tutorial that seems like too much to do. 

We are ready to merge -- thanks @zbeekman for the help with this!

kennyweiss:
Thanks @cyrush.

Since the exports header files are boilerplate, somewhat intricate, and necessary to get shared libs working, it seems like a great feature for BLT to take care for us.  E.g. with a simple macro/function that generates the file for us with a specified prefix. That way all blt-based projects can benefit from this.

If it is one config file per-library, perhaps it can be incorporated directly into ``blt_add_library`` ?

cyrush:
I have thought about automating it several times. 

I am in favor of a helper function that can generate a boiler plate header for you that you still explicitly include. 

I don't think it should be directly in `blt_add_library`:

Why? 
1) While its good form to do it on Linux (and OSX), its not required. Many folks may not care about export headers at all
2) Folks may not like my style of export headers :-). They many want to use their own pragmas or settings related to exports. While my style has been tested many places, its not a complete solution for all compilers. 
3) In some cases I share export headers across libraries, this requires setting the proper defines (vs the ones that CMake automatically creates for you)
4) The current windows solution for our tests w/ defines ugly, but thats really my opinion. I would suggest not using this solution, but some folks may want to use defines.

PR #113, opened 2017-11-03
Add new version of blt_add_code_checks and fix previous checkin

white238:


cmake/SetupCodeChecks.cmake
```diff
@@ -57,6 +57,109 @@ if(UNCRUSTIFY_FOUND)
...
+        else()
+            message(FATAL_ERROR "blt_add_code_checks given source file with unknown file extension.")
+        endif()
+    endforeach()
+
+    # Add code checks
+    if (UNCRUSTIFY_FOUND AND DEFINED arg_UNCRUSTIFY_CFG_FILE)
+        # Find unused name
+        if (NOT DEFINED BLT_${arg_BASE_NAME}_UNCRUSTIFY_INDEX)
+            set(BLT_${arg_BASE_NAME}_UNCRUSTIFY_INDEX "0" CACHE STRING "")
```
kennyweiss: This should be marked as advanced, and probably have a brief description

cmake/SetupCodeChecks.cmake
```diff
@@ -127,11 +234,11 @@ macro(blt_add_uncrustify_target)
             WORKING_DIRECTORY ${_wd} 
             COMMENT "${arg_COMMENT}Running uncrustify source code formatting checks.")
```
kennyweiss: Should there be a space between ``${arg_COMMENT}`` and ``Running`` ?

white238: If you put a space there when there isn't a comment it has an extra space at the beginning.

cmake/SetupCodeChecks.cmake
```diff
@@ -57,6 +57,109 @@ if(UNCRUSTIFY_FOUND)
...
+##
+## UNCRUSTIFY_CFG_FILE is the configuration file for Uncrustify. If UNCRUSTIFY_EXECUTABLE
+## is defined, found, and UNCRUSTIFY_CFG_FILE is defined it will create both check and
+## style function for the given C/C++ files.
+##
+## Repeated calls to this macro with the same BASE_NAME creates numbered targets.
+## For example: ${BASE_NAME}_uncrustify_check, ${BASE_NAME}_uncrustify_check_1, ...
+##------------------------------------------------------------------------------
+
+macro(blt_add_code_checks)
```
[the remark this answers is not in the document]
white238: Yes, I will be adding cppcheck soon but wanted to get the base in first.

cmake/SetupCodeChecks.cmake
```diff
@@ -57,6 +57,109 @@ if(UNCRUSTIFY_FOUND)
...
+
+        get_filename_component(_ext ${_full_path} EXT)
+        file(RELATIVE_PATH _relpath ${CMAKE_BINARY_DIR} ${_full_path})
+
+        if(${_ext} IN_LIST arg_C_FILE_EXTS)
+            list(APPEND _c_sources ${_relpath})
+        elseif(${_ext} IN_LIST arg_F_FILE_EXTS)
+            list(APPEND _f_sources ${_relpath})
+        else()
+            message(FATAL_ERROR "blt_add_code_checks given source file with unknown file extension.")
```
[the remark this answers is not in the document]
cyrush: i am fine with fatal error, we don't want folks passing things in that can't possibly be handled.
Creating another explicit list for this case  seems seems fine to me.

cmake/SetupCodeChecks.cmake
```diff
@@ -57,6 +57,111 @@ if(UNCRUSTIFY_FOUND)
...
+##                      C_FILE_EXTS         [ext1 [ext2 ...]]
+##                      F_FILE_EXTS         [ext1 [ext2 ...]]
+##                      UNCRUSTIFY_CFG_FILE <path to uncrustify config file>)
+##
+## This macro adds all enabled code check targets for the given SOURCES. It
+## filters based on file extensions.
+##
+## C_FILE_EXTS is an optional list of file extensions to filter out the C/C++ SOURCES.
+## Otherwise it defaults to: ".cpp" ".hpp" ".cxx" ".hxx" ".cc" ".c" ".h" ".hh"
+##
```
cyrush: Can we make it easy for a user to extend this list vs provide their own?

The case I am thinking about is something like  an ".inc" file as c that we want to check, but we don't want to have to re-add all the default c file cases.

white238: If we append to the list instead of replacing it, we wouldn't be able to remove extensions from the list.  Though I'm not sure why anyone would want to do that.

PR #115, opened 2017-11-07
Feature/add folder support

kennyweiss:
* Adds a new cmake option to blt: ``ENABLE_FOLDERS``, which helps organize cmake targets within an IDE.  This is off by default.
* Adds a new blt macro: ``blt_set_target_folder`` to add the folder property to a cmake target
* Adds optional keyword ``FOLDER`` to ``blt_add_library`` and ``blt_add_executable`` macros
* Applies the folder property to blt's test and buildtin thirdparty targets.

### DIFF
--- a/thirdparty_builtin/CMakeLists.txt
+++ b/thirdparty_builtin/CMakeLists.txt
@@ -173,7 +173,7 @@ if(ENABLE_TESTS)
     # Set the folder property of the blt thirdparty libraries 
     if(ENABLE_FOLDERS)
         foreach(tpl ${_blt_thirdparty})
-            blt_set_target_folder(TO ${tpl} NAME blt/thirdparty)
+            blt_set_target_folder(TARGET ${tpl} FOLDER blt/thirdparty)
         endforeach()
     endif()
 
--- a/thirdparty_builtin/CMakeLists.txt
+++ b/thirdparty_builtin/CMakeLists.txt
@@ -43,7 +43,7 @@
 if(ENABLE_TESTS)
     include(CTest)
 
-    set(_blt_thirdparty)
+    set(_blt_tpl_targets) # tracks names of enabled tpl targets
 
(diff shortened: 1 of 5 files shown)


thirdparty_builtin/CMakeLists.txt
```diff
@@ -154,10 +161,20 @@ if(ENABLE_TESTS)
...
         add_custom_target(run_benchmarks 
                           COMMAND ctest -C Benchmark -VV -R benchmark
                           WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
                           )
     endif()
+
+    # Set the folder property of the blt thirdparty libraries 
+    if(ENABLE_FOLDERS)
+        foreach(tpl ${_blt_thirdparty})
+            blt_set_target_folder(TO ${tpl} NAME blt/thirdparty)
```
cyrush: With the current incarnation --  I think folks may get confused and think that `TO` the directory not the target. 

I know it mirrors other args, but `TO` seems very strange in this context. 

The other methods are more like ``add things TO something``, where this is setting something about a target.

kennyweiss: I can change it to ``TARGET``

kennyweiss:
I rebased off master, and modified the keywords to ``blt_set_target_folder`` to ``TARGET`` and ``FOLDER``.  

Thanks for the suggestion @cyrush !

kennyweiss:
Here is a screenshot of how ``blt``'s projects are organized in Visual Studio 2017 when folders are enabled (``ENABLE_FOLDERS=ON``).  

blt_vs_soln_folders

Note that cmake creates folders for several of the built-in targets: ``CMakePredefinedTargets`` and ``CTestDashboardTargets``, and blt creates forders for its test and thirdparty library targets (``blt/tests`` and ``blt/thirdparty``, respectively).

The ``check``, ``docs`` and ``style`` targets are not currently organized into folders.

PR #120, opened 2017-12-01
Add "inl" to list of file extensions for code checks

davidbeckingsale:
`.inl` is nice for templates or inline functions, and I don't want to have to duplicate the whole list of C file extensions.


cyrush:
@white238  has some big changes in #88 that may conflict, but not sure. 

Seems like a simple fix, can we can remember to check that its propagated when Chris' changes are done?

white238:
@cyrush that whole PR is so out of date at this point (and some have been merged already) that I am going to have to just start over with it.  Hopefully soon when i finish the current Axom task.

PR #122, opened 2017-12-07
modify SetupMPI.cmake to treat mpi headers as system headers

rrsettgast:
I was getting a lot of warnings for mpi headers. Adding the "TREAT_INCLUDES_AS_SYSTEM ON" argument to the blt_register_library() in SetupMPI.cmake gets rid of the warnings. I am assuming that you wouldn't ever want to see these warnings?


rrsettgast:
@kennyweiss it is on my laptop. This is the only place I see it.

rrsettgast:
Is this ok to merge?

white238:
As soon as this passes the AppVeyor builds I will merge it.

Issue #126, opened 2018-01-24
appveyor is now using cmake 3.10.1


cyrush:
for conduit testing, this change caused issues with how we were specifying MS MPI.
The recipe I am using for conduit was taken from BLT's setup, so I think BLT will need updates. 

Working on resolving for conduit, then i plan to resolve for blt.

cyrush:
even after resolving issues with using env vars to pass MS MPI to cmake --  cmake 3.10.1 seems to insert a bad path into the MPI C and CXX includes, which undermines the build. 

My solution for conduit is to install cmake 3.9.6 in appveyor and use that version instead of the default.

cyrush:
This hit us on your last PR merge to master:

https://ci.appveyor.com/project/cyrush/blt/build/1.0.618

cyrush:
resolved in #130

PR #130, opened 2018-01-30
use cmake 3.9.6 for appveyor, resolve mpi setup issues

cyrush:


cyrush:
this resolves #126

appveyor.yml
```diff
@@ -67,11 +77,10 @@ build_script:
       -D ENABLE_BENCHMARKS=%ENABLE_BENCHMARKS% ^
       -D ENABLE_OPENMP=ON ^
       -D ENABLE_MPI=%ENABLE_MPI% ^
-      -D MPI_HOME:PATH="C:/Program Files (x86)/Microsoft SDKs/MPI" ^
-      -D MPI_C_INCLUDE_PATH:PATH="${MPI_HOME}/Include" ^
-      -D MPI_C_LIBRARIES:PATH="${MPI_HOME}/Lib/x86/msmpi.lib" ^
-      -D MPI_CXX_INCLUDE_PATH:PATH="${MPI_HOME}/Include" ^
-      -D MPI_CXX_LIBRARIES:PATH="${MPI_HOME}/Lib/x86/msmpi.lib" ^
+      -D MPI_C_INCLUDE_PATH:PATH="%MPI_HOME%/Include" ^
+      -D MPI_C_LIBRARIES:PATH="%MPI_HOME%/Lib/x86/msmpi.lib" ^
```
kennyweiss: Not sure if this is the problem here, but I've been able to get around an improper ``PATH`` issue in the path by telling cmake that it was a ``STRING``.   E.g. ``-D MPI_C_INCLUDE_PATH:STRING="%MPI_HOME%/Include``

cyrush: not a powershell pro, but i think for this style, the string will be sub-ed in before passed to cmake. The last style wasn't doing this & while wasn't a problem for cmake 3.8 and 3.9, it actually did cause an issue for 3.10.  That was the only thing I solved for 3.10 before climbing out of the rabbit hole.

PR #132, opened 2018-02-08
Add ability to turn off git support

white238:
This showed up on one of our codes. Even with setting GIT_EXECUTABLE to blank, it found the system one.  Small but needed feature.


cmake/BLTOptions.cmake
```diff
@@ -50,6 +50,12 @@ option(ENABLE_BENCHMARKS "Enables benchmarks" OFF)
 option(ENABLE_COVERAGE   "Enables code coverage support" OFF)
 
 
+################################
+# TPL Options
+################################
+option(ENABLE_GIT       "Enables Git support" ON)
```
kennyweiss: Could we expand the string a bit? Perhaps "Enables git-related macros in blt".

cmake/thirdparty/SetupThirdParty.cmake
```diff
@@ -47,11 +47,14 @@
...
-  message(STATUS "Git Executable: " ${GIT_EXECUTABLE} )
-  message(STATUS "Git Version: " ${GIT_VERSION_STRING} )
+message(STATUS "Git Support is ${ENABLE_GIT}")
+if (ENABLE_GIT)
+    find_package(Git)
+    if (Git_FOUND)
+        set(GIT_FOUND TRUE)
+        message(STATUS "Git Executable: " ${GIT_EXECUTABLE} )
+        message(STATUS "Git Version: " ${GIT_VERSION_STRING} )
+    endif()
```
kennyweiss: Please add an ``else()`` clause that says we couldn't find git, since the user requested it (perhaps by default), and the message above says that "Git Support is ON"

cmake/BLTOptions.cmake
white238: It's not just the macros in BLT that it is turning off/on.  It also is enables git in the codes build system as well via Git_FOUND/GIT_EXECUTABLE.

PR #134, opened 2018-02-20
Fix issue where device code was relocated too early in library builds.

robinson96:
Fixes an issue where LINK_WITH_NVCC would cause device symbols in libraries to be invisible at final link time. Only affected libraries that provided device side functionality, callable from kernels.

### DIFF
--- a/cmake/thirdparty/FindCUDA.cmake
+++ b/cmake/thirdparty/FindCUDA.cmake
@@ -1784,11 +1784,14 @@ macro(CUDA_ADD_LIBRARY cuda_target)
   if (NOT CUDA_LINK_WITH_NVCC)
       CUDA_LINK_SEPARABLE_COMPILATION_OBJECTS("${link_file}" ${cuda_target} "${_options}" "${${cuda_target}_SEPARABLE_COMPILATION_OBJECTS}")
   endif()
-  ####################
-  # END BLT CHANGE 
-  ####################
 
-  if(CUDA_SEPARABLE_COMPILATION)
+  #BLT edit: added AND NOT CUDA_LINK_WITH_NVCC
+  if(CUDA_SEPARABLE_COMPILATION AND NOT CUDA_LINK_WITH_NVCC)
+    # end up with duplicate symbols if cudadevrt is on the executable link line
+    # when linking with NVCC
+    ####################
+    # END BLT CHANGE 
+    ####################
     target_link_libraries(${cuda_target} PUBLIC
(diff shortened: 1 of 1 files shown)


cmake/thirdparty/FindCUDA.cmake
```diff
@@ -1844,6 +1830,8 @@ macro(CUDA_ADD_EXECUTABLE cuda_target)
   ####################
   if (NOT CUDA_LINK_WITH_NVCC)
       CUDA_COMPUTE_SEPARABLE_COMPILATION_OBJECT_FILE_NAME(link_file ${cuda_target} "${${cuda_target}_SEPARABLE_COMPILATION_OBJECTS}")
+  else()
+     set(link_file "")
   endif()
```
cyrush: hard to tell on quick review -- these stmts are already enclosed in BLT CHANGE comment blocks?

cmake/thirdparty/FindCUDA.cmake
```diff
@@ -1872,7 +1860,7 @@ macro(CUDA_ADD_EXECUTABLE cuda_target)
       # We need to link with nvcc. Host linking prevents calling device methods in one library from kernels in another (yes, even with separable
       # compilation)
       string (REPLACE ";" " " LD_FLAGS_STR "${CUDA_NVCC_FLAGS}")
-      set (CMAKE_CUDA_LINK_EXECUTABLE  "${CUDA_NVCC_EXECUTABLE}  <FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET> <LINK_LIBRARIES>" CACHE PATH "" )
+      set (CMAKE_CUDA_LINK_EXECUTABLE  "${CUDA_NVCC_EXECUTABLE}  <FLAGS> <LINK_FLAGS> <OBJECTS>  -o <TARGET> <LINK_LIBRARIES>" CACHE STRING "" )
       set_target_properties(${cuda_target} PROPERTIES LINKER_LANGUAGE "CUDA")
```
cyrush: hard to tell on quick review -- these stmts are already enclosed in BLT CHANGE comment blocks?

cmake/thirdparty/FindCUDA.cmake
robinson96: yup

cmake/thirdparty/FindCUDA.cmake
robinson96: this one isn't. Chris and I talked about at and we thought the change was minor enough to not warrant it...

cyrush: agreed

PR #137, opened 2018-03-21
Removed ENABLE_BENCHMARK dependency on ENABLE_TESTS

mcfadden8:
Removed ENABLE_BENCHMARK's dependency on ENABLE_TESTS so that benchmarks may be created independent from what ENABLE_TESTS is set to.


thirdparty_builtin/CMakeLists.txt
```diff
@@ -181,3 +153,32 @@ if(ENABLE_TESTS)
...
+                         INCLUDES ${benchmark_SOURCE_DIR}/include ${benchmark_SOURCE_DIR}
+                         LIBRARIES benchmark ${RT_LIBRARIES}
+                         TREAT_INCLUDES_AS_SYSTEM ON
+                         )
+
+    list(APPEND _blt_tpl_targets benchmark)
+
+    # This sets up a target to run the benchmarks
+    add_custom_target(run_benchmarks 
+                      COMMAND ctest -C Benchmark -VV -R benchmark
```
davidbeckingsale: Can ctest still be used if `enable_testing()` has not been called? @white238

kennyweiss: It is difficult to tell offhand -- @mcfadden8 can you please try this manually and see what happens?

To maintain this feature, perhaps we should add a configuration to travis that has benchmarks enabled and testing disabled?  (The downside is longer CI waiting times).

thirdparty_builtin/CMakeLists.txt
```diff
@@ -181,3 +153,32 @@ if(ENABLE_TESTS)
...
+                      WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+                      )
+endif()
```
kennyweiss: Please add the ``if(ENABLE_FOLDERS)`` part for benchmarks.

davidbeckingsale:
The context is building Umpire, which has both correctness tests and performance benchmarks. These are conditionally built based on the ENABLE_TESTS and ENABLE_BENCHMARKS options.

I don’t know why benchmarks would need to depend on tests? It doesn’t make sense.

kennyweiss:
Thanks for the context @davidbeckingsale.

I think I set it up originally, and the best way I could find to automate running the benchmarks (when they were enabled) was through ctest using the custom "Benchmarks" configuration (in the ``run_benchmarks`` custom target).

I can see the rationale to separate the two e.g. to build the benchmarks and manually run them. The ``run_benchmarks`` target will likely need to depend on both ``ENABLE_TESTING`` and ``ENABLE_BENCHMARKS`` -- unless you have found a better way to do this.

thirdparty_builtin/CMakeLists.txt
mcfadden8: The run_benchmarks target doesn't appear to work (even after adding the ENABLE_FOLDERS logic).  I see the following when ENABLE_TESTS=Off:

```script
$ make run_benchmarks
Scanning dependencies of target run_benchmarks
UpdateCTestConfiguration  from :/g/g0/martymcf/umpire/build/linux-rhel7-x86_64/umpire/DartConfiguration.tcl
UpdateCTestConfiguration  from :/g/g0/martymcf/umpire/build/linux-rhel7-x86_64/umpire/DartConfiguration.tcl
Test project /g/g0/martymcf/umpire/build/linux-rhel7-x86_64/umpire
Constructing a list of tests
Checking test dependency graph...
Checking test dependency graph end
No tests were found!!!
```

davidbeckingsale:
Can you try the same thing but call “enable_testing()” somewhere?

Also, are our benchmarks using the add_benchmark macro - they might not be getting added to that target.

mcfadden8:
Our (umpire) benchmarks are calling blt_enable_benchmark().  `make run_benchmarks` does work if we run with -DENABLE_TESTS=On.

mcfadden8:
When I placed ‘enable_testing()’ inside of the CMakeLists.txt file within the benchmark/ folder of umpire, I still get the same result.  The ‘make run_benchmarks’ still does not find any tests.

thirdparty_builtin/CMakeLists.txt
```diff
@@ -145,39 +145,40 @@ if(ENABLE_TESTS)
...
-    endif()
+    list(APPEND _blt_tpl_targets benchmark)
 
-    # Set the folder property of the blt thirdparty libraries 
-    if(ENABLE_FOLDERS)
-        foreach(tpl ${_blt_tpl_targets})
-            blt_set_target_folder(TARGET ${tpl} FOLDER blt/thirdparty)
-        endforeach()
-    endif()
+    # This sets up a target to run the benchmarks
```
davidbeckingsale: @mcfadden8 - can you guard this with `ENABLE_TESTS`?

kennyweiss: Thanks @davidbeckingsale -- that's what I was trying to say -- the ``run_benchmarks`` target depends on ``ENABLE_TESTING``

thirdparty_builtin/CMakeLists.txt
kennyweiss: To clarify:  The ``if(ENABLE_FOLDERS)`` part has nothing to do with testing or benchmarks.  It is a feature for multi-config generators in IDE's like visual studio that helps organize the code into hierarchical folders when you have a lot of projects.

thirdparty_builtin/CMakeLists.txt
```diff
@@ -145,39 +145,40 @@ if(ENABLE_TESTS)
...
+    add_custom_target(run_benchmarks 
+                      COMMAND ctest -C Benchmark -VV -R benchmark
+                      WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+                      )
+endif()
 
+# Set the folder property of the blt thirdparty libraries 
+if(ENABLE_FOLDERS)
+    foreach(tpl ${_blt_tpl_targets})
```
kennyweiss: With this code separated from the conditions, I think you will also need to move the line that initializes ``_blt_tpl_targets`` out of the ``ENABLE_TESTING`` part.

I.e. this line:
```cmake
 set(_blt_tpl_targets) # tracks names of enabled tpl targets
```
should be above the ``if(ENABLE_TESTING)```

See: https://github.com/LLNL/blt/pull/137/files#diff-f7bc02caeeac3bac796502ed63313eafR46

davidbeckingsale: 👍

mcfadden8:
It seems like the reported AppVeyor build failure [1.0.679](https://ci.appveyor.com/project/cyrush/blt/build/1.0.679/job/n2ydjwgdsiqdaqn5) was not caused by my changes.  What is the preferred way of restarting an AppVeyor test?

davidbeckingsale:
@cyrush - can you restart the failed AppVeyor build?

cyrush:
@davidbeckingsale @mcfadden8 , just restarted appveyor.

PR #148, opened 2018-05-08
Fix shared and static comingled builds, multivalue arg problem

white238:
* This fixes the ability to specify blt_add_library's SHARED parameter when the global BUILD_SHARED_LIBS or ENABLE_SHARED_LIBS is set
* Fix flags being lost due to being set as single value instead of multi value args


cmake/BLTMacros.cmake
```diff
@@ -258,36 +258,32 @@ macro(blt_register_library)
...
-        set(BLT_${uppercase_name}_LINK_FLAGS "BLT_NO_LINK_FLAGS" CACHE PATH "" FORCE)
+        set(BLT_${uppercase_name}_LINK_FLAGS "${arg_LINK_FLAGS}" CACHE STRING "" FORCE)
+        mark_as_advanced(BLT_${uppercase_name}_LINK_FLAGS)
     endif()
 
-    mark_as_advanced(BLT_${uppercase_name}_LINK_FLAGS)
-
     if( arg_DEFINES )
-        set(BLT_${uppercase_name}_DEFINES ${arg_DEFINES} CACHE PATH "" FORCE)
+        set(BLT_${uppercase_name}_DEFINES ${arg_DEFINES} CACHE LIST "" FORCE)
```
kennyweiss: Should this one be a ``STRING``? Or is cmake able to convert it to a space-separated string when it adds it to the compiler command?

cmake/BLTPrivateMacros.cmake
```diff
@@ -163,6 +163,9 @@ macro(blt_setup_target)
         endif()
 
         if ( DEFINED BLT_${uppercase_dependency}_LIBRARIES)
+            # This stops cmake from falling back on adding -l<library name>
```
kennyweiss: Suggestion:  
> This prevents cmake from adding...

kennyweiss:
The problem appears to be fixed, but for some reason appveyor is sporadically running into an issue downloading cmake:
```bash
[00:00:12] Start-FileDownload 'https://cmake.org/files/v3.9/cmake-3.9.6-win64-x64.msi'
[00:00:33] Start-FileDownloadInternal : Error downloading remote file: One or more errors occurred.
[00:00:33] Inner Exception: Unable to connect to the remote server
```

I don't seem to have the ability to respawn the appveyor job without pushing a commit.

cmake/BLTMacros.cmake
white238: It's a list.  They shouldnt be given as string of -DHAVE_FOO -DHAVE_BAR options... they should be given as a list of HAVE_FOO HAVE_BAR

cmake/BLTMacros.cmake
```diff
@@ -438,12 +455,20 @@ macro(blt_add_library)
...
     endforeach()
     if(_have_fortran)
-      target_include_directories(${arg_NAME} PRIVATE ${CMAKE_Fortran_MODULE_DIRECTORY})
+        target_include_directories(${arg_NAME} PRIVATE ${CMAKE_Fortran_MODULE_DIRECTORY})
     endif()
 
     blt_setup_target( NAME ${arg_NAME}
                       DEPENDS_ON ${arg_DEPENDS_ON} )
 
+    if ( arg_INCLUDES )
```
kennyweiss: Do we need to distinguish here between header only (``INTERFACE``) and compiled (``PUBLIC``) libraries?

cmake/BLTMacros.cmake
```diff
@@ -298,11 +294,14 @@ endmacro(blt_register_library)
 ## blt_add_library( NAME <libname>
 ##                  SOURCES [source1 [source2 ...]]
 ##                  HEADERS [header1 [header2 ...]]
+##                  INCLUDES [dir1 [dir2 ...]]
```
davidbeckingsale: Awesome :)

cmake/BLTMacros.cmake
davidbeckingsale: Can we use `INTERFACE` all the time?

white238: This needs to be more thoroughly tested. I spoke to kenny about it and we agreed to not hold this PR up due to possible problems with header-only libraries.  Switching this to INTERFACE would cause problems though as the includes would not be used for the current library.

PR #152, opened 2018-05-11
Remove gtest from all smoke tests except for gtest/gmock ones

white238:
This caused build errors if you turn off gtest but still want ENABLE_TESTS to be on.


white238:
Got hit by the AppVeyor download problem.  Really need a work around for that.

kennyweiss:
It looks like the default cmake version in appveyor is 3.11.0 -- https://www.appveyor.com/docs/build-environment

Can we just use that one instead of 3.9.6?

cyrush:
@kennyweiss  I think we should stick to a version of cmake that we are ~using, appveyor updates quite often which in itself would yield unhappy surprises.  Some of this is compounded by the fact that I am the only one that can restart the appveyor builds, maybe there is a way to change the config. 


RAJA folks are using a docker image to pin down all dependencies, we could go that route.

Issue #146, opened 2018-04-13
Fix llnl-ray host-config


kennyweiss:
While working on #145, I found that the CUDA-enabled ``llnl-ray`` host-config is no longer working.  
We should fix this.

### Details:

There is a problem with the ``pthreads`` flags in the CUDA example.

I configured the code with:
```bash
export BLT_ROOT=<path/to/blt/root>

cmake  \
   -C ${BLT_ROOT}/host-configs/llnl-ray-blue_os-gcc\@4.9.3.cmake  \
    ${BLT_ROOT}/tests/internal/
```

The error that I'm getting is:
```bash
[ 95%] Linking CUDA executable ../../bin/t_cuda_device_call_from_kernel_exec
nvcc fatal   : Unknown option 'pthread'
make[2]: *** [src/test_cuda_device_call_from_kernel/CMakeFiles/t_cuda_device_call_from_kernel_exec.dir/build.make:89: bin/t_cuda_device_call_from_kernel_exec] Error 1
make[1]: *** [CMakeFiles/Makefile2:1529: src/test_cuda_device_call_from_kernel/CMakeFiles/t_cuda_device_call_from_kernel_exec.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

Notes:
* @cyrush suspects that something in the ``blue_os`` environment changed since this used to work.
* @DavidPoliakoff was able to get the compilation to work by manually changing ``-pthread`` to ``-Xcompiler -pthread`` (thanks David!).  Anyone know where I can make that change?  (preferably in the host-config)

See the discussion in #145 for some additional details.

robinson96:
This is now fixed, and exercises the latest and greatest use of native CMake CUDA support. Surface got updated as well.

PR #154, opened 2018-05-15
Replace use of FindCUDA with native CMake CUDA support.

robinson96:
Meant to be a drop in replacement for FindCUDA. The only change I had to do do my host-config was to drop an explicit definition of -std=c++, as the native CMake puts it in there the right way and nvcc complains if you define it twice.


robinson96:
@cyrush : I'm using 3.9.2 - have not tested with earlier versions. 
@davidbeckingsale : I think I have something that in theory should exercise that, but will confirm for you. Might be nice if you had an existing project for me to clone and test this with.

cmake/SetupCompilerOptions.cmake
```diff
@@ -175,20 +175,33 @@ if(ENABLE_FORTRAN AND BLT_FORTRAN_FLAGS)
...
+############################################################
+# Map Legacy FindCUDA variables to native cmake variables
+# Note - we are intentionally breaking the semicolon delimited 
+# list that FindCUDA demanded of CUDA_NVCC_FLAGS so users
+# are forced to clean up their host configs.
+############################################################
+if (ENABLE_CUDA)
+   enable_language(CUDA) 
+   # where to find nvcc
+   if(CUDA_BIN_PATH)
```
white238: Please remove support for as many FindCUDA variables as possible.  I would like this as close to normal CMake conventions as possible.  This reduces the amount of upkeep we need to support going forward.  This applies to CUDA_BIN_PATH, CUDA_HOST_COMPILER, and CUDA_NVCC_FLAGS.

white238:
The existing CUDA examples will need to be switched as well as documentation updated.  I can help with any of that if you would like.

cmake/SetupCompilerOptions.cmake
robinson96: Regarding CUDA_SEPARABLE_COMPILATION - I think it's best practice to do this in general, I don't think there is harm to doing separable compilation but it's certainly a pitfall folks can encounter. Should I just do that by default for the sake of killing the CUDA_SEPARABLE_COMPILATION flag?

robinson96:
@davidbeckingsale : we are still using the CMAKE built in FindCUDA to set all the libraries and paths necessary for the cuda_runtime target. I've verified that that behavior is reasonable.

davidbeckingsale:
That’s fine with me. Is everyone okay with the require CMake version getting bumped up?

rhornung67:
@davidbeckingsale and @robinson96 you may want to check with @ajkunen about CMake versions. I recall him telling me that he had issues with some later versions.

cyrush:
@rhornung67 thanks for the warning, that rings with what I fear. 
I am hoping we can stick with 3.9.x.

rhornung67:
@cyrush my purpose is to ring your fears....

ajkunen:
I was having linking issues when using cmake 3.9.2, and had to back out to an early version.

I haven’t tried to figure out why… nor do I have the time.  It may be that we’re doing something stupid.


-Adam

davidbeckingsale:
Hmm, I'm sure you were doing something stupid 😄 did you ever try with a version > 3.9.2?

ajkunen:
Nope



-Adam

kennyweiss:
Thanks for fixing the host-configs @robinson96.  
Please close #146 when you merge.

PR #155, opened 2018-05-16
Unify source filtering to one macro

white238:
* Fix indenting in SetupBLT.cmake while I'm touching it anyways
* I ran into a problem with the regex to filter fortran sources out of cuda sources.  Instead of fixing the regex I unified the two places we were filtering sources.


cmake/BLTPrivateMacros.cmake
```diff
@@ -248,6 +244,46 @@ macro(blt_setup_cuda_source_properties)
...
+        if(${_ext} IN_LIST BLT_C_FILE_EXTS)
+            if (${arg_C_LIST})
+                list(APPEND ${arg_C_LIST} ${_file})
+            endif()
+        elseif(${_ext} IN_LIST BLT_Fortran_FILE_EXTS)
+            if (${arg_Fortran_LIST})
+                list(APPEND ${arg_Fortran_LIST} ${_file})
+            endif()
+        else()
+            message(FATAL_ERROR "blt_split_source_list_by_language given source file with unknown file extension. Add the missing extension to the corresponding list (BLT_C_FILE_EXTS or BLT_Fortran_FILE_EXTS).\n Unknown file: ${_file}")
```
kennyweiss: Would we ever have a file in the source list that wouldn't categorize into C or Fortran ?
I.e. should there be a third list variable for the extra files?
(Possibly with a warning/error when such files exist?)

cmake/BLTPrivateMacros.cmake
```diff
@@ -212,24 +213,19 @@ macro(blt_setup_cuda_source_properties)
...
-    foreach (_file ${arg_TARGET_SOURCES})
-        if (${_file} MATCHES "\\.(f|F)\\*")
-            set(_non_cuda_sources ${_non_cuda_sources} ${_file})
-        else()
-            set(_cuda_sources ${_cuda_sources} ${_file})
-        endif()
-    endforeach()
+    set(_cuda_sources)
+    set(_non_cuda_sources)
+    blt_split_source_list_by_language(SOURCES      ${arg_TARGET_SOURCES}
```
kennyweiss: Should we also handle cuda files (``*.cu``)?

cmake/BLTPrivateMacros.cmake
kennyweiss: If this is a ``FATAL_ERROR``, does that mean that we'd only see the first instance?

thirdparty_builtin/googletest-master-2018-02-01/googletest/CMakeLists.txt
```diff
@@ -27,7 +27,10 @@ option(
   "Build gtest with internal symbols hidden in shared libraries."
   OFF)
 
-set(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "Generate debug library name with a postfix.")
+# BLT FIX - This is causes global issues where it adds 'd' to the end of library names when
```
kennyweiss: Typo: "this is causes" -> "this causes"

docs/creating_execs_and_libs.rst
```diff
@@ -122,5 +122,5 @@ more work or CMake function calls.
    given sources and headers along with handling dependencies the same way as ``blt_add_executable``
    does.  It also provides a few commonly used build options, such as overriding the output name of the
    library and the output directory.  It defaults to building a static library unless you override it with
-   ``SHARED`` or with the global BLT option ``ENABLE_SHARED_LIBS``.
+   ``SHARED`` or with the global CMake option ``BUILD_SHARED_LIBS``.
```
kennyweiss: Should "option" be changed to "variable, e.g.: 
> ... with the global CMake **variable** ``BUILD_SHARED_LIBS``

cmake/BLTPrivateMacros.cmake
white238: Fixed.

cmake/BLTPrivateMacros.cmake
white238: yes it would error out on the first unknown file.

thirdparty_builtin/googletest-master-2018-02-01/googletest/CMakeLists.txt
white238: fixed.

docs/creating_execs_and_libs.rst
white238: I think option is fine.

Issue #72, opened 2017-08-18
consider blt macro that helps combine static libs


cyrush:
there are a compelling cases for collapsing symbols from all TPLS and your lib into a single static lib to make it easy for consumers to link your app.

I believe Axom wants this, Alpine would use it as well. 

Details still to be worked out.

cyrush:
Note: adding object lib support will help with this goal as well, but there are cases where object libs may not be practical:

* when using cuda
* when actually trying to re-package static TPLs

zbeekman:
A note on object libs and Fortran:

Object libs can be useful for avoiding race conditions in Fortran builds. A Fortran source file that defines a module that is listed as a source file for more than one target, whose builds are not ordered due to dependencies, can trigger a race condition resulting in a build failure during parallel compilation. The CMake generated build system may overwrite/delete the .mod file from one thread of the build while the other one still needs it. The best work around for this is to create object libs for all source files that declare a Fortran module. This ensures that the .mod file and object file are only generated once and then reused for all targets.

I've reported this to CMake, and vaguely remember them telling me this is not a bug, although I'm not convinced.

cyrush:
@zbeekman I have seen issues like this, but I didn't know the root cause. 
I always thought it was some issue with the compiler choosing colliding temporary names. 

Good to know that object libs avoids this issue.

cyrush:
@zbeekman, do you have a bug report or a link to the thread about this to the CMake folks?

zbeekman:
I asked again recently to confirm: https://gitlab.kitware.com/cmake/cmake/issues/17525

CMake's design is such that any source file that is listed as a part of multiple targets can expect to get recompiled for each target created through `add_library` or `add_executable`. 

Here is the discussion, feel free to weigh in: https://gitlab.kitware.com/cmake/cmake/issues/17525

cyrush:
@zbeekman thanks, good to get a (more) complete understanding of the issue!

rrsettgast:
@cyrush did that combine_static_libs macro take care of this?

white238:
This can be solved with object libraries.

Issue #109, opened 2017-10-30
Minimum cmake version for blt


kennyweiss:
We should define the minimum cmake version allowed for blt-based projects.  

This minimum version should be tested/checked within blt, and should be clearly indicated in our documentation.

One of our core macros used for transitive dependencies is using a cmake 3.3 feature:
```cmake
if(... IN_LIST ...)
```  
This feature also requires setting cmake policy ``CMP0057`` to NEW (as pointed out by @zbeekman in PR #108).

Note: Setting a value for ``CMP0057`` is only valid in Cmake versions above 3.3 and is required in versions above 3.3.2 when using ``if(.. IN_LIST ...)``. 

See: https://cmake.org/cmake/help/v3.3/policy/CMP0057.html

zbeekman:
Where would be the most appropriate place to set `CMP0057` to `NEW`. I'd like to include that in my PR (#108).

kennyweiss:
Thanks -- I'd think it should be around here -- https://github.com/LLNL/blt/blob/master/SetupBLT.cmake#L63 -- in ``SetupBLT.cmake``.

It should be guarded by the version until we decide about adding a ``cmake_minimum_version`` to blt (and which version that should be), since earlier cmake versions will complain about unknown policies.
E.g.
```cmake
if("${CMAKE_VERSION}" VERSION_GREATER 3.3)
    cmake_policy(SET CMP0057 NEW)
endif()
```

zbeekman:
I agree, but how do you propose working around IN_LIST? Without setting
CMP0057 to new, using IN_LIST is essentially a syntax error. I'd rather
have the unknown policy complaint to warn me that something may break. The
alternative is to proceed anyway in previous versions of CMake that don't
support IN_LIST at all. In fact, it caused CI failures when I introduced
IN_LIST in my PR, and I'm surprised the other instance of IN_LIST hasn't
caused a problem yet.

kennyweiss:
I think that it is feature is not yet in our blt tests. 
It comes up when there is a transitive dependency in the ``DEPENDS_ON`` lists.

Perhaps we should avoid resolving the policy part of this problem within BLT until the ``cmake_minimum_version`` is decided.   In the meantime, it is easy to add both a minimum version and the cmake policy part to your project's build system.

zbeekman:
[...]

PR #157, opened 2018-05-19
Combine static libraries

rrsettgast:
This macro combines static libraries into a single library. 

Issue #72


cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,53 @@ macro(blt_find_libraries)
...
+        list( APPEND libLinkLine ${arg_LINK_PREPEND} ${lib} ${arg_LINK_POSTPEND} )
+    endforeach()
+
+    add_library ( ${arg_NAME} SHARED )
+    target_link_libraries( ${arg_NAME} PRIVATE ${libLinkLine})
+    blt_register_library( NAME ${arg_NAME}
+                          DEPENDS_ON ${arg_SOURCE_LIBS}
+                          LIBRARIES ${arg_NAME}
+                         )
+
```
kennyweiss: Please unset the ``libLinkLine`` variable here.

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,53 @@ macro(blt_find_libraries)
...
+    if( NOT arg_SOURCE_LIBS )
+        message(FATAL_ERROR "blt_combine_static_libraries(NAME ${arg_NAME} ...) called with no given source libraries")
+    endif()
+
+    set( libLinkLine "" )
+    foreach( lib ${arg_SOURCE_LIBS} )
+        list( APPEND libLinkLine ${arg_LINK_PREPEND} ${lib} ${arg_LINK_POSTPEND} )
+    endforeach()
+
+    add_library ( ${arg_NAME} SHARED )
```
kennyweiss: Does this only work for building a ``SHARED`` library?

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,53 @@ macro(blt_find_libraries)
...
+    target_link_libraries( ${arg_NAME} PRIVATE ${libLinkLine})
```
kennyweiss: Is this making ``${arg_NAME}`` depend on a string containing link flags and library names? That doesn't seem very ``blt``-esque.  

Pretty much every other part of blt explicitly lists the flags in a separate flags variable.  Is that possible here?

white238:
Randy and I have talked about this pretty extensively.  This was more of a proof of concept for a asked for feature that needs to be explored if it can make it more general.

As for object libraries, it would be very nice for BLT to support them but there will need to be a lot of logic needed to hide the grossness of CMake's object library "support".  The two things I can think of without further explorations is:

1) This syntax needs to be hidden from users: $<TARGET_OBJECTS:foo> should be just foo.
2) There is no transitive include directories for object libraries.  It is lost unless you just through hoops.  Here is a page that explains https://stackoverflow.com/questions/38832528/transitive-target-include-directories-on-object-libraries

cmake/BLTMacros.cmake
rrsettgast: I needed a shared library so that is what have in there. It would work with a static as well. We could just pass it the argument.

cmake/BLTMacros.cmake
rrsettgast: It would be nice, however, in this case you have to have the link flags prepend and postpend the libraries. So on a linux system, if you want to get all the symbols for libfoo.a , the link line would have to be 
"-Wl,--whole-archive -lfoo -Wl,--no-whole-archive"
Without the postpend, the link flag applies to all libs that follow on the link line...bad news bears.

now you don't have to wrap each lib individually but a block of them instead, but on MacOSX there is only a
"-Wl,--force-load" prepend that applies only to the single lib that follows the command.

I really don't like the way this is done, but I didn't really see an alternative. Any ideas?

cmake/BLTMacros.cmake
rrsettgast: no problem...I don't think I understand scoping of cmake macros :)

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,53 @@ macro(blt_find_libraries)
...
+## blt_combine_static_libraries( NAME <libname>
+##                               SOURCE_LIBS [lib1 ...] 
+##                               LINK_PREPEND []
+##                               LINK_POSTPEND []
+##                 )
+##
+## Adds a library target, called <libname>, to be built from the set of 
+## static libraries given in SOURCE_LIBS.
+## 
+## The LINK_PREFIX argument will be prepended to the library on the link line,
```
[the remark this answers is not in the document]
rrsettgast: fixed

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,53 @@ macro(blt_find_libraries)
...
+
+    cmake_parse_arguments(arg
+        "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN} )
+
+    # sanity checks
+    if( "${arg_NAME}" STREQUAL "" )
+        message(FATAL_ERROR "blt_combine_static_libraries() must be called with argument NAME <name>")
+    endif()
+
+
```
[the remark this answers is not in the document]
rrsettgast: done

cyrush:
@white238 
Yes, i'll admit I am using the generator syntax for my object libs and it looks like voodoo. 

This interesting approach, have you got to the point where you can verify the linker is pulling in all of the expected symbols into the new static lib? Not sure it will pull things in if dummy.cpp doesn't ref anything? I would expect the linker to be "smart" and skip them if nothing is used from the lib.

rrsettgast:
@cyrush The prepend/postpend linker arguments are typically "-Wl,--whole-archive" and "-Wl,--no-whole-archive", which is intended to force the linker to include all symbols in the object/archive/static-lib that follow the prepend. We are using this now in geos code to link together all components into a single dynamic lib. The linker would indeed skip all unused symbols without them. Perhaps we can default the prepend/postpend variables to the appropriate platform specific values. It appears to work without issue, although I don't know what other test cases would be a good exercise of the reliability of this approach.

cyrush:
@rrsettgast Ok, I now understand much better -- this seems like a promising approach.

davidbeckingsale:
@rrsettgast - does this approach work to build an über static library, or only to bundle static libraries up into a shared lib?

rrsettgast:
A question is if we want to provide defaults for the prepend and postpend? I mean there are really only two options based on platform (LINUX, and OS X), so we could allow these arguments to be blank, and default them based on the platform.

davidbeckingsale:
Windows?

davidbeckingsale:
I would ask @robinson96 about the Windows stuff...

rrsettgast:
@davidbeckingsale , Peter said that they are not yet using blt on windows. Do you think it is acceptable to push off the windows support until there is a windows user to test this stuff out on?

kennyweiss:
Axom is using blt on Windows.  I think it would be ok to have this feature in blt that does not yet work with Windows as long as this is documented.

Alternatively, we could add an ``internal`` test to blt that exercises this feature and run it through appveyor.

rrsettgast:
@kennyweiss  how about a test where we take all of the axom libs and smash them together?

tests/internal/src/combine_static_library_test/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+                 SOURCES Foo1.cpp
+                 INCLUDES Foo1.hpp
+               )
+           
+blt_add_library( NAME Foo23
+                 SOURCES Foo2.cpp Foo3.cpp
+                 INCLUDES Foo2.hpp Foo3.hpp
+               )
+           
+blt_combine_static_libraries( NAME FooStatic
```
kennyweiss: Are you sure these are actually being built in ``SHARED`` and ``STATIC`` mode?  See my comment above.

kennyweiss:
@rrsettgast  -- Re: smashing the axom libraries together:

I think that would be a good test.  I think there is an (old) axom ticket to do exactly that.

kennyweiss:
Re: Windows:

This link seems relevant: https://stackoverflow.com/questions/3867254/what-is-the-microsoft-visual-studio-equivalent-to-gcc-ld-option-whole-archive

The equivalent appears to either be ``/WHOLEARCHIVE`` or ``/INCLUDE`` (or to use cmake ``OBJECT`` libraries)

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,66 @@ macro(blt_find_libraries)
...
+    if( NOT arg_SOURCE_LIBS )
+        message(FATAL_ERROR "blt_combine_static_libraries(NAME ${arg_NAME} ...) called with no given source libraries")
+    endif()
+
+    set( libLinkLine "" )
+    foreach( lib ${arg_SOURCE_LIBS} )
+        list( APPEND libLinkLine ${arg_LINK_PREPEND} ${lib} ${arg_LINK_POSTPEND} )
+    endforeach()
+
+    if( ${arg_LIB_TYPE} STREQUAL "STATIC" )
```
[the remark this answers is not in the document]
rrsettgast: Yeah. I had thought so, but it appears to function properly. Of course, if we expect that it shouldn't function properly, but does anyways, we can always do the if block.

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,66 @@ macro(blt_find_libraries)
...
+        set( _lib_type STATIC )
+    elseif( ${arg_LIB_TYPE} STREQUAL "SHARED" ) 
+        set( _lib_type SHARED )
+    else()
+        message(FATAL_ERROR "blt_combine_static_libraries(NAME ${arg_NAME} ...) LIB_TYPE must be SHARED OR STATIC")
+    endif()
+        
+    add_library ( ${arg_NAME} ${_lib_type} ${BLT_ROOT_DIR}/tests/internal/src/combine_static_library_test/dummy.cpp)
```
kennyweiss: This line doesn't belong here in the ``blt_combine_static_libraries`` macro:
```cmake
${BLT_ROOT_DIR}/tests/internal/src/combine_static_library_test/dummy.cpp
```

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,66 @@ macro(blt_find_libraries)
...
+    target_link_libraries( ${arg_NAME} PRIVATE ${libLinkLine})
+    blt_register_library( NAME ${arg_NAME}
```
kennyweiss: For some reason, this doesn't seem to be handling the transitive include dependencies correctly.

I configured cmake-3.9.2 using:
```bash
cmake  \ 
   -C ../host-configs/llnl-quartz-toss3-gcc@4.9.3.cmake  \
   -DCMAKE_EXPORT_COMPILE_COMMANDS=ON  \ 
   ../tests/internal/src/combine_static_library_test/
```
Here is the output when building ``FooStatic.x``:
```bash
[ 88%] Building CXX object CMakeFiles/FooStatic.x.dir/main.cpp.o
/usr/tce/packages/gcc/gcc-4.9.3/bin/g++    \
    -Wall -Wextra  -fPIE   -std=c++11 \
    -o CMakeFiles/FooStatic.x.dir/main.cpp.o \
    -c <_to_blt>/tests/internal/src/combine_static_library_test/main.cpp
<_to_blt>/tests/internal/src/combine_static_library_test/main.cpp:2:20: fatal error: foo1.hpp: No such file or directory
 #include "foo1.hpp"      
```
Note that there is no include flag to the path of ``"foo1.hpp"``

rrsettgast: sorry. I hadn't run it on unix. MacOSX filesystem isn't case sensitive....so it found the files that were upper case.

kennyweiss: Thanks for looking into it.  That makes sense.
After you fix this, could you please add this new test project to the travis plan?

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,66 @@ macro(blt_find_libraries)
...
+##                             )
+##
+## Adds a library target, called <libname>, to be built from the set of 
+## static libraries given in SOURCE_LIBS.
+## 
+## The LINK_PREPEND argument will be prepended to the library on the link line,
+## while the LINK_POSTPEND will be appended to the libray on the link line. These
+## values are typically "-Wl,--whole-archive", and "-Wl,--no-whole-archive" 
+## on linux systems
+##
```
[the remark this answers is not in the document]
rrsettgast: ok

tests/internal/src/combine_static_library_test/CMakeLists.txt
rrsettgast: yes

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,101 @@ macro(blt_find_libraries)
...
+    if( ${arg_LIB_TYPE} STREQUAL "STATIC" )
+        set( _lib_type STATIC )
+    elseif( ${arg_LIB_TYPE} STREQUAL "SHARED" ) 
+        set( _lib_type SHARED )
+    else()
+        message(FATAL_ERROR "blt_combine_static_libraries(NAME ${arg_NAME} ...) LIB_TYPE must be SHARED OR STATIC")
+    endif()
+    
+    add_library ( ${arg_NAME} ${_lib_type} ${BLT_ROOT_DIR}/tests/internal/src/combine_static_library_test/dummy.cpp)
+    target_link_libraries( ${arg_NAME} PRIVATE ${libLinkLine})
```
kennyweiss: @rrsettgast -- this might need to be quoted to avoid cmake changing things. I.e. 
```cmake
target_link_libraries( ${arg_NAME} PRIVATE "${libLinkLine}") 
```

kennyweiss:
That's a weird one.  I added a comment on a possible culprit.  If that doesn't resolve it, we can dive deeper.

cmake/BLTMacros.cmake
kennyweiss: Looks like a weird cmake thing.

According to this link (https://stackoverflow.com/a/37204841), we should be able to resolve the issue by replacing the forward slash with a dash (``/`` ->  ``-``) as both are acceptable to MSVC.

See also: https://github.com/conan-io/conan/issues/861

kennyweiss: What do you think @white238?

rrsettgast:
Yay. Victory. Anybody have time to complete a review or two?

SetupBLT.cmake
```diff
@@ -66,6 +66,12 @@ if (NOT BLT_LOADED)
     ################################
     include(${BLT_ROOT_DIR}/cmake/BLTOptions.cmake)
 
+    ################################
+    # CMP0057 New is required by blt_add_libraries
```
white238: Can you make this comment that it turns on the IN_LIST operator

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,101 @@ macro(blt_find_libraries)
     endforeach()
 
 endmacro(blt_find_libraries)
+
+
+##------------------------------------------------------------------------------
+## blt_combine_static_libraries( NAME <libname>
+##                               SOURCE_LIBS [lib1 ...] 
+##                               LIB_TYPE [STATIC,SHARED}
```
white238: Change } to ]

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,101 @@ macro(blt_find_libraries)
...
+##                               LINK_PREPEND []
+##                               LINK_POSTPEND []
+##                             )
+##
+## Adds a library target, called <libname>, to be built from the set of 
+## static libraries given in SOURCE_LIBS.
+## 
+## The LINK_PREPEND argument will be prepended to the library on the link line,
+## while the LINK_POSTPEND will be appended to the libray on the link line. These
```
white238: libray to library

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,101 @@ macro(blt_find_libraries)
...
+## values are defaulted to the appropriate values for CMAKE_HOST_APPLE and 
+## CMAKE_HOST_UNIX. CMAKE_HOST_WIN32 is not yet supported.
```
white238: WIN32 is now supported!

cmake/BLTMacros.cmake
```diff
@@ -882,3 +882,101 @@ macro(blt_find_libraries)
...
+    cmake_parse_arguments(arg
+        "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN} )
+
+    # sanity checks
+    if( "${arg_NAME}" STREQUAL "" )
+        message(FATAL_ERROR "blt_combine_static_libraries() must be called with argument NAME <name>")
+    endif()
+
+    if(TARGET ${arg_NAME})
+        message(FATAL_ERROR "$A target with the name {arg_NAME} already exists!")
```
white238: Please fix this error message

cmake/BLTMacros.cmake
white238: CMake macros just inject the code directly where it is called so it uses the scope of the parent.

SetupBLT.cmake
cyrush: Note:
You won't see this problem if the host project sets the cmake_min_version to a newish version of cmake. 
I suppose it can't hurt to set it in blt as well, but we haven't set any others yet.

rrsettgast: Yeah. blt/tests/internal/CMakeLists.txt has:
cmake_minimum_required(VERSION 3.1)

This message is actually wrong. It is blt_setup_target() that has the issue.
```
CMake Warning (dev) at /Users/settgast1/Codes/geosx/blt/cmake/BLTPrivateMacros.cmake:139 (if):
  Policy CMP0057 is not set: Support new IN_LIST if() operator.  Run "cmake
  --help-policy CMP0057" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  IN_LIST will be interpreted as an operator when the policy is set to NEW.
  Since the policy is not set the OLD behavior will be used.
  /Users/settgast1/Codes/geosx/blt/cmake/BLTMacros.cmake:582 (blt_setup_target)
  src/combine_static_library_test/CMakeLists.txt:33 (blt_add_executable)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at /Users/settgast1/Codes/geosx/blt/cmake/BLTPrivateMacros.cmake:139 (if):
  if given arguments:

    "NOT" "Foo1" "IN_LIST" "_expanded_DEPENDS_ON"

  Unknown arguments specified
  /Users/settgast1/Codes/geosx/blt/cmake/BLTMacros.cmake:582 (blt_setup_target)
  src/combine_static_library_test/CMakeLists.txt:33 (blt_add_executable)
```

cmake/BLTMacros.cmake
rrsettgast: done

cmake/BLTMacros.cmake
rrsettgast: done

cmake/BLTMacros.cmake
rrsettgast: removed

cmake/BLTMacros.cmake
rrsettgast: haha. fixed

SetupBLT.cmake
cyrush: given that we don't know exactly which min version to recommend (i am using 3.9, but that might be too stringent), I think it makes sense to set the policy here to avoid headaches

SetupBLT.cmake
```diff
@@ -66,6 +66,13 @@ if (NOT BLT_LOADED)
     ################################
     include(${BLT_ROOT_DIR}/cmake/BLTOptions.cmake)
 
+    ################################
+    # CMP0057 New is required by blt_setup_target()
+    # in order to support the IN_LIST if() operator
+    ################################
+    cmake_policy(SET CMP0057 NEW)
```
kennyweiss: I think this either does not belong here, or it must be guarded.  

If left as is, it will trigger an error in earlier versions of cmake.  See https://cmake.org/cmake/help/v3.12/manual/cmake-policies.7.html

Specifically, if we add it here, please change it to:
```cmake
if(POLICY CMP0057)
  cmake_policy(SET CMP0057 NEW)
endif()
```

tests/internal/src/combine_static_library_test/Foo1.cpp
```diff
@@ -0,0 +1,30 @@
+/*
+ * Foo.cpp
+ *
+ *  Created on: May 23, 2018
```
kennyweiss: We don't typically include author/creation dates in blt source files.  

grepping through the code, I saw one exception.  
I assume this came from your file editor.
Would you mind removing it from here and elsewhere? (not strictly necessary if you feel strongly about it)

SetupBLT.cmake
kennyweiss: Ticket #109 is relevant.  

Perhaps we should move logic related to setting appropriate policies (based on the cmake version) into its own file.

rrsettgast: I think the more alarming situation will be when using cmake version that don't support CMP0057, any blt macro that calls blt_setup_target() using a list of dependencies will fail. For instance 
```
blt_add_library( 
NAME foo
SOURCES bar1.cpp bar2.cpp
)
```
I think that ticket #109 addresses this in more detail. Can we leave this here for now so that the CI tests can pass and wait for resolution on #109 ?

kennyweiss: That's fine, but I don't see the downside in guarding it as in my snippet above.  It shouldn't break any current tests since it will always set the policy when that policy is defined.

I don't think we've addressed the question of the minimum cmake version that blt supports.  Certain features require newer versions (e.g. for cuda support).

I think the ``IN_LIST`` thing will only bite users of older cmake versions when they use ``blt_add_library`` and the added library has dependencies.

rrsettgast:
@white238 @kennyweiss @cyrush 
It seems like there are somewhat random appveyor failures due to connection issues. Is this ok to ignore, and can I merge the PR?

kennyweiss:
We've been having a weird problem downloading CMake on Appveyor jobs sometimes.  It is ok with me to merge since all the other tests are passing and this is unrelated.

cyrush:
@rrsettgast yes go ahead. 

Thanks for tackling this!

PR #164, opened 2018-06-21
Make finding TPL executables more robust

white238:
A user ran into an odd case where uncrustify was on his path during the CMake call but not the make call.  This fixes it by if you don't have UNCRUSTIFY_EXECUTABLE defined it will find it by find_program and store the full path for the make call.  It also adds ENABLE_UNCRUSTIFY (defaulted to ON).  It only spits out a message if it can't find uncrustify on your path and you don't turn off uncrustify support.

This was done for all TPL executables that follow this usage (6 in total).


cmake/thirdparty/FindCppcheck.cmake
```diff
@@ -50,14 +50,19 @@
 # CPPCHECK_EXECUTABLE
 ################################################################################
 
-find_program(CPPCHECK_EXECUTABLE
-             NAMES cppcheck
-             DOC "Path to cppcheck executable")
+if (CPPCHECK_EXECUTABLE)
```
kennyweiss: @white238 -- It looks like you are using the same logic for all these.  Can you convert this to a (private) macro to ensure consistency?

cmake/BLTOptions.cmake
```diff
@@ -51,9 +51,14 @@ option(ENABLE_COVERAGE   "Enables code coverage support" OFF)
 
 
 ################################
-# TPL Options
+# TPL Executable Options
 ################################
-option(ENABLE_GIT       "Enables Git support" ON)
+option(ENABLE_CPPCHECK     "Enables Cppcheck support" ON)
```
kennyweiss: The PR description makes it seem like this is an edge-case (i.e. wanting to use a system version of these applications, but only when they are on your path)

Perhaps these should be off by default, or controlled by the presence of ``${FOO}_EXECUTABLE``.  

Specifically, I am wondering if this feature should be opt-in rather than opt-out.

kennyweiss: Note: I am fine with either decision (opt-in or opt-out), but wanted to raise the question.

cmake/thirdparty/FindCppcheck.cmake
white238: I tried to make this into a nice general macro and it went poorly (too many things to guard against and cover).  I could definitely make it a private macro that only covers our one case.

cmake/BLTOptions.cmake
white238: controlling it based on the executable path variable is exactly what this PR is trying to fix.  We spoke about this last week when i stated it was not fully qualified during the cmake step (only on the path) and then not on the path during make.

white238: As for opt-in vs opt-out, it doesnt cause any problems if it cant find the executable (one very small line is output) and just creates some targets.

cmake/thirdparty/FindCppcheck.cmake
kennyweiss: A single purpose private macro sounds good.  (Or leave it for the future if you think it would be too messy).

cmake/BLTOptions.cmake
kennyweiss: Sorry if I wasn't clear. 

For the opt-in part, I was suggesting that if users want the new behavior and can't (or would prefer not to) explicitly supply the executable path, they should set ``ENABLE_UNCRUSTIFY = ON`` when they configure cmake.

I am fine with leaving it as opt-out.

cmake/thirdparty/SetupThirdParty.cmake
```diff
@@ -88,30 +88,27 @@ endif()
 ################################
 # Documentation Packages
 ################################
-if (DOXYGEN_EXECUTABLE)
+if (ENABLE_DOXYGEN)
```
kennyweiss: Doxygen doesn't have the same logic as the others?

Issue #165, opened 2018-06-25
BLT shouldn't set -DUSE_MPI


cyrush:
Projects use all kinds of flags to indicate the presence of mpi

We should remove this from 

https://github.com/LLNL/blt/blob/master/cmake/thirdparty/SetupMPI.cmake

and defer to what users want to do downstream.

rrsettgast:
Just to clarify, is the desire to remove the -DUSE_MPI flag from inclusion in the compilation flags, or undefine the USE_MPI cmake variable?

cyrush:
resolved in #167

PR #167, opened 2018-06-27
remove USE_MPI and USE_OPENMP defines

### LINKED ISSUE #7: revisit use of USE_OPENMP and USE_MPI defines
Since it is easy for users of BLT to add their own compiler defines for OpenMP or MPI targets, we wanted to discuss not adding them via blt_register_lib for openmp and mpi.

cyrush:
resolves #7  #165


white238:
you can't use blt_add_target_definitions on MPI because it is not a true CMake target (its one of our fake register targets).  Though we probably should augment our macros to support our various add macros.

gzagaris:
I believe it is possible to specify target definitions in place when calling `blt_add_library` and `blt_add_executable` . 

For example:
```
blt_add_library( foo
                           DEPENDS_ON mpi
                           DEFINES USE_MPI  )
```

white238:
George, Yes it is but not blt_add_target_definitions or blt_add_target_compile_flags and the like

gzagaris:
I understand that you cannot use `blt_add_target_definitions` etc. on the pseudo-targets that BLT creates like `mpi`, but you can do the following:

```
blt_add_library( foo DEPENDS_ON mpi )
blt_add_target_definitions( TO foo TARGET_DEFINITIONS USE_MPI ) 
```

Right?

gzagaris:
That should be equivalent with defining it in-place as mentioned earlier.

white238:
yes

gzagaris:
Thanks for clarifying!

kennyweiss:
@gzagaris -- yes, you can add a define to an individual target, but you have to remember to do that to every single MPI-dependent target.  I was hoping that we could attach it to the pseudo-targets so it could automatically be attached to all targets that depend on ``mpi``.  

As @white238 mentioned, this is a behavior that we could add to ``blt_add_target_definitions``

rrsettgast:
Is this PR going to be merged? I have a PR in geos that is hanging around waiting for this one and I would like to get it merged.

cyrush:
@rrsettgast yes, sorry it feel off my radar. I'll work on getting it merged

PR #176, opened 2018-08-30
Bugfix for nvcc builds on blueos

kennyweiss:
Otherwise, nvcc complains about redefining std:
```bash
nvcc fatal   : redefinition of argument 'std'
```

Looks like the previous attempt to resolve this no longer works (See #154).


kennyweiss:
Note: Due to an updated MPI installation, which is automatically added to user paths, I had to load the following two modules to pass the ``blt_mpi_smoke`` test on this platform:
```bash
ml clang/coral-2018.04.17
ml spectrum-mpi/2017.04.03
```

PR #177, opened 2018-08-31
 blt_add_target_compile_flags can now append flags to INTERFACE targets

kennyweiss:


kennyweiss:
I tested this within BLT (on the toss3 gcc host config) by adding the following flags in ``tests/internal/CMakeLists.txt``:
```cmake
# Add flag to a regular library target
blt_add_target_compile_flags(TO example FLAGS " -finstrument-functions")
# Add list of flags to a library target
blt_add_target_compile_flags(TO example FLAGS "-Wundef" "-ftrapv")

# Add list of flags to an interface target (in two ways)
blt_add_target_compile_flags(TO blt_header_only FLAGS "-Wswitch-enum" "-Wfloat-equal")
blt_add_target_compile_flags(TO blt_header_only FLAGS -finstrument-functions -ftrapv)
```

I confirmed that the new macro doesn't overwrite the flags on the second call and that flags are transitively sent to dependents.

I am not planning to commit these lines to the blt internal tests since the added flags are meaningless in this context and I did not define them for other compilers.

kennyweiss:
We can also apply similar logic to ``blt_add_target_definitions`` and ``blt_add_target_link_flags`` to make these work with ``INTERFACE`` targets.

PR #179, opened 2018-08-31
Feature/add windows host config

kennyweiss:
Adds a host-config for windows.  

~~Looks like we haven't been exercising the blt internal tests for a while.~~

There was a linker error in the shared version of the ``blt_combine_static_libraries`` example, so I temporarily disabled the example on Windows.  A future PR will address this.
> LINK : fatal error LNK1356: cannot find library `Foo1.lib' specified to be whole archived [<blt>\build-win\src\combine_static_library_test\FooShared.vcxproj]

UPDATE: The appveyor script was fine -- I missed a line where the internal tests are copied.

### DIFF
--- a/tests/internal/src/combine_static_library_test/CMakeLists.txt
+++ b/tests/internal/src/combine_static_library_test/CMakeLists.txt
@@ -1,8 +1,20 @@
- 
-if(NOT BLT_SOURCE_DIR)
-    set(BLT_SOURCE_DIR "${PROJECT_SOURCE_DIR}/../../../..")
-endif()
-
+###############################################################################
+# Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
+#
+# Produced at the Lawrence Livermore National Laboratory
+#
+# LLNL-CODE-725085
+#
+# All rights reserved.
+#
+# This file is part of BLT.
+#
+# For additional details, please also read BLT/LICENSE.
+###############################################################################
+#
+# Simple example that uses the blt_combine_static_libraries macro
+#
+###############################################################################
 
 blt_add_library( NAME Foo1
                  SOURCES Foo1.cpp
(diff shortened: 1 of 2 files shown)


tests/internal/src/combine_static_library_test/CMakeLists.txt
```diff
@@ -23,27 +35,28 @@ blt_combine_static_libraries( NAME FooStatic
...
+    blt_combine_static_libraries( NAME FooShared
+                       SOURCE_LIBS Foo1 Foo23
+                       LIB_TYPE SHARED
+                       LINK_PREPEND ""
+                       LINK_POSTPEND ""
+                     )
+
+    blt_add_executable( NAME blt_combine_static_libraries_shared_smoke
                     SOURCES blt_combine_static_libraries_shared_smoke.cpp
                     DEPENDS_ON FooStatic gtest
```
kennyweiss: It seems like this should depend on ``FooShared``, right?

tests/internal/src/combine_static_library_test/CMakeLists.txt
```diff
@@ -23,27 +35,28 @@ blt_combine_static_libraries( NAME FooStatic
...
                     SOURCES blt_combine_static_libraries_static_smoke.cpp
                     DEPENDS_ON FooStatic gtest
                   )
-
 blt_add_test( NAME blt_combine_static_libraries_static_smoke
               COMMAND blt_combine_static_libraries_static_smoke
             )
 
-blt_add_executable( NAME blt_combine_static_libraries_shared_smoke
+if(NOT WIN32)
```
kennyweiss: I disabled this test on Windows due to a linker error.

Due to the appveyor path mismatch above, it seems likely that this was never exercised.

tests/internal/src/combine_static_library_test/CMakeLists.txt
white238: The combine macro is unfortunately broken.  I have a new version from Burl that needs to be reviewed but I am pretty sure it works and solves the build dependency problem as well.

PR #184, opened 2018-09-05
fix blt_combine_static_libraries

rrsettgast:
modified blt_combine_static_libraries macro to transfer include directories but not link libraries

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -1011,8 +1011,6 @@ macro(blt_combine_static_libraries)
             list( APPEND libLinkLine "${_link_prepend}${lib}" )
         endif()
     endforeach()
-
-    message( "libLinkLine = ${libLinkLine}" )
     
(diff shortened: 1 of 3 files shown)


white238:
We may want to add a temporary guard against windows in this macro and disable the windows test for now.  @kennyweiss thoughts?

kennyweiss:
I suggested the same to @rrsettgast yesterday.
If you do this, please add a note to the function's documentation that this function does not currently work on Windows.

cmake/BLTMacros.cmake
```diff
@@ -951,6 +951,8 @@ endmacro(blt_find_libraries)
 ## These values are defaulted to the appropriate values for CMAKE_HOST_APPLE and 
 ## CMAKE_HOST_UNIX.
 ##
+## Note: This macro does not currently work for Windows
```
kennyweiss: Thanks

PR #185, opened 2018-09-05
Add .C/.H file extensions

jdahm:
This minor addition allows proper detection of .C and .H inside BLT.

### DIFF
--- a/SetupBLT.cmake
+++ b/SetupBLT.cmake
@@ -167,7 +167,7 @@ if (NOT BLT_LOADED)
     # Global variables needed by BLT
     #
     ################################
-    set(BLT_C_FILE_EXTS ".C" ".cpp" ".hpp" ".cxx" ".hxx" ".cc" ".c" ".h" ".hh" ".inl" ".cu"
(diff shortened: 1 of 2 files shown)


SetupBLT.cmake
```diff
@@ -167,7 +167,7 @@ if (NOT BLT_LOADED)
     # Global variables needed by BLT
     #
     ################################
-    set(BLT_C_FILE_EXTS ".cpp" ".hpp" ".cxx" ".hxx" ".cc" ".c" ".h" ".hh" ".inl" ".cu"
+    set(BLT_C_FILE_EXTS ".cpp" ".hpp" ".cxx" ".hxx" ".c" ".h" ".C" ".H" ".cc" ".hh" ".inl" ".cu"
```
kennyweiss: A different, possibly more general, solution would be to keep all file extensions in these lists lower case and to check against the lower case version of the file in the ``blt_split_source_list_by_language`` macro.

kennyweiss:
Thanks @jdahm

PR #187, opened 2018-09-06
ENH: add convenience blt_list_append macro

gzagaris:
This commit adds a blt_list_append() macro, which is
essentially a wrapper around CMake's list(APPPEND...)
that allows to inline a conditional check for convenience
and clarity.

As an example, code that looks like this:
```
set(mylist)

if ( ${USE_FOO} )
  list(APPEND mylist foo)
endif()

if ( ${USE_BAR} )
  list(APPEND mylist bar)
endif()
```

Can be simplified using the blt_list_append() macro like this:
```
set(mylist)
blt_list_append( TO mylist ELEMENTS foo IF ${USE_FOO} )
blt_list_append( TO mylist ELEMENTS bar IF ${USE_BAR} )
```

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -57,6 +57,9 @@ include(${BLT_ROOT_DIR}/cmake/BLTPrivateMacros.cmake)
 ##   (2) A condition to check by passing IF <bool>
 ##   (3) The list of elements to append by passing ELEMENTS [<element>...]
 ##
+## Note, the argument passed to the IF option has to be a single boolean value
+## and cannot be a boolean expression since CMake cannot evaluate those inline.
+##
 ## Usage Example:
 ##
 ##  set(mylist A B)
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -44,6 +44,52 @@
 include(${BLT_ROOT_DIR}/cmake/BLTPrivateMacros.cmake)
 
 ##------------------------------------------------------------------------------
+## blt_list_append( TO <list> ELEMENTS [ <element>...] IF <bool> )
+##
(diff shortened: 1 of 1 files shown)


cmake/BLTMacros.cmake
```diff
@@ -43,6 +43,52 @@
...
+##
+## Appends elements to a list if the specified bool evaluates to true.
+##
+## This macro is essentially a wrapper around CMake's `list(APPEND ...)`
+## command which allows inlining a conditional check within the same call
+## for clarity and convenience.
+##
+## This macro requires specifying:
+##   (1) The target list to append to by passing TO <list>
+##   (2) A condition to check by passing IF <bool>
```
white238: Please make a note that the IF argument cannot handle anything other than a single variable.  For example, the following won't work:

blt_list_append( TO mylist ELEMENTS C
                            IF NOT ${ENABLE_C} )

gzagaris: Yes, good point. I'll add a comment. 

As a side note, we might be able to make use of the boolean generator expression $<BOOL: NOT ${ENABLE_C}> to inline evaluate a boolean expression and pass the result to the macro. I'll give this a shot and add comments accordingly.

kennyweiss: I like that idea

white238: Another option is to add an IF_NOT. Though this is a pretty slippery slope. IF_AND, IF_OR.. where does it stop.

gzagaris: Yeah, it's a rabbit hole. Unfortunately, the `$<BOOL: ...>` doesn't work in this context either. I added a clarification to the documentation for the macro.

PR #189, opened 2018-09-16
Adds support for code formatting with AStyle

kennyweiss:
Details:
* AStyle project link -- http://astyle.sourceforge.net
* Adds option 'ENABLE_ASTYLE' to control whether to search for AStyle.
   As with uncrustify, users can specify an explicit 'ASTYLE_EXECUTABLE' if they do not want to use the one in their path.
* Adds a macro 'blt_add_astyle_target' that can be directly invoked
* Support for Astyle formatting has also been added to the 'blt_add_code_checks' macro, where users can supply an 'ASTYLE_CFG_FILE' rather than an 'UNCRUSTIFY_CFG_FILE' to use AStyle.
* Discovered that earlier versions of Astyle and Uncrustify do not support the dry run option ('MODIFY_FILES' = False), and added comments to the appropriate macros.
   Specifically, Uncrustify v0.61 and later and AStyle v2.05 and later support this option, but earlier versions do not.  BLT does not currently check for this.

Update:
* The AStyle and Uncrustify ``check`` targets are now only generated when supported in their corresponding program.
* Added a wrapper script to run AStyle with return codes indicating success or failure.

### DIFF
--- a/cmake/SetupCodeChecks.cmake
+++ b/cmake/SetupCodeChecks.cmake
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+# Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
 #
 # Produced at the Lawrence Livermore National Laboratory
 #
@@ -450,15 +450,23 @@ macro(blt_add_astyle_target)
     endif()
 
     if(_generate_target)
-        add_custom_target(${arg_NAME}
-                COMMAND ${ASTYLE_EXECUTABLE} ${arg_PREPEND_FLAGS}
-                        --options=${arg_CFG_FILE} --formatted 
-                        ${MODIFY_FILES_FLAG} ${arg_SRC_FILES} ${arg_APPEND_FLAGS}
-                WORKING_DIRECTORY ${_wd} 
-                VERBATIM
-                COMMENT "${arg_COMMENT}Running AStyle source code formatting checks.")
+
+        # AStyle doesn't report failure when there are files that require formatting.
+        # Fix this with a wrapper script that parses the output.
+        set(wrapped_astyle_script ${CMAKE_CURRENT_BINARY_DIR}/WrapAstyle_${arg_NAME}.cmake)
+
+        configure_file(
+            ${BLT_ROOT_DIR}/cmake/WrapAstyle.cmake.in
+            ${wrapped_astyle_script}
+            @ONLY )
+
+        add_custom_target(
+            ${arg_NAME}
+            COMMAND ${CMAKE_COMMAND} -P ${wrapped_astyle_script}
+            WORKING_DIRECTORY ${_wd}
+            COMMENT "${arg_COMMENT}Running AStyle source code formatting checks.")
         
-        # hook our new target into the proper dependency chain
+        # Hook our new target into the proper dependency chain
         if(${arg_MODIFY_FILES})
             add_dependencies(astyle_style ${arg_NAME})
         else()
--- a/cmake/SetupCodeChecks.cmake
+++ b/cmake/SetupCodeChecks.cmake
@@ -57,6 +57,16 @@ if(UNCRUSTIFY_FOUND)
     add_dependencies(style uncrustify_style)
 endif()
 
+if(ASTYLE_FOUND)
+    # targets for verifying formatting
+    add_custom_target(astyle_check)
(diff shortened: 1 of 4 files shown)


cmake/SetupCodeChecks.cmake
```diff
@@ -226,19 +271,20 @@ endmacro(blt_add_cppcheck_target)
 ## MODIFY_FILES, if set to TRUE, modifies the files in place and adds the created target to
 ## the style target.  Otherwise the files are not modified and the created target is added
 ## to the check target.
+## Note: Setting MODIFY_FILES to FALSE is only supported in Uncrustify v0.61 or greater.
```
kennyweiss: Should we add a check for this?

cmake/SetupCodeChecks.cmake
```diff
@@ -294,3 +340,89 @@ macro(blt_add_uncrustify_target)
...
+##                        COMMENT           <Additional Comment for Target Invocation>
+##                        WORKING_DIRECTORY <Working Directory>
+##                        SRC_FILES         [FILE1 [FILE2 ...]] )
+##
+## Creates a new target with the given NAME for running astyle over the given SRC_FILES.
+##
+## MODIFY_FILES, if set to TRUE, modifies the files in place and adds the created target to
+## the style target. Otherwise the files are not modified and the created target is added
+## to the check target.  
+## Note: Setting MODIFY_FILES to FALSE is only supported in AStyle v2.05 or greater.
```
kennyweiss: Should we add a check for this?

cmake/SetupCodeChecks.cmake
```diff
@@ -294,3 +340,89 @@ macro(blt_add_uncrustify_target)
...
+endif()
+
+if(${arg_MODIFY_FILES})
+    set(MODIFY_FILES_FLAG --suffix=none)
+else()
+    set(MODIFY_FILES_FLAG --formatted --dry-run)
+endif()
+
+add_custom_target(${arg_NAME}
+        COMMAND ${ASTYLE_EXECUTABLE} ${arg_PREPEND_FLAGS}
```
kennyweiss: Unlike uncrustify, astyle does not return non-zero when there are files that need to be formatted, so, as long as the config file is valid, these targets passes.

When there are files that need to be formatted, ``astyle`` outputs a line, per file that needs to be formatted of the form:
```cmake
Formatted <filename>
```

Should we add some code to run astyle and grep through the output for these lines so that the behavior is similar to that of our ``uncrustify`` targets?

cmake/SetupCodeChecks.cmake
white238: Not sure how you would...

cmake/SetupCodeChecks.cmake
```diff
@@ -294,3 +340,89 @@ macro(blt_add_uncrustify_target)
...
+## WORKING_DIRECTORY is the directory in which astyle will be run. It defaults to 
+## the directory where this macro is called.
+##
+## SRC_FILES is a list of source files to style/check with astyle.
+##
+##------------------------------------------------------------------------------
+macro(blt_add_astyle_target)
+
+## parse the arguments to the macro
+set(options)
```
white238: Indent this macro

cmake/SetupCodeChecks.cmake
white238: That is unfortunate. I know users want it to fail if files need to be formatted.

cyrush: If it's not too hard, we should try to parse and error. 
We will have to weigh with how brittle we feel parsing will be

kennyweiss: I added a wrapper script to run AStyle and parse the results. Both astyle_style and astyle_check targets now fail when there are improperly formatted source files.

kennyweiss:
@white238 @cyrush -- I've addressed all the concerns that were raised.

I also removed the code check targets from default builds since they were getting triggered when building an entire solution.

jdahm:
On this topic... Could we add others to conform (mostly) with the [GNU standard targets](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html)? It would be nice to have TAGS, uninstall, etc.

kennyweiss:
Thanks @jdahm -- seems like an interesting suggestion.  There are definitely good ideas there, but it would be good to discuss what is compatible with cmake/blt, since the backend generator is not always ``make``.

Could you create a separate issue for this so we can discuss/track it?

Issue #175, opened 2018-08-18
The BLT provided "openmp" pseudotarget doesn't (always) pass the correct openMP fortran flags


gzagaris:
The "openmp" pseudotarget assumes that the OpenMP Fortran Flags are the same as C/C++ compiler flags.

We register the "openmp" pseudotarget as follows:
```
#################################################
# OpenMP
# (OpenMP support is provided by the compiler)
#################################################
message(STATUS "OpenMP Support is ${ENABLE_OPENMP}")
if(ENABLE_OPENMP)
    find_package(OpenMP REQUIRED)
    message(STATUS "OpenMP CXX Flags: ${OpenMP_CXX_FLAGS}")
    
    # register openmp with blt
    blt_register_library(NAME openmp
                         COMPILE_FLAGS ${OpenMP_CXX_FLAGS} 
                         LINK_FLAGS    ${OpenMP_CXX_FLAGS} )
endif()
```

This works for the most part, as long as:
* The C++ language is enabled
* The C++ and Fortran OpenMP flags match 

The latter may not always be the case. For example, for a project that uses Clang for C/C++ in conjunction with GNU Fortran, CMake will compute the following flags:

```
//C++ compiler flags for OpenMP parallization
OpenMP_CXX_FLAGS:STRING=-fopenmp=libomp

//C compiler flags for OpenMP parallization
OpenMP_C_FLAGS:STRING=-fopenmp=libomp

//Fortran compiler flags for OpenMP parallization
OpenMP_Fortran_FLAGS:STRING=-fopenmp
```

But, when building an executable like this:
```
blt_add_executable(
    NAME foo
    SOURCES foo.F
    DEPENDS_ON  openmp
    )
```

This would pass `OpenMP_CXX_FLAGS`, leading to the following compiler error:
```
gfortran: error: unrecognized command line option '-fopenmp=libomp'
```

Perhaps, BLT should register two pseudo libraries/targets for C/C++ and Fortran respectively, like this:
```
if(ENABLE_OPENMP)
    find_package(OpenMP REQUIRED)
    message(STATUS "OpenMP CXX Flags: ${OpenMP_CXX_FLAGS}")
    
    # register openmp with blt
    blt_register_library(NAME openmp-cxx
                         COMPILE_FLAGS ${OpenMP_CXX_FLAGS} 
                         LINK_FLAGS    ${OpenMP_CXX_FLAGS} )

   if ( ENABLE_FORTRAN)
[...]

Issue #163, opened 2018-06-19
create an example that tests both openmp and cuda


cyrush:
In `blt_add_lib`, we can't use 

`DEPENDS_ON openmp cuda`

But you can do the following:

`DEPENDS_ON cuda`

```
if(ENABLE_OPENMP)
  if(CUDA_FOUND)
    blt_add_target_compile_flags(TO my_target FLAGS "-Xcompiler ${OpenMP_CXX_FLAGS}")
  else()
    add_target_compile_flags(TO my_target FLAGS "${OpenMP_CXX_FLAGS}")
  endif()
endif()
```

gzagaris:
That seems like something that BLT should handle under the covers in the future. 

The presence of the `openmp` and `cuda` dependencies could be detected and take proper action for the corresponding target. Although the provided solution is not that complicated and can be easily hidden in a macro in application code, it would be nice if BLT natively supports this.

cyrush:
@gzagaris yes, I think its pretty easy to do this cleanly in BLT

white238:
This works for C++ and there is a new blt_cuda_openmp_smoke test that exercises this capability.  Fortran may not work yet though.

PR #190, opened 2018-09-20
Using generators so that OpenMP can be passed to nvcc compilers.

rblake-llnl:
With a small change to the openMP flags in BLT, I'm able to use openMP can cuda at the same.  This change will unfortunately break compatibility with Visual Studio builds, but I'm not sure that OpenMP+CUDA was working for that case anyway.

### DIFF
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -55,8 +55,11 @@ if(ENABLE_OPENMP)
     
     # register openmp with blt
     blt_register_library(NAME openmp
-                         COMPILE_FLAGS ${OpenMP_CXX_FLAGS} 
-                         LINK_FLAGS    ${OpenMP_CXX_FLAGS} )
+                         COMPILE_FLAGS
+                         $<$<NOT:$<COMPILE_LANGUAGE:CUDA>>:${OpenMP_CXX_FLAGS}> 
+                         $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=${OpenMP_CXX_FLAGS}> 
+                         LINK_FLAGS ${OpenMP_CXX_FLAGS}
+                         )
 endif()
 
 #####################################################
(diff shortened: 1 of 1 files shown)


gzagaris:
Very cool! I think this is an elegant solution addressing some of what was mentioned in #163 

I think, a similar strategy, employing CMake's generator expressions, can be used to resolve #175 as well.

davidbeckingsale:
Agree, I like this solution. I would let others speak to the impact on VS though...

rblake-llnl:
This should work for the Fortran case as well, though the logic might get more complicated.  

Someone has proposed a patch to cmake that removes the need for the if statement here:
https://gitlab.kitware.com/FloopCZ/cmake/commi/4f82199befee194aa924fd4dcdb2f4aad1f7dab6

rblake-llnl:
Incorrect link, fixed here: 

https://gitlab.kitware.com/FloopCZ/cmake/commit/4f82199befee194aa924fd4dcdb2f4aad1f7dab6

rblake-llnl:
Modified the pull request to support any combination of Fortran, CUDA, and openmp with a sane fallback compatible with Visual Studio builds.

rblake-llnl:
@kennyweiss @gzagaris, The checks for this PR keep failing due to the build servers falling down.  Can someone on the team restart the build process?

gzagaris:
Hmm... it doesn't look like I have access to restart appveyor or travis. @cyrush do you? or is there no good way to do this? Alternatively, @rblake-llnl , you could make an "empty" commit to trigger the builds and remove it later, but, I'd be surprised if there is no other way to manually trigger the checks.

cyrush:
@rblake-llnl @gzagaris , I restarted the appveyor build.

gzagaris:
Thanks @cyrush.

rblake-llnl:
Still no luck.  AppVeyor gives the following error message:

> AppVeyor failed to assign job to 'gce' build environment after 3 attempt(s). The envronment is either unavailable or there are no free workers, however alternate build environment was not configured by the system administrator. Please try again later.

cyrush:
@rblake-llnl, fighting with the ci is not fun @davidbeckingsale can we disable the required checks for this PR?

davidbeckingsale:
@cyrush - you should be able to merge this now. Required checks can only be modified globally, so we need to turn it back on after this is merged.

davidbeckingsale:
Required Appveyor checks re-enabled 👍

rblake-llnl:
Thanks @cyrush and @davidbeckingsale !

PR #193, opened 2018-10-12
added CUDA_HOST_COMPILER to the CMAKE_CUDA_FLAGS list

rrsettgast:
Builds on ray seemed to be missing ccbin from the compile line. This seems to fix the issue.


cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -74,12 +74,16 @@ message(STATUS "CUDA Libraries:    ${CUDA_LIBRARIES}")
...
 set (CUDA_PROPAGATE_HOST_FLAGS Off)
-if (CMAKE_CXX_COMPILER)
+if (( NOT CUDA_HOST_COMPILER ) AND CMAKE_CXX_COMPILER)
   set (CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
 else ()
   set (CUDA_HOST_COMPILER ${CMAKE_C_COMPILER})
 endif ()
 
+if (CUDA_HOST_COMPILER)
+      list(APPEND CMAKE_CUDA_FLAGS "-ccbin ${CUDA_HOST_COMPILER}")
```
davidbeckingsale: Does CMake do this automatically if you set `CMAKE_CUDA_HOST_COMPILER`? https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_HOST_COMPILER.html

rrsettgast: I tried setting '''CMAKE_CUDA_HOST_COMPILER''' in my hostconfig and it didn't work. Upon further inspection, it seems that it gets wiped out at some point. I'll try to find out where.

rrsettgast: @davidbeckingsale I didn't make ```CMAKE_CUDA_HOST_COMPILER``` a cache variable in my previous test. If it is set as a cache variable, it works correctly. So why are there ```CMAKE_CUDA_HOST_COMPILER``` and ```CUDA_HOST_COMPILER```?

gzagaris: @rrsettgast, the `CUDA_*` flags are used by the `FindCUDA` module, which is now superseded by native support for the CUDA language in CMake. My guess is that this code predates that, hence why `CUDA_HOST_COMPILER` is used here.

We should probably bump the required CMake version for BLT and use the `CMAKE_CUDA_*` flags consistently.

Glad to hear that setting the `CMAKE_CUDA_HOST_COMPILER` flag fixes your issue.

rrsettgast:
This works, or you can force the CMAKE_CUDA_HOST_COMPILER cache variable to be set in the hostconfig or command line. Merge or close this PR?

cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -45,8 +45,19 @@
 ################################
 set (CMAKE_MODULE_PATH "${BLT_ROOT_DIR}/cmake/thirdparty;${CMAKE_MODULE_PATH}")
 
+
+if ( NOT CMAKE_CUDA_HOST_COMPILER )
+    if ( CMAKE_CXX_COMPILER )
+        set ( CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER} CACHE STRING "" FORCE)
+    else ()
+        set ( CMAKE_CUDA_HOST_COMPILER ${CMAKE_C_COMPILER} CACHE BOOL "" FORCE)
```
gzagaris: For consistency this should also be a `CACHE STRING` variable.

cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -45,8 +45,19 @@
...
+    endif ()
```
gzagaris: It may be helpful to add `message()` here to spit out the value of `CMAKE_CUDA_HOST_COMPILER`

cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -45,8 +45,22 @@
 ################################
 set (CMAKE_MODULE_PATH "${BLT_ROOT_DIR}/cmake/thirdparty;${CMAKE_MODULE_PATH}")
 
+
+if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.9.0" )
```
white238: I think 3.8.0 might actually be fine.  Did you come up with this number from somewhere?

davidbeckingsale: I think  there is quite a big difference in "native" CUDA support between 3.8 and 3.9, @cyrush has recommended 3.9 in the past.

rrsettgast: I tried 3.9 and it worked. Tried 3.7 and it didn’t work. There wasn’t 3.8 on Ray. The cmake documentation for CMAKE_CUDA_HOST_COMPILER appears to start at 3.10 so I figured 3.9 would be the earliest that should be specified.

cmake/thirdparty/SetupCUDA.cmake
kennyweiss: Just wanted to add that we should only bump the required CMAKE version to 3.9 or 3.10 for CUDA features when ``ENABLE_CUDA`` is set.

When not using CUDA, I think BLT only currently requires something like 3.3 or 3.5.

gzagaris: It is going to be confusing to have different "supported" versions of CMake depending on what feature is enabled etc. I also don't recommend we sprinkle CMake version guards everywhere in BLT -- that is error prone and can get difficult to maintain in the long run.

I suggest the docs are updated to mention that the "recommended" version of CMake is 3.9+, perhaps with a warning (`AUTHOR_WARNING`)

PR #194, opened 2018-10-17
fixed regex statement for uncrustify version check

rrsettgast:
to allow for trailing letters that are used for non-release repo commits. For instance, the version my uncrustify returns is 0.67f...which will fail this test. In this change we will allow the trailing letter, but not use it for anything.


kennyweiss:
Thanks for the fixup @rrsettgast!

Issue #181, opened 2018-09-04
Upgrade flag-related blt macros to work with INTERFACE targets


kennyweiss:
PR #177 updated the ``blt_add_target_compile_flags`` macro to work with ``INTERFACE`` targets.

We can apply a similar upgrade to the ``blt_add_target_definitions`` and ``blt_add_target_link_flags`` macros for ``INTERFACE`` targets.

white238:
blt_add_target_definitions is already fixed.

Maybe I am not understanding something but blt_add_target_link_flags doesn't really make sense for interface targets (you don't link header only libraries). @kennyweiss ?

PR #197, opened 2018-10-30
Bugfix/gtest cxxabi

kennyweiss:
* Patches gtest to allow users to override GTEST_HAS_CXXABI_H_
* Improves blt_add_target_definitions macro (partially fulfilling #181), with unit test
* Adds host-config for clang@4 using clang's ``libc++`` instead of ``libstdc++``


kennyweiss:
This PR was based on a user request to allow configuring clang with the llvm ``libc++`` library instead of ``libstdc++``.  The patch to ``gtest`` is already in gtest's master branch.

PR #198, opened 2018-11-09
add status messages that display cmake version and cmake exe path

cyrush:


cyrush:
@white238  let me know if i should move to a better place

kennyweiss:
Good idea

Issue #205, opened 2018-12-04
Language-based generator expressions


kennyweiss:
@mclarsen came across a problem with the openmp generator expressions that were introduced in #190.

In his BLT-based code (ascent), the main library was built with Fortran support (``ENABLE_FORTRAN=ON``) and openmp support (``ENABLE_OPENMP=ON``).  He then has a ``C++`` example application that uses the installed version of this library.

The example application pulls in the cmake install file for the library, which has Fortran generator expressions in its link flags, leading to a (non-terminal, but annoying) cmake error:

```cmake
CMake Error at CMakeLists.txt:82 (target_link_libraries):
  Error evaluating generator expression:
    $<COMPILE_LANGUAGE:Fortran>
  $<COMPILE_LANGUAGE:...> Unknown language.
```

Here is a sanitized version of the offending lines in the generated cmake install file from the library:
```cmake
set_target_properties(ascent PROPERTIES
  INTERFACE_COMPILE_OPTIONS "<some-compile-options>;\$<\$<NOT:\$<COMPILE_LANGUAGE:Fortran>>:-fopenmp>;\$<\$<COMPILE_LANGUAGE:Fortran>:-fopenmp>"
  INTERFACE_INCLUDE_DIRECTORIES "<more-includes"
  INTERFACE_LINK_LIBRARIES "<link-flags"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "<system-includes>"
)
``` 

Note: It seems that this warning has been removed from newer versions of cmake (i.e. versions greater than 3.12 -- see https://gitlab.kitware.com/cmake/cmake/merge_requests/2031).
 
Thoughts: @rblake-llnl (or others)

rblake-llnl:
Cross referencing with #175.  In that commit, @gzagaris has a library that has different OpenMP flags when using CXX and Fortran.  

I'm not sure what the good answer here is.  On the one hand, it's less than ideal if generator expressions leak out of the project that uses them because they might use different languages and cause the annoying error seen above for older cmake versions.  

On the other hand, if CXX, CUDA, and Fortran do require different flags, shouldn't the calling code know that?

rblake-llnl:
Does #206 fix the problem?

kennyweiss:
@mclarsen -- please check out PR #206 for the cmake error that you reported, when you get a chance.

kennyweiss:
Thanks @rblake-llnl!

white238:
[...]

PR #206, opened 2018-12-04
Possibly fixing bugs in #175 and #205

rblake-llnl:
* Use correct flags when using Fortran and CUDA.
* If Fortran and CXX flags are the same, don't use a generator expression.


cmake/SetupCompilerOptions.cmake
```diff
@@ -121,12 +121,18 @@ message(STATUS "OpenMP Support is ${ENABLE_OPENMP}")
 if(ENABLE_OPENMP)
     find_package(OpenMP REQUIRED)
     message(STATUS "OpenMP CXX Flags: ${OpenMP_CXX_FLAGS}")
-    
+
+    if (ENABLE_FORTRAN AND NOT OpenMP_CXX_FLAGS STREQUAL OpenMP_Fortran_FLAGS)
```
kennyweiss: Perhaps it would be a good idea to add a line about why this is here/necessary?

rblake-llnl: Yep, added a comment.

white238:
Not sure why this sat for so long but I will update the branch and as long as it passes, I will merge.

cyrush:
@white238 thanks, I was just going to ask about getting this merged!

PR #223, opened 2019-02-14
Uncrustify changed heir --no-backup flag to not replace

white238:
Uncrustify decided to be unhelpful and changed "--no-backup" to no longer replace the file it is modifying.  So it does nothing by itself (In fact if you call just no-backup it only creates a backup and does not modify the file). You now need "--replace --no-backup".  I tested this on 0.63 and 0.68.

I want my 3 hours back.

### DIFF
--- a/cmake/SetupCodeChecks.cmake
+++ b/cmake/SetupCodeChecks.cmake
@@ -445,7 +445,7 @@ macro(blt_add_uncrustify_target)
     set(_generate_target TRUE)
 
     if(${arg_MODIFY_FILES})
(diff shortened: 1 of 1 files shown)


kennyweiss:
Does ``--replace`` still work with 0.61 ?

Uncrustify is not backwards compatible, so upgrades need to be deliberate.

white238:
Sorry I actually checked this on 0.61 and 0.68.

kennyweiss:
Thanks @white238.

PR #224, opened 2019-02-16
Support Object libraries

white238:
This also includes a release notes file (much needed) and a bump in version number so we actually start tracking it.  This was tested in a small test case and also a proof of concept for a unified Axom library.


tests/internal/src/object_library_test/base_object.cpp
```diff
@@ -0,0 +1,19 @@
...
+// This file is part of BLT.
+//
+// For additional details, please also read BLT/LICENSE.
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+#include "base_object.hpp"
+
+int base_number()
+{
+    return 2;
+}
```
davidbeckingsale: Missing a trailing newline 😄

davidbeckingsale: Note that this actually triggers a warning in PGI, and thus will not compile with `-Werror`

tests/internal/src/object_library_test/base_object.hpp
```diff
@@ -0,0 +1,19 @@
...
+// This file is part of BLT.
+//
+// For additional details, please also read BLT/LICENSE.
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+#ifndef BASE_OBJECT_HPP
+#define BASE_OBJECT_HPP
+
+int base_number();
+
+#endif
```
davidbeckingsale: Same comment as above.

tests/internal/src/object_library_test/object.cpp
```diff
@@ -0,0 +1,20 @@
...
+//
+// For additional details, please also read BLT/LICENSE.
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+#include "base_object.hpp"
+#include "object.hpp"
+
+int object_number()
+{
+    return 1 + base_number();
+}
```
davidbeckingsale: Trailing newline.

tests/internal/src/object_library_test/object.hpp
```diff
@@ -0,0 +1,19 @@
...
+// This file is part of BLT.
+//
+// For additional details, please also read BLT/LICENSE.
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+#ifndef OBJECT_HPP
+#define OBJECT_HPP
+
+int object_number();
+
+#endif
```
davidbeckingsale: Trailing newline.

RELEASE-NOTES.md
```diff
@@ -0,0 +1,35 @@
...
+### Fixed
+- Incorrect use of cuda vs cuda_runtime targets
+- Improved tutorial documentation
+- Incorrect use of Fortran flags with CUDA (Thanks Robert Blake)
+- Handle correctly CMake version differences with CUDA host compiler variables
+  (Thanks Randy Settgast)
+- Handle uncrustify (version 0.68) command line option changes (--no-backup)
+
+### Known Bugs
+
```
cyrush: you can omit the empty sections

white238: I didn't want to forget about them since it was the first one.  Though my code is always bug free.

RELEASE-NOTES.md
```diff
@@ -0,0 +1,35 @@
...
+- Object library support through blt_add_library(... OBJECT TRUE ...)
+- Now reporting BLT version through CMake cache variable BLT_VERSION
+- Output CMake version and executable used during CMake step
+- Clang-query support now added (Thanks David Poliakoff)
+
+### Removed
+
+### Deprecated
+
+### Changed
```
kennyweiss: Thanks for adding this @white238!

cmake/BLTMacros.cmake
```diff
@@ -370,24 +378,33 @@ endmacro(blt_register_library)
...
 ## Adds a library target, called <libname>, to be built from the given sources.
 ## This macro uses the BUILD_SHARED_LIBS, which is defaulted to OFF, to determine
 ## whether the library will be build as shared or static. The optional boolean
 ## SHARED argument can be used to override this choice.
 ##
+## The OBJECT argument creates a CMake object library. Basically it is a collection
+## of compiled source files that are not archived or linked. Unlike regular CMake
+## object libraries you do not have to use the $<TARGET_OBJECTS:<libname>> syntax,
+## you can just use <libname>.
+##    Note: Object libraries do not follow CMake's transitivity rules until 3.13.
```
kennyweiss: Forgot to mention -- can we add a check/warning for usage of object libraries if the cmake version is too low?  
This will give users an explanation for unexpected behavior, and make it more obvious that they should upgrade if they want to depend on this capability.

cyrush: @kennyweiss I think for some cases it will work just fine, but the transitive case won't be flushed out until folks use newer cmake. Not sure a warning is necessary b/c it's subtle. 

For example, I think I can use this now b/c in a few cases where transitive wont be a requirement.

white238: The more I think about this the more I think transitivity is not a problem.  This utilized the registered library logic and will add info from all the dependencies.  I don’t think you can create a ordering problem with object libraries because they aren’t linked.

davidbeckingsale:
Look like the CMake 3.4 tests are failing - is that expected? If so, can we make this feature conditional.

white238:
3.4 is supposed to support object libraries but it is throwing an error about the generator expression that is needed.  I am going to play around with it and see if I can fix it but this might be the point at which we just say BLT requires at least CMake 3.8...

davidbeckingsale:
Sounds good to me!

white238:
Attempting to get access to a windows machine so I can debug this.  It is only happening on one variant of the windows build.

white238:
This is now passing tests.  Can someone give it a once over then ill merge it.

cmake/BLTMacros.cmake
```diff
@@ -302,7 +303,7 @@ endmacro(blt_add_target_link_flags)
 ##------------------------------------------------------------------------------
 macro(blt_register_library)
 
-    set(singleValueArgs NAME TREAT_INCLUDES_AS_SYSTEM)
+    set(singleValueArgs NAME OBJECT TREAT_INCLUDES_AS_SYSTEM)
```
kennyweiss: Please document ``OBJECT``, as well as its default value in the signature for ``blt_register_library`` above (around line 270)

kennyweiss: Thanks for the updated docs -- it wasn't obvious (to me) that this should not be manually set by users (but it makes sense).

PR #229, opened 2019-02-22
Fix in-source error message

markcmiller86:
If you mistakenly run CMake in the ``src`` dir, you have to remove *both* ``CMakeLists.txt`` and ``CMakeFiles`` directory. This updates the error message to reflect that.


cyrush:
good catch -- thanks @markcmiller86 !

PR #230, opened 2019-02-27
Initial concept for setup/find third-party libraries

white238:
The idea is this would unify a lot of copy and pasted find tpl logic into BLT.  This would not only find the tpl but also call blt_register_library as well.

I tested these two in Axom but some discusion should be had before this is merged and the flood gates are opened.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -1209,3 +1209,68 @@ macro(blt_print_target_properties)
     unset(_is_cmake_target)
 endmacro(blt_print_target_properties)
 
+
+##------------------------------------------------------------------------------
+## blt_setup_library( NAME     <name>
+##                    PATH     <path to installed TPL>
+##                    VERSION  <version>
+##                    REQUIRED [TRUE (default) | FALSE ])
+##
+## This macro is used to find and setup the named third-party library using
+## BLT's setup modules located in:
+##       <BLT Root>/setup_tpl_modules/<lowercase NAME>.cmake.
+## 
+## PATH in which to search for named TPL. No system paths will be searched.
+##
+## VERSION is not always required but if differences in versions of a library then it will
+## be handled inside of the setup module.
+##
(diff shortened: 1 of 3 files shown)


cmake/BLTMacros.cmake
```diff
@@ -1209,3 +1209,68 @@ macro(blt_print_target_properties)
...
+    if(NOT DEFINED arg_PATH)
+        message(FATAL_ERROR "PATH is a required parameter for the blt_setup_library macro")
+    endif()
+
+    # Uppercase and lowercase the name to be used later
+    string(TOUPPER ${arg_NAME} _uppercase_name)
+    string(TOLOWER ${arg_NAME} _lowercase_name)
+
+    # Check that we have a setup tpl module
+    set(_setup_module_dir "${BLT_ROOT_DIR}/setup_tpl_modules")
```
kennyweiss: We should allow user codes to add a local ``tpl module`` path, which overrides blt's default module path.

setup_tpl_modules/hdf5.cmake
```diff
@@ -0,0 +1,125 @@
...
+        set(check_hdf5_inc_dir_ok 1)
+    endif()
+endforeach()
+
+if(NOT check_hdf5_inc_dir_ok)
+    message(FATAL_ERROR " ${HDF5_INCLUDE_DIRS} does not include HDF5_DIR")
+endif()
+
+#
+# filter HDF5_LIBRARIES to remove hdf5_hl if it exists
```
kennyweiss: Are we planning to support additional configurability?

E.g. for this section here, if users want to use the ``HDF5_HL`` libraries, would we support overriding these options, or would they have to write their own module?

cyrush:
Hmmm, I don't really like the idea of providing our own cmake find method tpl logic as part of BLT.

For example: Conduit already provides it's own, as part of it's install -- which can be used.

white238:
This has side effects (like repo bloat) and also has the problem that everyone's find's are slightly different. Maybe we could provide a blt-extras repo with recipes people could alter... closing for now.

PR #233, opened 2019-03-05
Forward gtest_disable_pthreads to gtest.

corbett5:
I needed this on blueos.

### DIFF
--- a/thirdparty_builtin/CMakeLists.txt
+++ b/thirdparty_builtin/CMakeLists.txt
@@ -26,6 +26,10 @@ if(ENABLE_TESTS)
         list(APPEND gtest_defines "-DGTEST_HAS_DEATH_TEST=0")
     endif()
 
+    if(gtest_disable_pthreads)
+      list(APPEND gtest_defines "-DGTEST_HAS_PTHREAD=0")
+    endif()
+
     #
     # Use BLT's ENABLE_GMOCK option to set the build options for 
     # gtest's cmake project


thirdparty_builtin/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ if(ENABLE_TESTS)
         list(APPEND gtest_defines "-DGTEST_HAS_DEATH_TEST=0")
     endif()
 
+    if(gtest_disable_pthreads)
+      list(APPEND gtest_defines "-DGTEST_HAS_PTHREAD=0")
+    endif()
```
kennyweiss: It seems like if ``gtest_disable_pthreads`` is defined, then the following should already add this compiler define:
* https://github.com/google/googletest/blob/master/googletest/cmake/internal_utils.cmake#L58-L65
* https://github.com/google/googletest/blob/master/googletest/cmake/internal_utils.cmake#L121-L127

Is something else in your configuration undoing this?  (e.g. multiple blt projects, one of which is not setting ``gtest_disable_pthreads`` correctly?

corbett5: Yup, I saw that. I'm not sure why that doesn't work. I know it's not an issue with multiple blt projects though.

corbett5:
@kennyweiss nvcc doesn't like the way gtest uses pthreads and pthreads calls fail at runtime. It only becomes an issue with some advanced gtest features like parameterized tests.

https://github.com/LLNL/blt/blob/a5560f91526829751b40f714680cd4422d019224/host-configs/llnl-ray-blue_os-clang-coral%402018.05.23.cmake#L57

PR #234, opened 2019-03-05
Remove quoting from blt_add_target_link_flags

davidbeckingsale:
This was causing problems with OpenMP when OpenMP_CXX_FLAGS was a list.


cyrush:
> Pesky quotes... 👍

PR #236, opened 2019-03-06
Let blt_add_target_link_flags accept a list

davidbeckingsale:
This is an extension of the issues we observed in #234.

Without this PR, `blt_add_target_link_flags` and `blt_add_target_compile_flags` accept their `FLAGS` argument in two different forms. This makes both macros take a list.

Internally, CMake still requires the `LINK_OPTIONS` property be a string, so we parse the flags and remove ";" before adding them.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -200,6 +200,8 @@ endmacro(blt_set_target_folder)
 ## blt_add_target_link_flags (TO <target> FLAGS [FOO [BAR ...]])
 ##
 ## Adds linker flags to a target by appending to the target's existing flags.
+##
+## The flags argument contains a list of linker flags to add to the target.
 ##------------------------------------------------------------------------------
 macro(blt_add_target_link_flags)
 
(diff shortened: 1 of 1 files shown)


cmake/BLTMacros.cmake
```diff
@@ -219,8 +221,10 @@ macro(blt_add_target_link_flags)
         endif()
         # append new flag
         set(_LINK_FLAGS "${arg_FLAGS} ${_LINK_FLAGS}")
+        # LINK_FLAGS property _must_ be a string
```
kennyweiss: Are there ever cases where link flags could legitimately contain a semicolon?

kennyweiss: It might be more helpful if this comment said that it was converting ``_LINK_FLAGS`` from a (cmake) list to a string

davidbeckingsale:
```suggestion
        # Convert _LINK_FLAGS from a CMake ;-list to a string
```

davidbeckingsale: I don't know, but that does concern me.

white238: Wouldn't the command line parser think it was two commands then... IE "rm -rf *; cmake .."

davidbeckingsale: I think you mean `rm -rf *cmake*` 😄

davidbeckingsale: You could pass a quoted argument with a semicolon though? Like `"blah;blah;blah"` - this regex would trash that.

kennyweiss: @davidbeckingsale -- I was going to say the same thing 
(e.g. https://stackoverflow.com/a/20233998), 
but that does help with my concern.

Is it worth mentioning the semicolon stripping in the documentation, so that if it does come up, users can have some idea about what caused it ?

cyrush: I don't think I have ever seen a case where semicolon's would be used.  
b/c semicolon would terminate the current command and truncate it as @white238  states.

davidbeckingsale: Okay, I updated the documentation like @kennyweiss suggested.

davidbeckingsale:
Is everyone okay if this gets merged? @kennyweiss @cyrush @white238

Issue #235, opened 2019-03-06
Object library not respecting DEPENDS_ON


mclarsen:
I am trying to use an object library that depends on a library that is registered with BLT. It previously worked and all I changed was adding `OBJECT TRUE` to the `blt_add_library` command, but it does not pull in the dependencies anymore. Is this supported?

mclarsen:
This is sort of resolved.

PR #237, opened 2019-03-06
Fix inherited includes for object libraries

### LINKED ISSUE #232: Object library build failing.
When trying to build an object library using the "OBJECT True" parameter in blt_add_library, the build of the library fails.  The library depends on 4 other libraries.  Two of them are external libraries (hdf5 and Boost) and are registered with blt_register_library.  The other two are submodules of the project and are built as part of the project using blt_add_library.  If the library is built as a standard, static library (OBJECT False), it builds without problem.

Looking into the CMake coding in BLTMacros.cmake and BLTPrivateMacros.cmake, I found that for the submodule libraries, BLT_<lib>_INCLUDES and BLT_<lib>_DEFINES are not set.  Rather the values passed into blt_add_library for include paths and defines are attached to those libraries' targets using target_include_directories and target_compile_definitions in blt_add_library.  The registered libraries have the BLT variables set in blt_register_library.  In blt_setup_target, it uses the values of the BLT variables, if set, to call target_include_directories and target_compile_definitions for that target.  So for my library, in either build, they are not called.

Since object libraries cannot be linked,  blt_set_target omits the step of calling target_link_libraries for the object library since CMake will not allow it.  However,  it is this step which causes the object library build to fail.  With the call to target_link_libraries (static library build), CMake gets the include paths and defines from the target information of the linked library and the build succeeds.  Without the call (object library call), the build fails.

white238:
Should fix #232 and #235


cmake/BLTPrivateMacros.cmake
```diff
@@ -97,6 +97,12 @@ macro(blt_setup_target)
         message( FATAL_ERROR "Must provide a NAME argument to the 'blt_setup_target' macro" )
     endif()
 
+    # CMake fix object transitivity
```
kennyweiss: Could you please expand this comment a bit?
E.g.
> Workaround for bug in CMake's object library dependency transitivity.  Fixed in CMake 3.12+.

cyrush:
@white238  Looks like pthread stuff is biting us on windows? Any idea why thats popping up now?

@mclarsen also got bit by pthread linking, I think post @corbett5's recent changes.  Matt can you confirm what version you where using so we can unravel the mystery?

Could it be that recent changes fixed in one place and broke in others?

mclarsen:
`[  9%] Linking CUDA executable ../../../tests/blt_cuda_runtime_smoke
nvcc fatal   : Unknown option 'pthread'`

commit 800ef02c5d173db3725c62f7e68d312afc52e7b7
Author: Benjamin Curtice Corbett <corbett5@llnl.gov>
Date:   Mon Mar 4 12:12:44 2019 -0800

    Forward gtest_disable_pthreads to gtest.

corbett5:
@mclarsen do you have `set(gtest_disable_pthreads ON CACHE BOOL "")` in your host config?

cyrush:
@corbett5 that might be the key. 

Q: was the pthreads fix needed for you in the case when you were linking with NVCC?

mclarsen:
The pthreads issue is orthogonal to the object library and no I did not have that set. Is there a reason pthreads shouldn't always be off for gtest?

corbett5:
@cyrush the axom blueos nvcc host config had `set(gtest_disable_pthreads ON CACHE BOOL "")`. This was enough to build everything and run tests which didn't depend on pthreads. However tests which used gtest features that used pthreads (`TEST_P` for example) would fail at runtime.  800ef02 fixed that.

@mclarsen I would try adding that flag and see what happens.

kennyweiss:
@mclarsen -- I am pretty sure the gtest death tests use threads. They are disabled by default though and have to be explicitly enabled. I am not sure what other gtest features use threads. 

I am not sure why you are seeing this issue now, but I don't think that @corbett5's change should be affecting you or object libraries, since it only adds a compiler define when the ``gtest_disable_pthreads`` option is added.  (Said differently, his change will not affect you if you have not set this option).

The better question is why @corbett5's change was needed when things previously worked on ``blue_os`` with ``nvcc`` (although @corbett5, please correct me if I'm wrong).  Perhaps there was a change to the system that required this?

Links:
* Death tests: https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#death-tests
* BLT support for death tests: https://github.com/LLNL/blt/blob/master/thirdparty_builtin/CMakeLists.txt#L22-L31
* @corbett5's PR: https://github.com/LLNL/blt/pull/233/files

corbett5:
@kennyweiss we didn't have a `blue_os` `nvcc` build in axom, I encountered the issue that led to #233 while creating that build. And the issue I was having wasn't with death tests, it was with the parameterized tests `TEST_P` which uses at least a pthreads mutex.

white238:
I tested this on the now more complicated test on both 3.9.2 (old target libraries) and 3.13.4 (new ones)

mclarsen:
I just tried this and I am now getting a CMake error:
`Object library target "vtkm_compiled_filters" may not link to anything.`

Doing some googling,  `Object libraries cannot be linked` from the CMake 3.9 documentation. I think the ability to link object files was added in 3.12.

mclarsen:
Specifically i am adding an object library:
`blt_add_library(
  NAME vtkm_compiled_filters
  SOURCES ${vtkm_filters_sources}
  HEADERS ${vtkm_filters_headers}
  DEPENDS_ON ${vtkm_filters_deps}
  OBJECT TRUE
  )`

where `vtkm_filter_deps` has an external library registered with BLT. It hits the `target_link_libraries` command on line 185 of `BLTPrivateMacros.cmake`  with the error:

  Object library target "vtkm_compiled_filters" may not link to anything.

cyrush:
@mclarsen given the vtkm deps are bonafide imported cmake targets, for now can you check if you see if this works if you pass their proper names to `blt_add_library`, instead of using the name you created in `blt_register_library`?  I think that may work and it would also point out what we need to test next to get BLT full working.

white238:
This is definitely the case where I am surprised this passed my tests... try adding “arg_OBJECT OR “ to the beginning of the if check on line 173 of BLTPrivateMacros.cmake.

mclarsen:
If i use the `vtkm_cont` target directly, it also fails. If I remove the vtkm dependency all together, it fails trying to link `cuda` in, all in with the same error message.

mclarsen:
@white238 Adding that allowed cmake to complete the configure, but cmake still seems not to propagate the dependencies. On the compile line, the headers are not being included, and with the cuda dependency, the object is being compiled with gcc instead of nvcc.

If you have any time this week, I would be happy to sit down with you.

cyrush:
FYI: We worked through our issues and tested with ascent - i think this is ready to merge.

PR #241, opened 2019-03-18
Fix clang-cuda support

davidbeckingsale:
This PR guards the `enable_language(CUDA)` statement when CLANG_CUDA is enabled. This is required because nvcc cannot use newer versions of clang, so CMake's built in checks fail. 

It also updates the flags registered for `cuda` to be a list rather than a string.

### DIFF
--- a/cmake/thirdparty/SetupCUDA.cmake
+++ b/cmake/thirdparty/SetupCUDA.cmake
@@ -56,7 +56,10 @@ endif()
 ############################################################
 # Basics
 ############################################################
-enable_language(CUDA)
+# language check fails when using clang-cuda
+if (NOT ENABLE_CLANG_CUDA)
+  enable_language(CUDA)
+endif ()
 
 
 ############################################################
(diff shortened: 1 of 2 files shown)


cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -104,7 +107,7 @@ endif()
 
 set(_cuda_compile_flags " ")
 if (ENABLE_CLANG_CUDA)
-    set (_cuda_compile_flags "-x cuda --cuda-gpu-arch=${BLT_CLANG_CUDA_ARCH} --cuda-path=${CUDA_TOOLKIT_ROOT_DIR}")
+    set (_cuda_compile_flags "-x;cuda;--cuda-gpu-arch=${BLT_CLANG_CUDA_ARCH};--cuda-path=${CUDA_TOOLKIT_ROOT_DIR}")
```
kennyweiss: I think you should be able to accomplish the same by removing the outer quotation marks.
I.e.
```cmake
     set (_cuda_compile_flags -x cuda --cuda-gpu-arch=${BLT_CLANG_CUDA_ARCH} --cuda-path=${CUDA_TOOLKIT_ROOT_DIR}) 
```

(either way should be fine though)

davidbeckingsale: Good suggestion, I switched it.

PR #242, opened 2019-03-20
Switch order of input file checks for Sphinx

jdahm:
This allows us to support using both a local and ReadTheDocs build of Sphinx.

Read the docs runs both Doxygen and Sphinx in-place, while Umpire and CMake use separate build and source directories. Without this, Sphinx fails with the message:

```Exhale: the specified folder [.../Umpire/docs/doxygen/xml] does not exist.  Has Doxygen been run?```

However, the Doxygen files are (correctly) located in `${CMAKE_BINARY_DIR}/docs/doxygen/xml`. If we use a CMake-d `config.py` (`.in`) specifying the correct directory, we can work around the quirk in readthedocs when manually building them.

### DIFF
--- a/cmake/SetupDocs.cmake
+++ b/cmake/SetupDocs.cmake
@@ -77,11 +77,10 @@ macro(blt_add_sphinx_target sphinx_target_name )
     # HTML output directory
     set(SPHINX_HTML_DIR "${CMAKE_CURRENT_BINARY_DIR}/html")
 
-    # support both direct use of a conf.py file and a cmake config-d
-    # sphinx input file (conf.py.in). The cmake config-d input file is
+    # support both direct use of a conf.py file and a cmake-configured
+    # sphinx input file (conf.py.in). The cmake-configured input file is
     # preferred when both exist.
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/conf.py.in")
-
         configure_file("${CMAKE_CURRENT_SOURCE_DIR}/conf.py.in"
                        "${SPHINX_BUILD_DIR}/conf.py"
                        @ONLY)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmake/SetupDocs.cmake [hunk 1]
```diff
@@ -77,29 +77,30 @@ macro(blt_add_sphinx_target sphinx_target_name )
     # HTML output directory
     set(SPHINX_HTML_DIR "${CMAKE_CURRENT_BINARY_DIR}/html")
 
-    # support both direct use of a conf.py file and a cmake config-d 
-    # sphinx input file (conf.py.in)
-    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/conf.py")
+    # support both direct use of a conf.py file and a cmake config-d
```
kennyweiss: > config-d

Is this an abbreviation for ``configured``? Or does it mean something else?

If it is the former, I think it would be clearer to change it to ``configured``.

In  this context, we are discussing files with names ``config.py`` and ``config.py.in`` and I struggled a bit to parse the comment.

davidbeckingsale: This was an existing comment 😄 I agree that "configured" is better.

cmake/SetupDocs.cmake
```diff
@@ -77,11 +77,10 @@ macro(blt_add_sphinx_target sphinx_target_name )
     # HTML output directory
     set(SPHINX_HTML_DIR "${CMAKE_CURRENT_BINARY_DIR}/html")
 
-    # support both direct use of a conf.py file and a cmake config-d
-    # sphinx input file (conf.py.in). The cmake config-d input file is
+    # support both direct use of a conf.py file and a cmake-configured
```
kennyweiss: Thanks

jdahm: No problem. This is certainly cleaner 👍

PR #243, opened 2019-03-22
Don't define BLT_CXX_STD by default

davidbeckingsale:
This allows projects that detect and add C++ standard flags with their own logic (e.g. RAJA) to avoid having two -std flags added.


cmake/SetupCompilerOptions.cmake
```diff
@@ -282,6 +282,8 @@ endif()
 SET( CMAKE_CXX_EXTENSIONS OFF )
 SET( CMAKE_CXX_STANDARD_REQUIRED ON )
 
+if (BLT_CXX_STD)
+
 if( BLT_CXX_STD STREQUAL c++98 )
```
white238: Is this indented?

davidbeckingsale: No, but it should be 😄

white238: Approved after you fix your heinous crime against code styling... thanks for fixing this!

cmake/BLTOptions.cmake
```diff
@@ -75,12 +75,13 @@ endif()
...
 option(ENABLE_WARNINGS_AS_ERRORS "Enables treating compiler warnings as errors on all build targets" OFF)
 
-set(BLT_CXX_STD "c++11" CACHE STRING "Version of C++ standard")
+set(BLT_CXX_STD "" CACHE STRING "Version of C++ standard")
 set_property(CACHE BLT_CXX_STD PROPERTY STRINGS c++98 c++11 c++14)
 
-if( NOT ( ( BLT_CXX_STD STREQUAL "c++98" ) 
-       OR ( BLT_CXX_STD STREQUAL "c++11" ) 
-       OR ( BLT_CXX_STD STREQUAL "c++14" ) ) )
+if( BLT_CXX_STD AND (
```
white238: Should we be adding c++17?

cmake/SetupCompilerOptions.cmake
davidbeckingsale: No problems. I should use tabs right? 😂

white238: Please mix tabs and spaces...

cmake/BLTOptions.cmake
davidbeckingsale: We could do at a later date. One thing we have in RAJA (and why we don't need BLTs support at the moment) is flag auto-detection for compilers CMake doesn't know about. It might be useful to move that logic into BLT.

davidbeckingsale:
I've added a line to the RELEASE-NOTES - can you confirm the format/placement is correct? @kennyweiss @white238

RELEASE-NOTES.md
```diff
@@ -7,6 +7,13 @@ The format of this file is based on [Keep a Changelog](http://keepachangelog.com
 
 The project release numbers follow [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- BLT_CXX_STD is no longer defined to "c++11" by default. If undefined, BLT will
```
white238: Looks great thanks

PR #244, opened 2019-03-23
Don't require exported/installed object libraries

white238:
Make object libraries private but inherit info.  This was causing object libraries to be have to installed which makes no sense whatsoever.
Add include macro that filters duplicates


cmake/BLTPrivateMacros.cmake
```diff
@@ -77,6 +77,57 @@ macro(blt_find_executable)
 endmacro(blt_find_executable)
 
 
+##------------------------------------------------------------------------------
+## blt_inherit_target_info( TO       <target>
+##                          FROM     <target>)
+##
+##  The purpose of this macro is if you want to grab all the inheritable info
```
kennyweiss: Could you please list the "inheritable info" (or provide some examples)?

cmake/BLTPrivateMacros.cmake
```diff
@@ -77,6 +77,57 @@ macro(blt_find_executable)
...
+                                INCLUDES ${_interface_system_includes}
+                                SYSTEM   TRUE)
+    endif()
+
+    get_target_property(_interface_includes
+                        ${arg_FROM} INTERFACE_INCLUDE_DIRECTORIES)
+    if ( _interface_includes )
+        blt_add_target_includes(TO       ${arg_TO}
+                                INCLUDES ${_interface_includes})
+    endif()
```
kennyweiss: Might as well make the `SYSTEM` flag explicit (i.e. add ``SYSTEM FALSE``)

cmake/BLTPrivateMacros.cmake
```diff
@@ -77,6 +77,57 @@ macro(blt_find_executable)
...
+
+    get_target_property(_interface_defines
+                        ${arg_FROM} INTERFACE_COMPILE_DEFINITIONS)
+    if ( _interface_defines )
+        target_compile_definitions( ${arg_TO} PUBLIC ${_interface_defines})
+    endif()
+
```
kennyweiss: Would we ever need link flags or link directories (or anything else)?

cmake/BLTPrivateMacros.cmake
```diff
@@ -129,52 +180,25 @@ macro(blt_setup_target)
...
-
-                get_target_property(_interface_defines
-                                    ${dependency} INTERFACE_COMPILE_DEFINITIONS)
-                if ( _interface_defines )
-                    target_compile_definitions( ${arg_NAME} PUBLIC ${_interface_defines})
-                endif()
-            endif()
+            blt_add_target_includes(TO       ${arg_NAME}
+                                    INCLUDES ${BLT_${uppercase_dependency}_INCLUDES}
+                                    SYSTEM   ${BLT_${uppercase_dependency}_TREAT_INCLUDES_AS_SYSTEM})
```
kennyweiss: nice

cmake/BLTPrivateMacros.cmake
cyrush: First,  this is an awesome new macro. 

Second, i do think we may want linking info in the future - don't think we need it yet for this PR, so we can we create a ticket as a reminder?

cmake/BLTPrivateMacros.cmake
```diff
@@ -129,52 +180,25 @@ macro(blt_setup_target)
...
-                target_include_directories( ${arg_NAME} SYSTEM PUBLIC
-                    ${BLT_${uppercase_dependency}_INCLUDES} )
-            else()
-                target_include_directories( ${arg_NAME} PUBLIC
-                    ${BLT_${uppercase_dependency}_INCLUDES} )
-            endif()
-        endif()
-
-        # Handle inherited information for object libraries in CMake versions
-        # less than 3.12, because target_link_libraries can't be called
```
cyrush: Whoa -- do we still need any of the pre-3.12 logic?

cmake/BLTMacros.cmake
```diff
@@ -153,6 +153,63 @@ macro(blt_add_target_compile_flags)
...
 
 
+##------------------------------------------------------------------------------
+## blt_add_target_includes(TO       <target>
+##                         INCLUDES [PATH1 [PATH2 ...]]
+##                         SYSTEM   <TRUE|FALSE (default)>)
+##
+## Adds includes to the given target.
+##
+## Adds pre-processor definitions to a particular. This macro provides very
```
kennyweiss: > Adds pre-processor definitions to a particular.

Please fix this line.

cmake/BLTPrivateMacros.cmake
white238: It all went into the new inherit macro.  Now we just guard the target_link_library call against previous versions.  We might not even need the call in new versions but it does protect us against any inheritance we miss.

cmake/BLTMacros.cmake
white238: done.

cmake/BLTPrivateMacros.cmake
white238: Done.

white238:
Well that took a lot longer than I thought.  The include macro was somehow causing system includes to be lost in one case.  I spent two days trying to figure out what i was missing and gave up.  This is probably better anyways.  The original idea was the not add any includes that were duplicates but it's probably more efficient to just add them all then remove duplicates once per target at the end of the blt_add_(library|executable) macros.

jdahm:
In Umpire we are building a set of object libraries, that are then linked together to a single library and installed. When we `install(TARGETS umpire EXPORT umpire-targets ...)`, where umpire-targets contains other things to be installed, CMake complains that
```
CMake Error: install(EXPORT "umpire-targets" ...) includes target "umpire" which requires target "umpire_resource" that is not in the export set.
```
for each dependency that is not specifically added to umpire-targets. Note that these are object libraries, e.g.
```
blt_add_library(
  NAME umpire_resource
  ...
  OBJECT TRUE)
```
I've updated the BLT version in my local repo to `c012f37` (latest commit) and re-configured from scratch, but I'm still running into it. Should that error have been addressed by this PR?

davidbeckingsale:
@white238 - everything works with commit 74b208564917c78e9ddfc2311e5e73b01e95a981, but something must have changed in between that and current master that's triggering this error.

PR #245, opened 2019-03-28
Remove target_link_libraries from object libraries completely

white238:
Even PRIVATE causes new versions to require exported targets to install object libraries.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -29,7 +29,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 ### Deprecated
 
 ### Changed
-- Object libraries are no longer call target_link_libraries() but will pass inherited information
+- Object libraries no longer call target_link_libraries() but will pass inherited information
   because why would anyone ever want to install/export a bunch of object files.
 - Remove duplicates of select target properties at the end of blt_add_library and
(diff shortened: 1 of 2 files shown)


RELEASE-NOTES.md
```diff
@@ -29,7 +29,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 ### Deprecated
 
 ### Changed
-- Object libraries are no longer marked PUBLIC but will pass inherited information
+- Object libraries are no longer call target_link_libraries() but will pass inherited information
```
davidbeckingsale:
```suggestion
- Object libraries no longer call target_link_libraries() but will pass inherited information
```

white238: Fixed thanks.

cyrush:
what version of cmake?

We noticed that 3.9.2 doesn't complain -- newer seem simply complain (not error)

davidbeckingsale:
3.13

cyrush:
@davidbeckingsale thanks, I guess I would ask how well you think you have vetted 3.13.  We are sticking to 3.9.x b/c thats the only one we have tested across a wide range of platforms.

white238:
I verified this didn't cause any export errors in Axom in 3.13.4 and the exported targets look correct.

PR #251, opened 2019-04-16
OpenMP override link/compiler flags

white238:
* Move OpenMP logic with all other third-party stuff
* Clean up various logic
* Allow overriding of compile and link flags for OpenMP since this cant be fixed with a generator expression
* Add smoke test for OpenMP + CUDA


tests/smoke/CMakeLists.txt
```diff
@@ -118,7 +118,16 @@ if (ENABLE_CUDA)
     blt_add_test(NAME blt_cuda_runtime_smoke
                  COMMAND blt_cuda_runtime_smoke)
 
-
+    if (ENABLE_OPENMP)
```
kennyweiss: Would it make sense to test ``cuda_runtime`` + ``openmp`` ?

README.md
```diff
@@ -1,8 +1,8 @@
-BLT v0.2
-========
+# <img src="https://cdn.rawgit.com/LLNL/blt/develop/share/blt/logo/blt_logo.png" width="128" valign="middle" alt="BLT"/> BLT v0.2
```
white238: I think this logo will show up once I push to master. I hope.

tests/smoke/CMakeLists.txt
white238: Honestly kind of torn.. On one hand I don't want a million smoke tests but then it might also fail.

Issue #252, opened 2019-04-16
Target property LINK_FLAGS changed to LINK_OPTIONS in CMake 3.13+


white238:
This needs to be supported in blt_add_target_link_flags.  LINK_OPTIONS now supports generator expressions but also it needs to be a ; delimited list not a string.

Issue #239, opened 2019-03-13
Should use target property?


markcmiller86:
https://github.com/LLNL/blt/blob/07d3e3cc8613e05af0a6a75ac0d0a70cb91ebc30/docs/tutorial/calc_pi/CMakeLists.txt#L22

I am passing a long a comment to BLT code I copied into my project from a CMake expert @bartlettroscoe and ontributor to Sandia's TRIBITS CMake infrastructure.

The above referenced line should use target property instead

cyrush:
the blt macros now handle this. 

I am also surprised that example still needs that flag. We may be able to just take it out.

kennyweiss:
Thanks @markcmiller86  and @bartlettroscoe

We also do the same thing here:
https://github.com/LLNL/blt/blob/07d3e3cc8613e05af0a6a75ac0d0a70cb91ebc30/tests/internal/CMakeLists.txt#L24-L26

I agree that target properties would be better and we should make the switch. 
As @cyrush mentioned, we should also check that they are still needed.

Issue #218, opened 2019-01-28
MPIEXEC is deprecated in cmake 3.10


rrsettgast:
It appears that ```MPIEXEC``` is deprecated in favor of ```MPIEXEC_EXECUTABLE``` in ```cmake 3.10``` as seen [here](https://cmake.org/cmake/help/v3.10/module/FindMPI.html) . I was having some issues with my builds that may have been related to this, but I did notice that BLT uses ```MPIEXEC``` inside SetupMPI():

https://github.com/LLNL/blt/blob/09173186b693f86e9b5b50ce5f1e61a9c9172ccf/cmake/thirdparty/SetupMPI.cmake#L61

and in blt_add_test()

https://github.com/LLNL/blt/blob/09173186b693f86e9b5b50ce5f1e61a9c9172ccf/cmake/BLTMacros.cmake#L706

Is this something that should be updated?

white238:
I'll pop a branch soon that toggles this based on version. Thanks for the reported issue @rrsettgast !

Issue #183, opened 2018-09-04
blt_add_library INCLUDES option doesn't support header only


rrsettgast:

kennyweiss:
Thanks @rrsettgast -- this is related to #181 and can be resolved in a similar manner as #177.

PR #254, opened 2019-04-17
Various changes

white238:
* Use readthedocs link
* Handle MPIEXEC->MPIEXEC_EXECUTABLE variable change #218 
* Handle LINK_FLAGS->LINK_OPTIONS target property change #252 
* Quiet compile warning
* Remove global defines #239 
* Make blt_add_library INCLUDES support header only libraries #183

### DIFF
--- a/SetupBLT.cmake
+++ b/SetupBLT.cmake
@@ -51,11 +51,19 @@ if (NOT BLT_LOADED)
     include(${BLT_ROOT_DIR}/cmake/BLTOptions.cmake)
 
     ################################
-    # CMP0057 New is required by blt_setup_target()
-    # in order to support the IN_LIST if() operator
+    # CMake Policies
     ################################
+    # Support IN_LIST operator for if()
     if(POLICY CMP0057)
         cmake_policy(SET CMP0057 NEW)
+    endif()
+
+    # Turns relative target_sources() paths to absolute
+    # Policy added in 3.13+
+    # NOTE: this will be deprecated eventually but NEW causes
+    #  problems in header only libraries
+    if(POLICY CMP0076)
+        cmake_policy(SET CMP0076 OLD)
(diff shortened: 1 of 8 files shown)


SetupBLT.cmake
```diff
@@ -51,13 +51,21 @@ if (NOT BLT_LOADED)
     include(${BLT_ROOT_DIR}/cmake/BLTOptions.cmake)
 
     ################################
-    # CMP0057 New is required by blt_setup_target()
-    # in order to support the IN_LIST if() operator
+    # CMake Policies
     ################################
+    # Support IN_LIST operator for if()
```
davidbeckingsale: What version was this added? Might be worth noting, like you have in the comment below.

cmake/BLTMacros.cmake
```diff
@@ -217,17 +217,26 @@ macro(blt_add_target_link_flags)
...
-        get_target_property(_LINK_FLAGS ${arg_TO} LINK_FLAGS)
-        if(NOT _LINK_FLAGS)
-            set(_LINK_FLAGS "")
+        if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.13.0" )
+            # In CMake 3.13+, LINK_FLAGS was converted to LINK_OPTIONS.
+            # This now supports generator expressions but expects a list
+            # not a string
+
+            # Convert from a string to a CMake ;-list
+            string (REPLACE " " ";" _link_flags_list "${arg_FLAGS}")
```
davidbeckingsale: This breaks any flags with spaces?

white238: Yes this was definitely going to cause problems.  Good catch.  I removed it because this macro required a list anyways and fixed the one place in BLT we were violating that (MPI).

PR #258, opened 2019-04-25
Add support for C++17

white238:
This adds support for C++17 which works in most cases but XL and CMake's CUDA support haven't added this yet.

This is CMake's problem:
https://cmake.org/cmake/help/v3.14/prop_tgt/CUDA_STANDARD.html

There might be a work around we can fix up in BLT but I need to spend more time on it.


cmake/SetupCompilerOptions.cmake
```diff
@@ -226,38 +226,53 @@ if(BUILD_SHARED_LIBS)
...
 ################################
-# Enable C++11/14
+# C++ Standard
 ################################
 
 SET( CMAKE_CXX_EXTENSIONS OFF )
 SET( CMAKE_CXX_STANDARD_REQUIRED ON )
 
+set(BLT_CXX_STD "" CACHE STRING "Version of C++ standard")
+set_property(CACHE BLT_CXX_STD PROPERTY STRINGS c++98 c++11 c++17 c++17)
```
kennyweiss: `c++17` is here twice.  

Looks like the first one should be ``c++14`` ?

white238: Fixed thanks.

white238:
I added an issue #259 to remind me to go back and see if we can work around the CUDA issue.  This doesn't have to do with nvcc, its clearly a CMake deficiency.  If you set CUDA_STANDARD to 17 it throws a lot of target property warnings and fails to configure.  I am pretty sure we can work around it though.

The person that asked for this support doesn't care about CUDA (yet) so I figured we should get it in.

Issue #262, opened 2019-05-01
Docs examples don't include the correct cmake lines


mcneish1:
On https://llnl-blt.readthedocs.io/en/latest/creating_execs_and_libs.html, all the code boxes but the last are various shades of wrong.

The first probably should point to lines 21-25, the second should fully bracket `<BLT repository>`, and the third probably should point to lines 27-32.

https://llnl-blt.readthedocs.io/en/latest/unit_testing.html and https://llnl-blt.readthedocs.io/en/latest/external_dependencies.html look like they have the same issues, but the rest of the docs look fine.

white238:
Thank you. I'll take a look at this soon and update it to the better/more robust start-after/end-before sphinx feature that doesn't cause so many problems like line numbers does.

PR #267, opened 2019-05-09
Add ability to override MPI variables

white238:
Provides a way to override default MPI behavior in a clean way.
Also converts any code snippets in the tutorial from line numbers (always breaking) to tags. #262

### DIFF
--- a/cmake/thirdparty/SetupMPI.cmake
+++ b/cmake/thirdparty/SetupMPI.cmake
@@ -18,10 +18,12 @@ else()
     endif()
 endif()
 
-set(_mpi_compile_flags)
-set(_mpi_includes)
-set(_mpi_libraries)
-set(_mpi_link_flags)
+set(_mpi_compile_flags )
+set(_mpi_includes )
+set(_mpi_libraries )
+set(_mpi_link_flags )
+
+message(STATUS "Enable FindMPI:  ${ENABLE_FIND_MPI}")
 
 if (ENABLE_FIND_MPI)
     find_package(MPI REQUIRED)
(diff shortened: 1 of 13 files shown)


cmake/thirdparty/SetupMPI.cmake
```diff
@@ -67,8 +69,24 @@ if (ENABLE_FIND_MPI)
     list(REMOVE_DUPLICATES _mpi_libraries)
 endif()
 
+# Allow users to override CMake's FindMPI
```
kennyweiss: Do we always want to override, as opposed to appending?

docs/creating_execs_and_libs.rst
```diff
@@ -26,14 +27,17 @@ like ``blank_project`` in the previous section, we can start using BLT's macros.
 Creating an executable is as simple as calling the following macro:
 
 .. literalinclude:: tutorial/calc_pi/CMakeLists.txt
+   :start-after: _blt_tutorial_example_executable_start
```
kennyweiss: Thanks for fixing these!

docs/creating_execs_and_libs.rst
```diff
@@ -106,3 +115,8 @@ without the linker removing unused symbols in the larger library.
     blt_add_exectuble(NAME       helloWorld
                       SOURCES    main.cpp
                       DEPENDS_ON myObjectLibrary)
+
+.. note::
+  Due to record keeping on BLT's part to make object libraries as easy to use
+  as possible, you need to define object libraries before you use them
```
kennyweiss: Just to make sure that I understand this -- it sounds like typical CMake targets can be used before they are defined, but that is not the case for object libraries. Is that right?

docs/external_dependencies.rst
```diff
@@ -69,33 +72,46 @@ which uses MPI to parallelize the calculation over the integration intervals.
...
 
 .. literalinclude:: tutorial/calc_pi/test_2.cpp
+   :start-after: _blt_tutorial_calcpi_test2_main_start
+   :end-before:  _blt_tutorial_calcpi_test2_main_end
    :language: cpp
-   :lines: 40-54
+
+.. note::
+  While we have tried to ensure that BLT chooses the correct setup information
+  for MPI, but there are several niche cases where the default behavior is
```
kennyweiss: Typo -- please remove `but`

cmake/thirdparty/SetupMPI.cmake
white238: I have run into cases where the variables are wrong and cannot be solved by appending.

docs/creating_execs_and_libs.rst
white238: I actually looked for this feature when i started this and have no idea how i missed it.  So much better than line numbers.

docs/creating_execs_and_libs.rst
white238: Yes. I find it very strange but you can.

cmake/thirdparty/SetupMPI.cmake
kennyweiss: ok then.

PR #272, opened 2019-05-31
Update SetupMPI.cmake

metalberkeley:
This modification allows for smooth MPI+CUDA RAJA configuration and compilation using nvcc with ENABLE_FIND_MPI. It needs RAJA_EXISTS variable to be set ON in RAJA (part of a RAJA patch).

### DIFF
--- a/cmake/thirdparty/SetupMPI.cmake
+++ b/cmake/thirdparty/SetupMPI.cmake
@@ -125,21 +125,17 @@ if (ENABLE_FORTRAN)
     endif()
 endif()
 
-# Create the registered library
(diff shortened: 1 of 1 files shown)


cyrush:
It would be good to know which mpi compiler flags need the `-XCompiler` treatment. 

Typically this is only needed for cases where the compiler provides features that don't look like standard libs. Like pthreads, or openmp flags -- where the features you are accessing a built into the compiler and nvcc doesn't know the special cases.  

Given that mpi is a library nvcc should be less picky about how it is presented. 

Maybe this is a case where MPI compiler wrappers are echoing magic flags?

metalberkeley:
Canceling the PR because of version mismatch between RAJA's BLT and BLT/master.

PR #276, opened 2019-06-12
Just punt on generator expressions

### LINKED ISSUE #275: cuda support fails when passed an object lib
blt add library with one of the sources like the following:

` $<TARGET_OBJECTS:blah>`

fails, in this macro:
```
macro(blt_split_source_list_by_language)

    set(options)
    set(singleValueArgs C_LIST Fortran_LIST)
    set(multiValueArgs SOURCES)

    # Parse the arguments
    cmake_parse_arguments(arg "${options}" "${singleValueArgs}"
                            "${multiValueArgs}" ${ARGN} )

    # Check arguments
    if ( NOT DEFINED arg_SOURCES )
        message( FATAL_ERROR "Must provide a SOURCES argument to the 'blt_split_source_list_by_language' macro" )
    endif()

    # Generate source lists based on language
    foreach(_file ${arg_SOURCES})
        get_filename_component(_ext ${_file} EXT)
        string(TOLOWER ${_ext} _ext_lower)

```

This returns empty:
` get_filename_component(_ext ${_file} EXT)`

Then, this fails:

`        string(TOLOWER ${_ext} _ext_lower)`

```
CMake Error at blt/cmake/BLTPrivateMacros.cmake:273 (string):
  string no output variable specified
  blt/cmake/BLTPrivateMacros.cmake:226 (blt_split_source_list_by_language)
  blt/cmake/BLTMacros.cmake:456 (blt_setup_cuda_source_properties)
  ascent/CMakeLists.txt:286 (blt_add_library)

```
(Line numbers might be off, using older version of blt)

white238:
Resolves #275 

There isn't anything we can do other than ignore these.  If a user gives a generator expressions we should just ignore it when it comes to setting properties on source files.  CMake does not provide a way to evaluate generator expressions.

### DIFF
--- a/cmake/BLTPrivateMacros.cmake
+++ b/cmake/BLTPrivateMacros.cmake
@@ -463,6 +463,12 @@ macro(blt_split_source_list_by_language)
 
     # Generate source lists based on language
     foreach(_file ${arg_SOURCES})
+        # Assume the user knows what they are doing if using a
+        # generator expressions because BLT can't evaluate them
+        if(${_file } MATCHES "^$<")
+            continue()
+        endif()
+
         get_filename_component(_ext ${_file} EXT)
         string(TOLOWER ${_ext} _ext_lower)
 
--- a/cmake/BLTPrivateMacros.cmake
+++ b/cmake/BLTPrivateMacros.cmake
@@ -465,7 +465,7 @@ macro(blt_split_source_list_by_language)
     foreach(_file ${arg_SOURCES})
         # Assume the user knows what they are doing if using a
         # generator expressions because BLT can't evaluate them
-        if(${_file } MATCHES "^$<")
+        if(${_file} MATCHES "^$<")
             continue()
         endif()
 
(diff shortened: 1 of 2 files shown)


cmake/BLTPrivateMacros.cmake
```diff
@@ -463,6 +463,12 @@ macro(blt_split_source_list_by_language)
 
     # Generate source lists based on language
     foreach(_file ${arg_SOURCES})
+        # Assume the user knows what they are doing if using a
```
kennyweiss: Should we also be printing out any files that cannot be split (i.e. in this list or in the `else()` clause below)?

The assumption that users know what they are doing is probably fine for the users that create these generator expressions, but not necessarily for their consumers.

How about creating a list of any variables that cannot be split and printing them out after the `foreach()`.
We can combine this with a separate variable that indicates if there is a `FATAL_ERROR` condition, which at this point would be a file that is not a generator expression or tht does not have a C- or Fortran extension.

cyrush: we could also filter these before we call split_source_list - however @white238's change is the most surgical fix.

To make other crazy situations easier to debug in the future, it would be good to error if we have an empty extension (and state that fact, showing the input). 
It took me quite a bit to understand where things went off the rails.

white238:
PR still good guys?

PR #279, opened 2019-06-14
Update GTest to #f7c178 (2019-06-14)

davidbeckingsale:
Contains lots of updates, but most important is fixing builds with MSVC when using C++17.


davidbeckingsale:
Google Test now requires C++11, so maybe we should leave this for now. I think I can fix the MSVC C++17 issue with a patch to your current version.

Thoughts? @kennyweiss @white238

davidbeckingsale:
Closing in favor of #280

PR #280, opened 2019-06-14
Add patch to make GoogleTest work with MSVC and C++ 17

davidbeckingsale:
This is less invasive that #279


white238:
Good to go David!

PR #281, opened 2019-06-19
fixed blt_add_target_link_flags for cmake >= 3.13

### LINKED ISSUE #266: issue with cmake 3.13.2, blt + rpath?
Did a test run updating Conduit to latest blt, one of my test cases using CMake 3.13.2 failed. 

```
[  3%] Building CXX object blt/tests/smoke/CMakeFiles/blt_mpi_smoke.dir/blt_mpi_smoke.cpp.o
[  4%] Linking CXX executable ../../../tests/blt_mpi_smoke
/usr/bin/ld: bad -rpath option
collect2: error: ld returned 1 exit status
make[2]: *** [tests/blt_mpi_smoke] Error 1
make[1]: *** [blt/tests/smoke/CMakeFiles/blt_mpi_smoke.dir/all] Error 2
make: *** [all] Error 2
```

https://circleci.com/gh/LLNL/conduit/452

Not 100% sure it's a blt issue --  Conduit does set it's own rpath rules, but those didn't change in this case. Wanted to share to make sure I document current knowledge of the mystery.

rrsettgast:
changed blt_add_target_link_flags to convert input flags to list, and loop over list

resolves #266

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -270,9 +270,9 @@ macro(blt_add_target_link_flags)
             # In CMake 3.13+, LINK_FLAGS was converted to LINK_OPTIONS.
             # This now supports generator expressions but expects a list
             # not a string
-            string(REPLACE " " ";" _flag_list ${arg_FLAGS})
+            separate_arguments(_flag_list NATIVE_COMMAND "${arg_FLAGS}" )
             foreach( _flag  ${_flag_list} )
-                set_property(TARGET ${arg_TO} APPEND PROPERTY LINK_OPTIONS ${_flag})
+                set_property(TARGET ${arg_TO} APPEND PROPERTY LINK_OPTIONS "${_flag}" )
             endforeach()
         else()
             get_target_property(_link_flags ${arg_TO} LINK_FLAGS)
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -270,7 +270,10 @@ macro(blt_add_target_link_flags)
             # In CMake 3.13+, LINK_FLAGS was converted to LINK_OPTIONS.
             # This now supports generator expressions but expects a list
             # not a string
-            set_property(TARGET ${arg_TO} APPEND PROPERTY LINK_OPTIONS ${arg_FLAGS})
+            string(REPLACE " " ";" _flag_list ${arg_FLAGS})
+            foreach( _flag  ${_flag_list} )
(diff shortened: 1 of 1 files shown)


cmake/BLTMacros.cmake
```diff
@@ -270,7 +270,10 @@ macro(blt_add_target_link_flags)
             # In CMake 3.13+, LINK_FLAGS was converted to LINK_OPTIONS.
             # This now supports generator expressions but expects a list
             # not a string
-            set_property(TARGET ${arg_TO} APPEND PROPERTY LINK_OPTIONS ${arg_FLAGS})
+            string(REPLACE " " ";" arg_FLAGS_LIST ${arg_FLAGS})
```
white238: Please rename arg_FLAG_LIST and arg_FLAG to _flag_list and _flag respective to not make it look like it is coming in as an argument.

rrsettgast:
would it be appropriate to use 
```
target_link_options( ${arg_TO} PUBLIC "${arg_FLAG}" )
```
in place of
```
set_property(TARGET ${arg_TO} APPEND PROPERTY LINK_OPTIONS ${arg_FLAGS})
```
? Or should the `INTERFACE_LINK_OPTIONS ` not be set also?

cmake/BLTMacros.cmake
rrsettgast: done

white238:
This was discussed in the previous PR (#254). The concern was there would be spaces in the arguments, which is something we can't guard against.

rrsettgast:
> This was discussed in the previous PR (#254). The concern was there would be spaces in the arguments, which is something we can't guard against.


well, the arg comes in as a string, and `LINK_OPTIONS` expects a list....so is the solution to make it a list in `SetupMPI()`?

white238:
I'd vote to use separate_arguments inside of blt_add_target_link_flags

kennyweiss:
I agree with @white238 about trying `separate_arguments()` instead of explicitly replacing the spaces since it is not clear that there will never be valid spaces in the `args_FLAGS` variable.  

Thanks for the suggestion in #281  @gzagaris

rrsettgast:
@white238 @kennyweiss My question is more targeted toward, should we check that the input to `blt_add_target_link_flags()` is indeed a list and error out. And separately in `SetupMPI()` convert `_mpi_link_flag` from a string into a list.

kennyweiss:
@rrsettgast -- I'd be ok with either path.  

Fixing it in `SetupMPI` would resolve the immediate issue.

On the other hand, if `separate_arguments` does the right thing, then that would be a more general solution.

PR #291, opened 2019-07-19
Update SetupOpenMP.cmake

metere1llnl:
Without this fix RAJA won't compile on my desktop machine because the if (BLT_OPENMP_COMPILE/LINK_FLAGS)  overwrite the CUDA conditional above.

### DIFF
--- a/cmake/thirdparty/SetupOpenMP.cmake
+++ b/cmake/thirdparty/SetupOpenMP.cmake
@@ -20,16 +20,16 @@ else()
 endif()
 
 set(_compile_flags ${OpenMP_CXX_FLAGS})
-set(_link_flags  ${OpenMP_CXX_FLAGS})
+set(_link_flags    ${OpenMP_CXX_FLAGS})
 
 if(NOT COMPILER_FAMILY_IS_MSVC AND ENABLE_CUDA AND ESCAPE_FORTRAN)
     set(_compile_flags
-        $<$<AND:$<NOT:$<COMPILE_LANGUAGE:CUDA>>,$<NOT:$<COMPILE_LANGUAGE:Fortran>>>:${OpenMP_CXX_FLAGS}> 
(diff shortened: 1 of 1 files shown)


cmake/thirdparty/SetupOpenMP.cmake
```diff
@@ -40,13 +40,24 @@ endif()
 
 # Allow user to override
 if (BLT_OPENMP_COMPILE_FLAGS)
```
kennyweiss: Note: This `if()` conditional should only be active when there is a `BLT_OPENMP_COMPILE_FLAGS` variable in scope, which should only happen if it is defined in your host-config (or the application build system that brings in blt).

metere1llnl: both BLT_OPENMP_COMPILE_FLAGS and BLT_OPENMP_LINK_FLAGS get autofilled. Nothing I added manually.

white238:
This PR breaks the functionality of allowing users to completely override the flags the come out of SetupOpenMP.  Either you need to use the override variables BLT_OPENMP_COMPILE_FLAGS and BLT_OPENMP_LINK_FLAGS or show the case in where the above generator expressios dont work for you.  Though keep in mind those generator expressions have been iterated over quite a few times and I am hesitant to change them.

Can you email me more about your case?

metere1llnl:
I'm trying to use BLT to compile with MPI, CUDA and OpenMP. In such a scenario, the options "-fopenmp" and "-fpthread" are generated, and when passed to nvcc, they need to be wrapped accordingly with the "-XCompiler" or "-Xlinker" options, otherwise the compilation fails. With the modifications in this PR, such problem ceases to exist.

white238:
I am not doubting it works on your configuration.  What I am saying is we should not be altering the BLT_OPENMP flags because they are specifically for users to override all of BLT's behavior regarding those flags.  If we alter the override behaviors it defeats their entire purpose. Also you wouldn't be hitting those if statements unless you have them set before BLT is loaded, most likely in your host-config.  Do you have  PR i can look at?

PR #300, opened 2019-08-09
API docs

white238:
Creating new API docs in sphinx.  The idea is to pull out all but a single sentence and the macros arguments into sphinx docs that are easy to read/navigate for users.

https://llnl-blt.readthedocs.io/en/feature-white238-apidocs/


docs/api/utility.rst
```diff
@@ -0,0 +1,325 @@
...
+General Utility Macros
+----------------------
+
+
+blt_list_append
+~~~~~~~~~~~~~~~
+
+.. code-block:: cmake
+
+    blt_list_append( TO <list> ELEMENTS [ <element>...] IF <bool> )
```
kennyweiss: I think it would be easier to read these macro prototypes if each argument was on its own line.

docs/api/utility.rst
```diff
@@ -0,0 +1,325 @@
...
+
+Appends elements to a list if the specified bool evaluates to true.
+
+This macro is essentially a wrapper around CMake's `list(APPEND ...)`
```
kennyweiss: I think these need to be double quoted to display properly in sphinx.
I.e.
```
``list(APPEND ...)``
```
instead of
```
`list(APPEND ...)`
```

docs/api/utility.rst
```diff
@@ -0,0 +1,325 @@
...
+command which allows inlining a conditional check within the same call
+for clarity and convenience.
+
+This macro requires specifying:
+
+    * The target list to append to by passing TO <list>
```
kennyweiss: I think the API docs would be clearer if the arguments were given as a "definition list" instead of a bulleted list.

See: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html?highlight=definition%20list

docs/api/utility.rst
```diff
@@ -0,0 +1,325 @@
...
+
+
+General Utility Macros
+----------------------
+
+
+blt_list_append
+~~~~~~~~~~~~~~~
+
+.. code-block:: cmake
```
kennyweiss: Would it be clearer if the prototypes/signatures had a caption, e.g. ``:caption: **Signature** `` ?

docs/api/utility.rst
```diff
@@ -0,0 +1,325 @@
...
+
+    blt_list_append( TO <list> ELEMENTS [ <element>...] IF <bool> )
+
+Appends elements to a list if the specified bool evaluates to true.
```
kennyweiss: It might be easier to read if the one-line descriptions were listed above the signatures

docs/api/utility.rst
white238: I agree.. though I'm going to leave that until a follow up PR which restructures a lot of the content.

docs/api/utility.rst
white238: Trying that in the next build for the first utility.

docs/api/utility.rst
white238: fixed.

docs/api/utility.rst
white238: fixed.

white238:
> Thanks @white238 -- this is a great start.
> 
> Some suggestions (for the future):
> 
>     * Would it make sense for the documentation of each macro to be in its own file (as in the cmake docs)?
> 
>     * It might also be a good idea to add a subsection to each macro doc that discusses work arounds that we implemented for  behavior changes in different cmake versions, where applicable.

I am torn on the each macro in its own file.. This would blow up how many clicks you need to get around.  If we were as large as the CMake docs I would completely agree but we aren't.

As for the work arounds, yes. That will come next in our/me improving what docs are already there.

docs/api/utility.rst
white238: Due to sphinx formatting it was weird to have it there...

white238:
I am happy with the first cut of this.  The getting the general layout to readable and get the content into one place.  I will merge this soon and then work on documenting all of the parameters first.

Issue #307, opened 2019-09-12
Compile definitions all public


mclarsen:
Compile definitions added to a target through BLT are all public and are propagated to all downstream deps. Is this a design decision or would it be appropriate to add a flag for `PUBLIC' versus`PRIVATE`?

I note that all of our compile definitions do not need to be propagated, and in one case should not be propagated.

white238:
@mclarsen does #311 solve your problem?

mclarsen:
@white238 Yes, that would solve my problem.

Issue #305, opened 2019-08-27
Improve the consistency of BLT's flag-related macros


kennyweiss:
We have several macros that deal with compiler flags, but they do not consistently treat the flags or share an underlying implementation.  Some of them return a CMake list, while others return a CMake string with spaces. 

These should be made more consistent and should work better with each other.

### Details

* There was some effort involved in getting ``blt_add_target_link_flags`` to work with different cmake versions.  This involved using CMake's ``separate_arguments`` macro to properly separate command line arguments.

* Our ``blt_append_custom_compiler_flag()`` macro returns a string with spaces, which is not directly compatible with the underlying macros that add flags to targets, e.g. CMake's ``target_compile_options()`` or blt's ``blt_add_target_compile_flags()``.

* ``blt_add_target_compile_flags`` should also take an extra parameter to indicate if the flags should be made ``PUBLIC``, ``PRIVATE`` or ``INTERFACE``, which would override the default (based on the target type).

Related issues: #181, #234, #236, #307

cyrush:
+1 for adding `PUBLIC`, `PRIVATE` or `INTERFACE` option to `blt_add_target_compile_flags`

We have a case where we need `PRIVATE` and it's going to be messy w/o blt

white238:
I have a fix I think... ill put it in a PR.

PR #311, opened 2019-09-20
Add scope to target property macros

white238:
Addresses #307 #305.  This won't fix any registered targets from being PUBLIC that is a lot trickier and will be solved with imported targets.  I actually started working on that when I realized this would simplify some of the work there.

Docs to come later in this PR...

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,7 +22,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   (see https://gitlab.kitware.com/cmake/cmake/issues/18008).
 - In CMake 3.13+, "SHELL:" was added to blt_add_target_link_flags.  This stops CMake from de-duplicating
   needed linker flags.
-- Added SCOPE to all target property macros, blt_add_target_link_flags, etc.  It defaults to PUBLIC.
+- Added optional SCOPE to all target property macros, blt_add_target_link_flags, etc.  It defaults to PUBLIC.
 
 ### Changed
 - Restructured the host-config directory by site and platform.
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,6 +22,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   (see https://gitlab.kitware.com/cmake/cmake/issues/18008).
 - In CMake 3.13+, "SHELL:" was added to blt_add_target_link_flags.  This stops CMake from de-duplicating
   needed linker flags.
+- Added SCOPE to all target property macros, blt_add_target_link_flags, etc.  It defaults to PUBLIC.
 
 ### Changed
 - Restructured the host-config directory by site and platform.
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -77,8 +77,8 @@ endmacro(blt_list_remove_duplicates)
 
 
 ##------------------------------------------------------------------------------
-## blt_add_target_definitions(TO <target>
-##                            SCOPE <public (default)|private|interface>
+## blt_add_target_definitions(TO     <target>
+##                            SCOPE  <PUBLIC (Default)| INTERFACE | PRIVATE>
 ##                            TARGET_DEFINITIONS [FOO [BAR ...]])
 ##
 ## Adds pre-processor definitions to the given target.
@@ -114,7 +114,7 @@ endmacro(blt_add_target_definitions)
 
 ##------------------------------------------------------------------------------
 ## blt_add_target_compile_flags(TO    <target>
-##                              SCOPE <public (default)|private|interface>
+##                              SCOPE  <PUBLIC (Default)| INTERFACE | PRIVATE>
 ##                              FLAGS [FOO [BAR ...]])
 ##
 ## Adds compiler flags to a target (library, executable or interface) by 
@@ -187,7 +187,7 @@ endmacro(blt_set_target_folder)
 
 ##------------------------------------------------------------------------------
 ## blt_add_target_link_flags (TO    <target>
-##                            SCOPE <public (default)|private|interface>
+##                            SCOPE <PUBLIC (Default)| INTERFACE | PRIVATE>
 ##                            FLAGS [FOO [BAR ...]])
 ##
 ## Adds linker flags to a target by appending to the target's existing flags.
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -20,7 +20,7 @@ macro(blt_list_append)
     cmake_parse_arguments(arg
         "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN} )
 
-     # sanity checks
+     # Sanity checks
     if( NOT DEFINED arg_TO )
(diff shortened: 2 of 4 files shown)


cmake/BLTPrivateMacros.cmake
```diff
@@ -8,6 +8,52 @@ include(CMakeParseArguments)
 ## Internal BLT CMake Macros
 
 
+##-----------------------------------------------------------------------------
+## blt_determine_scope(TARGET <target>
```
kennyweiss: This macro is nice.

It looks like you are handling the case to make this function case-insensitive.
It might be a bit confusing for the documentation to list the scopes using lower case since cmake actually uses upper case.  

Could you please either switch the case in the documentation to upper case, or add a note that the function is case sensitive, but returns the scope in ``OUT`` as upper case?

cmake/BLTPrivateMacros.cmake
```diff
@@ -8,6 +8,52 @@ include(CMakeParseArguments)
...
+
+    string(TOUPPER "${arg_SCOPE}" _uppercaseScope)
+    if("${_uppercaseScope}" MATCHES "")
+        # Default to public
+        set(_uppercaseScope PUBLIC)
+    elseif(NOT ("${_uppercaseScope}" STREQUAL "PUBLIC" OR
+                "${_uppercaseScope}" STREQUAL "INTERFACE" OR
+                "${_uppercaseScope}" STREQUAL "PRIVATE"))
+        message(FATAL_ERROR "Given SCOPE (${arg_SCOPE}) is not valid, valid options are:"
+                            "public, interface, private")
```
kennyweiss: Here too, consider making these upper case in the error message..

cmake/BLTPrivateMacros.cmake
```diff
@@ -8,6 +8,52 @@ include(CMakeParseArguments)
 ## Internal BLT CMake Macros
 
 
+##-----------------------------------------------------------------------------
+## blt_determine_scope(TARGET <target>
+##                     SCOPE  <public (Default)| interface | private >
```
kennyweiss: It might be worthwhile adding a note here that the default is ``PUBLIC`` for regular targets, but ``INTERFACE`` for ``INTERFACE_LIBRARY`` targets. This might help avoid user confusion.

cmake/BLTPrivateMacros.cmake
```diff
@@ -8,6 +8,52 @@ include(CMakeParseArguments)
...
+                            "public, interface, private")
+    endif()
+
+    if(TARGET ${arg_TARGET})
+        get_property(_targetType TARGET ${arg_TO} PROPERTY TYPE)
+        if(${_targetType} STREQUAL "INTERFACE_LIBRARY")
+            # Interface targets can only set INTERFACE
+            set(${arg_OUT} INTERFACE)
+        else()
+            set(${arg_OUT} ${_uppercaseScope})
```
kennyweiss: Is ``PRIVATE`` valid for ``INTERFACE_LIBRARY`` scopes? (I would guess not).

cmake/BLTMacros.cmake
```diff
@@ -209,19 +209,23 @@ macro(blt_add_target_link_flags)
                        _flags "${_flags}")
     endif()
 
+    # Only add the flag if it is not empty
```
kennyweiss: Wow -- there are a lot of special cases here!

cmake/BLTMacros.cmake
```diff
@@ -77,76 +77,74 @@ endmacro(blt_list_remove_duplicates)
 
 
 ##------------------------------------------------------------------------------
-## blt_add_target_definitions(TO <target> TARGET_DEFINITIONS [FOO [BAR ...]])
+## blt_add_target_definitions(TO <target>
+##                            SCOPE <public (default)|private|interface>
```
kennyweiss: Do we have a way to indicate optional vs. required parameters?
Similarly for ``options`` vs. ``singleValueArgs`` vs. ``multiValueArgs``?

cmake/BLTPrivateMacros.cmake
```diff
@@ -8,6 +8,52 @@ include(CMakeParseArguments)
...
 
 
+##-----------------------------------------------------------------------------
+## blt_determine_scope(TARGET <target>
+##                     SCOPE  <public (Default)| interface | private >
+##                     OUT    <out variable name>)
+##
+## Returns the scope from a given target and scope string
+##-----------------------------------------------------------------------------
+macro(blt_determine_scope)
```
gzagaris: I am a little confused on why the `SCOPE` argument is needed here, but, I am probably missing something.... 

It seems that the intent of the `blt_determine_scope()` macro is to determine the scope of a "target", i.e., PUBLIC, INTERFACE, etc.  Why does the caller need to specify a `SCOPE` argument? Could this macro just take two arguments: (a) the target, and (b) `SCOPE` as an output argument? 

Please, correct me if I misunderstood the intent of this.

cmake/BLTPrivateMacros.cmake
```diff
@@ -8,6 +8,52 @@ include(CMakeParseArguments)
...
+
+    set(options)
+    set(singleValueArgs TARGET SCOPE)
+    set(multiValueArgs OUT)
```
gzagaris: Should `OUT` be a multi-value argument?

cmake/BLTPrivateMacros.cmake
```diff
@@ -8,6 +8,52 @@ include(CMakeParseArguments)
...
+        set(_uppercaseScope PUBLIC)
+    elseif(NOT ("${_uppercaseScope}" STREQUAL "PUBLIC" OR
+                "${_uppercaseScope}" STREQUAL "INTERFACE" OR
+                "${_uppercaseScope}" STREQUAL "PRIVATE"))
+        message(FATAL_ERROR "Given SCOPE (${arg_SCOPE}) is not valid, valid options are:"
+                            "public, interface, private")
+    endif()
+
+    if(TARGET ${arg_TARGET})
+        get_property(_targetType TARGET ${arg_TO} PROPERTY TYPE)
```
gzagaris: Is `${arg_TO}` a typo here? Should this be `${arg_TARGET}`?

white238:
I have updated the documentation both for the private macro (in file) and the API docs. Which can be seen here:

https://llnl-blt.readthedocs.io/en/feature-white238-targetpropertyscopes/api/target_properties.html

After discussing the PR with @gzagaris and @kennyweiss , I think we came to an agreement about this being okay now.

kennyweiss:
Thanks @white238. The docs look good.

Minor issue: The ``note`` for ``blt_add_target_definitions`` looks a bit weird with the code box outside the note. Is there any way to include it in the note?

RELEASE-NOTES.md
```diff
@@ -22,6 +22,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   (see https://gitlab.kitware.com/cmake/cmake/issues/18008).
 - In CMake 3.13+, "SHELL:" was added to blt_add_target_link_flags.  This stops CMake from de-duplicating
   needed linker flags.
+- Added SCOPE to all target property macros, blt_add_target_link_flags, etc.  It defaults to PUBLIC.
```
kennyweiss: Suggestion: ``Added *optional SCOPE parameter* to all...``

PR #312, opened 2019-09-25
Clarify code check docs, attempt to fix github's auto-detect of license

white238:
Clarifies more docs.
Page that changed: https://llnl-blt.readthedocs.io/en/feature-white238-clarifydocs/api/code_check.html

@tgamblin noticed we were not being autodetected by github's license logic due to dual files.  This hopefully fixes it.


white238:
It's not auto-detecting the license in my branch.  Maybe after it merges to develop?

tgamblin:
I think license detection only runs every so often, and maybe only on the default branch.  I copied CHAI's license, which is detected properly, and stuck it in here.  I *think* this should work once merged.

tgamblin:
If it doesn't the only difference I see is that CHAI's actually says BSD 3-clause License on the first line -- That isn't in the standard template so I left it out.  https://github.com/LLNL/CHAI/edit/develop/LICENSE

PR #313, opened 2019-09-26
Bugfix for blt_determine_scope

kennyweiss:
Matching string to empty string is always true.


kennyweiss:
The following line:
```
  blt_add_target_compile_flags(TO <target> SCOPE private FLAGS <flags>)
```
was incorrectly setting the scope to ``PUBLIC`` rather than ``PRIVATE``.

cmake/BLTPrivateMacros.cmake
```diff
@@ -33,8 +33,11 @@ macro(blt_determine_scope)
     cmake_parse_arguments(arg "${options}" "${singleValueArgs}"
                         "${multiValueArgs}" ${ARGN} )
 
+    # Convert to upper case and strip white space
     string(TOUPPER "${arg_SCOPE}" _uppercaseScope)
-    if("${_uppercaseScope}" MATCHES "")
```
kennyweiss: This was always evaluating as ``TRUE``

PR #316, opened 2019-09-30
finding openmp on mac requires uppercase

nicolasaunai:
OpenMP is randomly found/not found unless OpenMP is all uppercase on my mac.
cmake 3.15.3, clang 9, libomp 8.0.1 on two machines, one finds it the other not. if in uppercase, both find it.

fix found there: 
https://stackoverflow.com/a/53823820

have not tested on linux wether the uppercase version works.

### DIFF
--- a/cmake/thirdparty/SetupOpenMP.cmake
+++ b/cmake/thirdparty/SetupOpenMP.cmake
@@ -8,7 +8,7 @@
 # (OpenMP support is provided by the compiler)
 #################################################
 
-find_package(OpenMP REQUIRED)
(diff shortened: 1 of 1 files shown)


white238:
Thanks for the fix @nicolasaunai . I need to test this on our various platforms we support. I should get to it by the end of the day.

white238:
I just can't reproduce this on any mac machine we have.  I am going to chalk this up to a weird versioning bug either on the openmp or cmake side.  I have seen a lot of cmake bugs that come and go with versions.

PR #317, opened 2019-11-03
Fix Fortran OpenMP flags

### LINKED ISSUE #175: The BLT provided "openmp" pseudotarget doesn't (always) pass the correct openMP fortran flags

gzagaris:
Resolves #175

### DIFF
--- a/cmake/BLTPrivateMacros.cmake
+++ b/cmake/BLTPrivateMacros.cmake
@@ -1,6 +1,6 @@
 # Copyright (c) 2017-2019, Lawrence Livermore National Security, LLC and
 # other BLT Project Developers. See the top-level COPYRIGHT file for details
-# 
+#
 # SPDX-License-Identifier: (BSD-3-Clause)
 
 include(CMakeParseArguments)
@@ -83,6 +83,34 @@ function(blt_error_if_target_exists target_name error_msg)
     endif()
 endfunction()
 
+##-----------------------------------------------------------------------------
+## blt_fix_fortran_openmp_flags()
+##
+## Fixes the openmp flags for a Fortran target if they are different from the
+## corresponding C/C++ openmp flags.
+##-----------------------------------------------------------------------------
+function(blt_fix_fortran_openmp_flags target_name)
+
+
+    if (ENABLE_FORTRAN AND ENABLE_OPENMP AND BLT_OPENMP_FLAGS_DIFFER)
+
+        get_target_property(target_link_flags ${target_name} LINK_FLAGS)
+        if ( target_link_flags )
+
+            message(STATUS "Fixing OpenMP Flags for target[${target_name}]")
+
+            string( REPLACE "${OpenMP_CXX_FLAGS}" "${OpenMP_Fortran_FLAGS}"
+                    correct_link_flags
+                    "${target_link_flags}"
+                    )
+
+            set_target_properties( ${target_name} PROPERTIES LINK_FLAGS
+                                   "${correct_link_flags}" )
+        endif()
+
+    endif()
+
+endfunction()
 
 ##-----------------------------------------------------------------------------
 ## blt_find_executable(NAME         <name of program to find>
@@ -213,8 +241,8 @@ endmacro(blt_inherit_target_info)
 
 
 ##------------------------------------------------------------------------------
-## blt_setup_target( NAME       [name] 
-##                   DEPENDS_ON [dep1 ...] 
+## blt_setup_target( NAME       [name]
+##                   DEPENDS_ON [dep1 ...]
 ##                   OBJECT     [TRUE | FALSE])
 ##------------------------------------------------------------------------------
 macro(blt_setup_target)
@@ -351,7 +379,7 @@ macro(blt_setup_cuda_target)
     endif()
 
     if (${_depends_on_cuda_runtime} OR ${_depends_on_cuda})
-        if (CUDA_LINK_WITH_NVCC) 
+        if (CUDA_LINK_WITH_NVCC)
             set_target_properties( ${arg_NAME} PROPERTIES LINKER_LANGUAGE CUDA)
         endif()
     endif()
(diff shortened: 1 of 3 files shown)


cmake/BLTPrivateMacros.cmake
```diff
@@ -83,6 +83,34 @@ function(blt_error_if_target_exists target_name error_msg)
     endif()
 endfunction()
 
+##-----------------------------------------------------------------------------
+## blt_fix_fortran_openmp_flags()
+##
+## Fixes the openmp flags for a Fotran target if they are different from the
```
white238: Typo in Fortran

cmake/BLTPrivateMacros.cmake
```diff
@@ -83,6 +83,34 @@ function(blt_error_if_target_exists target_name error_msg)
...
+##-----------------------------------------------------------------------------
+function(blt_fix_fortran_openmp_flags target_name)
+
+
+    if (ENABLE_FORTRAN AND ENABLE_OPENMP AND BLT_OPENMP_FLAGS_DIFFER)
+
+        get_target_property(target_link_flags ${target_name} LINK_FLAGS)
+        if ( target_link_flags )
+
+            message(STATUS "Fixing OpenMP Flags for target[${target_name}]")
```
kennyweiss: Will this message be too verbose for projects with a lot of fortran targets? Perhaps the message should be moved to ``SetupOpenMP``, where it will only be printed once (but will not list the targets that it modifies?

cmake/thirdparty/SetupOpenMP.cmake
```diff
@@ -10,7 +10,15 @@
...
+set(BLT_OPENMP_FLAGS_DIFFER FALSE CACHE BOOL "")
+if (ENABLE_FORTRAN)
+  string(COMPARE NOTEQUAL "${OpenMP_CXX_FLAGS}" "${OpenMP_Fortran_FLAGS}"
+         BLT_OPENMP_FLAGS_DIFFER )
+endif()
+
+# avoid generator expressions if possible, as generator expressions can be
 # passed as flags to downstream projects that might not be using the same
 # languages. See https://github.com/LLNL/blt/issues/205
 if (ENABLE_FORTRAN AND NOT OpenMP_CXX_FLAGS STREQUAL OpenMP_Fortran_FLAGS)
```
kennyweiss:
```cmake
if (ENABLE_FORTRAN AND NOT OpenMP_CXX_FLAGS STREQUAL OpenMP_Fortran_FLAGS)
```

Can this whole line be replaced by the new variable? 
E.g. 
```cmake
if(BLT_OPEMP_FLAGS_DIFFER)
```

cmake/BLTPrivateMacros.cmake
gzagaris: Fixed. Thanks.

cmake/BLTPrivateMacros.cmake
gzagaris: Let's leave it for now. I'd rather see an explicit message that flags are being modified under the hood. We can change it later if it gets annoying.

cmake/thirdparty/SetupOpenMP.cmake
gzagaris: Yes -- upon closer inspection, it looked like the older code was setting another variable, `ESCAPE_FORTRAN`. Since we are defining the `BLT_OPENMP_FLAGS_DIFFER` cache variable now, that variable is no longer needed. I changed those checks accordingly.

PR #320, opened 2019-11-21
Fix another object library use case

white238:
This fixes the following use case:

registered library (NAME foo LIBRARIES actual-cmake-target)
blt_add_library(NAME bar DEPENDS_ON foo)


cmake/BLTPrivateMacros.cmake
```diff
@@ -189,7 +189,7 @@ endmacro(blt_find_executable)
 ##------------------------------------------------------------------------------
 macro(blt_inherit_target_info)
     set(options)
-    set(singleValueArgs TO FROM)
+    set(singleValueArgs TO FROM OBJECT)
```
kennyweiss: good catch!

PR #323, opened 2019-12-04
New option: enable changing /MD to /MT

agcapps:
When building with MS Visual Studio, the /MD command line option links in the dynamic C runtime and /MT links the static runtime.  At the link step, every binary must have used the same runtime link flag or link errors will result.  Some libraries' build systems insist on one flag or the other: CMake only uses /MD, HDF5 strongly prefers /MD for both dynamic and static builds, while GTest changes all occurences of /MD to /MT for static builds.

This commit establishes ENABLE_MSVC_STATIC_MD_TO_MT, defaulting to ON.  This flag has effect only if building static binaries using MS Visual Studio.  If the option is OFF, BLT will leave all /MD flags generated by CMake alone and throw an error if GTest is enabled for a static build under MSVC.  If ON, BLT will change each occurence of /MD in the compiler flags to /MT to match (and enable) GTest.

### DIFF
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -6,6 +6,8 @@
 # Sets up configuration options for BLT
 #------------------------------------------------------------------------------
 
+include(CMakeDependentOption)
+
 #------------------------------------------------------------------------------
 # Build Targets
 #------------------------------------------------------------------------------
@@ -82,7 +84,12 @@ endif()
 #------------------------------------------------------------------------------
 option(ENABLE_ALL_WARNINGS         "Enables all compiler warnings on all build targets" ON)
 option(ENABLE_WARNINGS_AS_ERRORS   "Enables treating compiler warnings as errors on all build targets" OFF)
-option(ENABLE_MSVC_STATIC_MD_TO_MT "For static linking with MS Visual Studio, enables changing /MD to /MT" ON)
+cmake_dependent_option(BLT_ENABLE_MSVC_STATIC_MD_TO_MT
+                                   "When linking statically with MS Visual Studio, enables changing /MD to /MT"
+                                   ON
+                                   "NOT BUILD_SHARED_LIBS"
+                                   OFF)
+mark_as_advanced(BLT_ENABLE_MSVC_STATIC_MD_TO_MT)
 
 #------------------------------------------------------------------------------
 # Generator Options
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -80,8 +80,9 @@ endif()
 #------------------------------------------------------------------------------
 # Compiler Options
 #------------------------------------------------------------------------------
-option(ENABLE_ALL_WARNINGS       "Enables all compiler warnings on all build targets" ON)
-option(ENABLE_WARNINGS_AS_ERRORS "Enables treating compiler warnings as errors on all build targets" OFF)
+option(ENABLE_ALL_WARNINGS         "Enables all compiler warnings on all build targets" ON)
+option(ENABLE_WARNINGS_AS_ERRORS   "Enables treating compiler warnings as errors on all build targets" OFF)
+option(ENABLE_MSVC_STATIC_MD_TO_MT "For static linking with MS Visual Studio, enables changing /MD to /MT" ON)
 
 #------------------------------------------------------------------------------
 # Generator Options
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -333,12 +333,28 @@ blt_append_custom_compiler_flag(
      )
 
 #
-# Modify flags to avoid static linking runtime issues on windows.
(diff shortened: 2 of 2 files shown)


cmake/BLTOptions.cmake
```diff
@@ -80,8 +80,9 @@ endif()
 #------------------------------------------------------------------------------
 # Compiler Options
 #------------------------------------------------------------------------------
-option(ENABLE_ALL_WARNINGS       "Enables all compiler warnings on all build targets" ON)
-option(ENABLE_WARNINGS_AS_ERRORS "Enables treating compiler warnings as errors on all build targets" OFF)
+option(ENABLE_ALL_WARNINGS         "Enables all compiler warnings on all build targets" ON)
+option(ENABLE_WARNINGS_AS_ERRORS   "Enables treating compiler warnings as errors on all build targets" OFF)
+option(ENABLE_MSVC_STATIC_MD_TO_MT "For static linking with MS Visual Studio, enables changing /MD to /MT" ON)
```
agcapps: @kennyweiss suggested making this option dependent on BUILD_STATIC_LIBS.  I agree

agcapps: @kennyweiss suggested making this option dependent on whether we're using MSVC.  I don't see any way to do this at this point.  CMake code further in the configuration process decides whether we're using MSVC based on the build tool in use (see SetupCompilerOptions.cmake line 19).

agcapps: Also need to mark this option as advanced

kennyweiss: ok -- probably not worth it then, especially if the option is marked as advanced.

kennyweiss:
Note for the future: We should discuss this option in the user docs (when that page exists #324 )

cmake/SetupCompilerOptions.cmake
```diff
@@ -338,16 +338,21 @@ blt_append_custom_compiler_flag(
 #
```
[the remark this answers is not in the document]
agcapps: I'll include a discussion of the reasons projects might want to do this.

I don't think we can always link dynamically: someone might have a requirement to link gtest statically on Windows (or unknown_future_library, if not gtest).

cmake/BLTOptions.cmake
white238: Please prefix this option with BLT_

cmake/SetupCompilerOptions.cmake
```diff
@@ -338,16 +338,21 @@ blt_append_custom_compiler_flag(
...
+          set(_flag CMAKE_${_lang}_${_build})
+          if(${_flag} MATCHES "/MD")
+              string(REGEX REPLACE "/MD" "/MT" ${_flag} "${${_flag}}")
+          endif()
+      endforeach()
     endforeach()
-  endforeach()
+  elseif (ENABLE_GTEST)
+    message(FATAL_ERROR
+      "For static linking with MS Visual Studio using GTEST, you must enable changing /MD to /MT.")
```
white238: Reference the option they turned off to cause this

cyrush:
there are other considerations beyond gtest (other TPLS), it would be good for us to understand distill the issues in user docs (after this PR)

cmake/BLTOptions.cmake
agcapps: done

agcapps: done

cmake/SetupCompilerOptions.cmake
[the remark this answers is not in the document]
agcapps: Expanded the comment to discuss why it might be important.

cmake/SetupCompilerOptions.cmake
agcapps: done

cmake/SetupCompilerOptions.cmake
```diff
@@ -333,21 +333,42 @@ blt_append_custom_compiler_flag(
...
+# See https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2019.
+#
+# Mixing /MD with /MT can cause linking errors.  CMake specifies
+# /MD when generating project files for MSVC and provides no way to
+# change this.  This can be a problem with Google Test in particular,
+# which when building statically replaces all /MD with /MT.  HDF5, on
+# the other hand, sternly warns against the use of /MT, even when
+# built statically.
+#
+# See https://gitlab.kitware.com/cmake/community/wikis/FAQ#dynamic-replace.
```
kennyweiss: Thanks @agcapps -- this is interesting.

After this PR, we can consider directly incorporating this when the user's cmake is at ``3.15+``

cyrush: great note explaining this!

white238:
Sorry I forgot you can't merge this yourself @agcapps .  Merging now.

agcapps:
Thanks Chris.

PR #330, opened 2019-12-31
blt_add_test learned NUM_OMP_THREADS

white238:
#2 
#231

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -596,17 +596,18 @@ endmacro(blt_add_executable)
 
 
 ##------------------------------------------------------------------------------
-## blt_add_test( NAME           [name]
-##               COMMAND        [command] 
-##               NUM_MPI_TASKS  [n]
-##               CONFIGURATIONS [config1 [config2...]])
+## blt_add_test( NAME            [name]
+##               COMMAND         [command] 
+##               NUM_MPI_TASKS   [n]
+##               NUM_OMP_THREADS [n]
+##               CONFIGURATIONS  [config1 [config2...]])
 ##
 ## Adds a test to the project.
 ##------------------------------------------------------------------------------
 macro(blt_add_test)
 
     set(options )
-    set(singleValueArgs NAME NUM_MPI_TASKS)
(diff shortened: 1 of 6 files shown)


cmake/BLTMacros.cmake
```diff
@@ -596,17 +596,18 @@ endmacro(blt_add_executable)
...
 
 ##------------------------------------------------------------------------------
-## blt_add_test( NAME           [name]
-##               COMMAND        [command] 
-##               NUM_MPI_TASKS  [n]
-##               CONFIGURATIONS [config1 [config2...]])
+## blt_add_test( NAME            [name]
+##               COMMAND         [command] 
+##               NUM_MPI_TASKS   [n]
+##               NUM_OMP_THREADS [n]
```
kennyweiss: Would ``NUM_CPU_THREADS`` be better?

cmake/BLTMacros.cmake
```diff
@@ -664,6 +665,12 @@ macro(blt_add_test)
              COMMAND        ${test_command}
              CONFIGURATIONS ${arg_CONFIGURATIONS})
 
+    # Handle OpenMP
+    if( arg_NUM_OMP_THREADS )
+        set_property(TEST ${arg_NAME}
+                     APPEND PROPERTY ENVIRONMENT OMP_NUM_THREADS=${arg_NUM_OMP_THREADS})
```
kennyweiss: Really nice!

Presumably, this would override an existing value in the user's environment.

Does this property only apply when a user runs ``ctest`` or ``make test``?

kennyweiss:
@white238 -- would it make sense to provide a blt option to set the default number of threads for blt tests?  

(Alternatively, one could argue that this would add too much ``magic`` to the build system..)

Obviously, this could be in a separate PR.

cmake/BLTMacros.cmake
white238: I am torn on this.  On one hand OMP is specific, it avoids future conflicts with other threading libraries (maybe not a problem but i like being specific).  

I have spoken to several people (and plan on a few more) and the consensus so far is what I have.

cmake/BLTMacros.cmake
white238: It is my understanding that this only affects the named test.

Issue #293, opened 2019-07-19
Build failure under gcc-9.1.0


pjh40:
Attempting to build Umpire under gcc-9.1.0, I get the following error during the build:
```
[  3%] Building CXX object blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/CMakeFiles/gmock.dir/src/gmock-all.cc.o
In file included from /Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/include/gmock/gmock-spec-builders.h:70,
                 from /Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/include/gmock/gmock-generated-function-mockers.h:43,
                 from /Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/include/gmock/gmock.h:61,
                 from /Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/src/gmock-all.cc:40:
/Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/include/gmock/gmock-matchers.h: In instantiation of 'testing::PolymorphicMatcher<Impl>::PolymorphicMatcher(const Impl&) [with Impl = testing::internal::StrEqualityMatcher<std::__cxx11::basic_string<char> >]':
/Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/include/gmock/gmock-matchers.h:496:10:   required from 'testing::PolymorphicMatcher<Impl> testing::MakePolymorphicMatcher(const Impl&) [with Impl = testing::internal::StrEqualityMatcher<std::__cxx11::basic_string<char> >]'
/Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/include/gmock/gmock-matchers.h:3978:65:   required from here
/Users/phenning/Projects/blt/thirdparty_builtin/googletest-master-2018-02-01/googlemock/include/gmock/gmock-matchers.h:431:67: error: implicitly-declared 'testing::internal::StrEqualityMatcher<std::__cxx11::basic_string<char> >::StrEqualityMatcher(const testing::internal::StrEqualityMatcher<std::__cxx11::basic_string<char> >&)' is deprecated [-Werror=deprecated-copy]
  431 |   explicit PolymorphicMatcher(const Impl& an_impl) : impl_(an_impl) {}
[...]

Issue #289, opened 2019-07-17
Upgrade gtest for cmake@3.14


kennyweiss:
CMake@3.14 is generating a warning about gtest's use of policy `CMP0042`:

> CMake Deprecation Warning at .../blt/thirdparty_builtin/googletest-master-2018-02-01/googletest/CMakeLists.txt:65 (cmake_policy):
  The OLD behavior for policy CMP0042 will be removed from a future version
  of CMake.

> The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.

It appears to be resolved in a newer version of gtest, so we should probably upgrade our copy, when we get a chance.

Related issues: #50, #161

Issue #150, opened 2018-05-10
gtest macro warnings


rrsettgast:
When compiling with clang5, on my mac laptop there are compiler warnings from gtest about using NULL. The issue is when using the TEST macro (gtest), it gets expanded into the source, thus -system_include will not help here. The error points you exactly to the problem:

```
/Users/settgast1/Codes/geosx/GEOSX_github/src/components/core/tests/testGeosxTraits.cpp:32:1: warning: zero as null pointer constant [-Wzero-as-null-pointer-constant]
TEST(testGeosxTraits,test_is_tensorT)
^
/Users/settgast1/Codes/geosx/GEOSX_github/src/cmake/blt/thirdparty_builtin/googletest-master-2018-02-01/googletest/include/gtest/gtest.h:2250:42: note: expanded from macro 'TEST'
# define TEST(test_case_name, test_name) GTEST_TEST(test_case_name, test_name)
                                         ^
/Users/settgast1/Codes/geosx/GEOSX_github/src/cmake/blt/thirdparty_builtin/googletest-master-2018-02-01/googletest/include/gtest/gtest.h:2244:3: note: expanded from macro 'GTEST_TEST'
  GTEST_TEST_(test_case_name, test_name, \
  ^
/Users/settgast1/Codes/geosx/GEOSX_github/src/cmake/blt/thirdparty_builtin/googletest-master-2018-02-01/googletest/include/gtest/internal/gtest-internal.h:1233:38: note: expanded from macro 'GTEST_TEST_'
        #test_case_name, #test_name, NULL, NULL, \
                                     ^
/opt/local/libexec/llvm-5.0/lib/clang/5.0.1/include/stddef.h:100:18: note: expanded from macro 'NULL'
#    define NULL __null

```

I work around this by putting 
```
#define __null nullptr
```
into my test file source.

PR #332, opened 2020-01-08
Update GoogleTest and GoogleMock

white238:
This updates us to the current Master branch for googletest and googlemock.  I am 99% sure the only patch we needed was @kennyweiss 's deathtest patch.  The other 3 were pulled from their repo so were already there.

We were 2 years behind them which is just going to continue to cause us problems.  This does require anyone that uses these two libraries to use C++11 but I don't know anyone that is still stuck at this point.

#293
#289
#150
#109


white238:
Verified this also passes with Axom. Though there are a few now deprecated gtest features we are using.

Issue #340, opened 2020-02-04
GoogleTest/Benchmark have enabled warnings as errors


white238:
blt/thirdparty_builtin/googletest-master-2020-01-07/googletest/cmake/internal_utils.cmake
blt/thirdparty_builtin/benchmark-1.5.0/CMakeLists.txt

Seems like the google suite decided to add -Werror all over the place by default.  We probably should patch this out.

PR #347, opened 2020-03-11
Remove googletest adding -Werror

white238:
This was causing build errors in spack builds of very unimportant warnings that were not the fault of the code project.


white238:
Also not something you want the ability to not turn off propagating to all your tests.

white238:
Missed a file.  #340

PR #349, opened 2020-03-16
Upgrade FRUIT to 3.4.3 and namespace its API

goxberry:
Most of the FRUIT API is not namespaced, in that there isn't a naming convention to attempt to distinguish symbols exported by FRUIT from symbols defined by an application. The lack of namespacing increases the chance of name collisions between FRUIT and applications that use it. To fix this issue, this PR does two things:

* upgrades FRUIT to 3.4.3 (the most recent version, though `fruit.f90` is unchanged)

* changes name of each public symbol in FRUIT to start with `fruit_` or `FRUIT_` if it does not already do so

These changes have been tested within a fork of FRUIT using its internal test suite; see https://github.com/goxberry/FRUIT/tree/namespace-library for more details. 

The bash script https://github.com/goxberry/FRUIT/blob/namespace-library/namespace_fruit.sh might be useful for migrating existing code using the upstream FRUIT API. If said code uses the Fortran 2003 (or later) standard, renaming procedures in `use fruit` statements is another viable approach.


white238:
Thanks! @ltaylor16 is our resident fortran expert so I will leave this up to his opinion.  Please update the documentation (it is small) here: docs/tutorial/unit_testing.rst

white238:
My main concern is the maintenance required if we ever upgraded again and the fact this now breaks googling for help with FRUIT.  I do recognize that FRUIT is quite stable.  What brought this change on?

goxberry:
> What brought this change on?

@white238 I work at LLNL on software projects across multiple languages, and I have been encouraged to use BLT by my management. I also prefer to develop software using test-driven development. If possible, I'd rather avoid using third-party test libraries that pollute the global Fortran namespace with generic names.

To give a concrete example, if I were writing a UQ application in Fortran and wanted to run simulations at a number of parameter values, it might be natural for me to queue successful simulations for post-processing using a procedure named `add_success`, which would then clash with FRUIT's `add_success` subroutine.

To avoid such name clashes with library client applications and with other libraries, a best practice for libraries written in languages such as C and Fortran with limited namespacing facilities is to prefix API calls with a short, common string. Using such a motif also makes more clear the provenance of these calls.

See, for instance:

* http://fortranwiki.org/fortran/show/Source+conventions
* https://ejrh.wordpress.com/2012/01/24/namespaces-in-c/
* http://davidz25.blogspot.com/2011/06/writing-c-library-part-3.html
* https://lucumr.pocoo.org/2013/8/18/beautiful-native-libraries/

An alternative approach to this PR would be to write a "wrapper module" that aliases the current FRUIT API to prefixed names. This approach would be less intrusive, but the wrapper module could only be used by client code using Fortran 2003 or later.

Of course, one could always try to force all other code to avoid using names that clash with FRUIT, but doing so may not be practical because the offending code may be code out of the control of the author, such as another library distributed only in binary form.

> My main concern is the maintenance required if we ever upgraded again and the fact this now breaks googling for help with FRUIT. I do recognize that FRUIT is quite stable.

re: maintenance, I agree, which is why I automated the changes in a script written to try to make these changes obvious, viz.:

```bash
#!/usr/bin/env bash

replace()
{
    find . -iname "fruit_f90_source.txt" -exec sed -i '' s/$1/$2/g {} +
    for suffix in "rb" "f90" "f95" "f03" "f08"; do
        find . -iname "*.${suffix}" -exec sed -i '' s/$1/$2/g {} +
    done
}

REPO_ROOT=$(git rev-parse --show-toplevel)

pushd ${REPO_ROOT}
replace init_fruit fruit_init
replace get_last_message fruit_get_last_message
replace is_last_passed fruit_is_last_passed
replace is_case_passed fruit_is_case_passed
replace add_success fruit_add_success
replace addSuccess fruit_addSuccess
replace set_unit_name fruit_set_unit_name
replace get_unit_name fruit_get_unit_name
replace set_case_name fruit_set_case_name
replace get_case_name fruit_get_case_name
replace failed_assert_action fruit_failed_assert_action
replace get_total_count fruit_get_total_count
replace getTotalCount fruit_getTotalCount
replace get_failed_count fruit_get_failed_count
replace getFailedCount fruit_getFailedCount
replace is_all_successful fruit_is_all_successful
replace isAllSuccessful fruit_isAllSuccessful
replace run_test_case fruit_run_test_case
replace runTestCase fruit_runTestCase
replace assert_equals fruit_assert_equals
replace assertEquals fruit_assertEquals
replace assert_not_equals fruit_assert_not_equals
replace assertNotEquals fruit_assertNotEquals
replace assert_true fruit_assert_true
replace assertTrue fruit_assertTrue
replace stash_test_suite fruit_stash_test_suite
replace restore_test_suite fruit_restore_test_suite
replace override_xml_work fruit_override_xml_work
replace end_override_xml_work fruit_end_override_xml_work
replace get_assert_and_case_count fruit_get_assert_and_case_count
replace initializeFruit fruit_initialize
replace getTestSummary fruit_getTestSummary
replace add_fail fruit_add_fail
replace addFail fruit_addFail
replace init_fruit_xml fruit_init_xml
replace case_passed_xml fruit_case_passed_xml
replace case_failed_xml fruit_case_failed_xml
replace override_stdout fruit_override_stdout
replace end_override_stdout fruit_end_override_stdout
replace override_xml_work fruit_override_xml_work
replace end_override_xml_work fruit_end_override_xml_work
replace get_xml_filename_work fruit_get_xml_filename_work
replace set_xml_filename_work fruit_set_xml_filename_work
replace get_message_index fruit_get_message_index
replace get_messages fruit_get_messages
replace get_message_array fruit_get_message_array
replace set_prefix fruit_set_prefix
replace get_prefix fruit_get_prefix
trap popd EXIT SIGINT SIGKILL
```

To further reduce the maintenance burden, I've also added a script to locally build and test FRUIT without installing anything at system scope, and another script that reverts all unstaged changes from the files changed by the replace script above.

I found Googling for FRUIT documentation unhelpful because most documentation links are dead. The mailing list was also unhelpful because it hasn't been used in 12 years. I had to look at the source code in the upstream tarball to understand how to use FRUIT. I would guess other users would take a similar approach.

I'd propose submitting these changes upstream, but the sparse project activity leads me to believe I won't get a response.

ltaylor16:
I hesitate using a fork of fruit since then we'd own it.  I would suggest trying to see about getting the changes upstream.
The name spacing issue in Fortran can be addressed by using `USE, ONLY` and rename as described in http://fortranwiki.org/fortran/show/Source+conventions to create a wrapper module. That would involve no changes to Fruit.
Fruit needs support for additional type like INT64. That should also be pushed upstream.

goxberry:
> I hesitate using a fork of fruit since then we'd own it. I would suggest trying to see about getting the changes upstream.

@ltaylor16 Agreed, but as I noted in my reply, I am not optimistic that changes would be accepted due to lack of activity on the Sourceforge site.

> The name spacing issue in Fortran can be addressed by using `USE, ONLY` and rename as described in http://fortranwiki.org/fortran/show/Source+conventions to create a wrapper module. That would involve no changes to Fruit.

I noted this alternative in my previous message in this thread. Instead of preferences, what I need is a decision from BLT maintainers about what they will accept.

If renaming FRUIT procedures is unacceptable, does the BLT team expect users to write their own wrapper modules, or would it make more sense for me to contribute a wrapper module to BLT?

If I contribute a wrapper module, should that be in a separate source file, or should it be appended to `fruit.f90`? I presume that if the BLT maintainers prefer this option that I'll have to write tests for the wrapper module.

> Fruit needs support for additional type like INT64. That should also be pushed upstream.

While I agree with you, I'm hesitant to make and to suggest many changes to upstream at once, particularly given the lack of activity on the upstream project. Increasing the number of proposed changes to upstream strengthens the argument for forking FRUIT, which you've said you'd prefer to avoid.

Adding more types looks doable, but is out of scope of the projects I currently work on. I would need funding to take on that task.

ltaylor16:
Add the rename wrapper module as a separate file that is in the same directory as fruit.f90.  Once that is there, I can add subroutines for additional types to that file in a later pull request. I am going to say that it doesn't need any tests.  Fruit itself does not have any tests, at least any that we're running.
Adding it as a separate module also makes it easier to convert to the newer usage since the current usage will be unchanged.
It sounds like we can say we've updated to 3.4.3 since fruit.f90 has not changed.

kennyweiss:
I agree with @ltaylor16  -- a wrapper file will be much easier for us to maintain than a fork of ``fruit``.

white238:
We should revisit this when we have a solution that doesn't cause such breaking changes to FRUIT.

PR #350, opened 2020-03-19
Feature/white238/clangformat

white238:
Adds ClangFormat support.

ClangFormat has a couple short-comings I had to work around and I documented.
 * No support for --dry-run until version 10, which doesn't seem to be many places yet.  I grabbed a widely used script online that helps with this
 * No support for a command line option to use a specific config file.  I copied it to the working directory during the cmake phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,17 +61,21 @@ Authors
 
 Developers include:
 
- * Chris White (white238@llnl.gov)
- * Kenneth Weiss (kweiss@llnl.gov)
- * Cyrus Harrison (harrison37@llnl.gov)
- * George Zagaris (zagaris2@llnl.gov)
- * Lee Taylor (taylor16@llnl.gov)
- * Aaron Black (black27@llnl.gov)
- * David A. Beckingsale (beckingsale1@llnl.gov)
- * Richard Hornung (hornung1@llnl.gov)
- * Randolph Settgast (settgast1@llnl.gov)
+ * Chris White, LLNL
+ * Kenneth Weiss, LLNL
+ * Cyrus Harrison, LLNL
+ * George Zagaris, LLNL
+ * Lee Taylor, LLNL
+ * Aaron Black, LLNL
+ * David A. Beckingsale, LLNL
+ * Richard Hornung, LLNL
+ * Randolph Settgast, LLNL
 
-Please see the [BLT Contributors Page](https://github.com/LLNL/BLT/graphs/contributors) for the full list of project contributors.
+Please see our [contributing guide](https://github.com/LLNL/blt/blob/develop/CONTRIBUTING.md) 
+for details about how to contribute to the project.
+
+The full list of project contributors can be found on the 
+[BLT Contributors Page](https://github.com/LLNL/BLT/graphs/contributors).
 
 Open-Source Projects using BLT
 ------------------------------
--- a/README.md
+++ b/README.md
@@ -29,13 +29,15 @@ operating systems and technologies:
       [FRUIT](https://sourceforge.net/projects/fortranxunit),
       [gbenchmark](https://github.com/google/benchmark)
  * Documentation:
-      [Doxygen](http://www.doxygen.nl/), 
+      [Doxygen](http://www.doxygen.nl/),
       [Sphinx](http://www.sphinx-doc.org)
- * Code style and health:
-      [Uncrustify](http://uncrustify.sourceforge.net), 
-      [AStyle](http://astyle.sourceforge.net), 
-      [Cppcheck](http://cppcheck.sourceforge.net),
-      [clang-query](http://clang.llvm.org/docs/LibASTMatchers.html)
+ * Code style:
+      [AStyle](http://astyle.sourceforge.net),
+      [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html),
+      [Uncrustify](http://uncrustify.sourceforge.net)
+ * Code quality
+      [clang-query](http://clang.llvm.org/docs/LibASTMatchers.html),
+      [Cppcheck](http://cppcheck.sourceforge.net)
  
 
 Getting started
@@ -132,3 +134,7 @@ PackageLicenseDeclared: BSD-3-Clause
 PackageName: gtest  
 PackageHomePage: https://github.com/google/googletest  
 PackageLicenseDeclared: BSD-3-Clause  
+
+PackageName: run-clang-format  
+PackageHomePage: https://github.com/Sarcasm/run-clang-format  
+PackageLicenseDeclared: MIT  
--- a/cmake/SetupCodeChecks.cmake
+++ b/cmake/SetupCodeChecks.cmake
@@ -9,16 +9,6 @@
 add_custom_target(${BLT_CODE_CHECK_TARGET_NAME})
 add_custom_target(${BLT_CODE_STYLE_TARGET_NAME})
 
-if(UNCRUSTIFY_FOUND)
-    # targets for verifying formatting
-    add_custom_target(uncrustify_check)
-    add_dependencies(${BLT_CODE_CHECK_TARGET_NAME} uncrustify_check)
-
-    # targets for modifying formatting
-    add_custom_target(uncrustify_style)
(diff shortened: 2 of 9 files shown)


tests/internal/src/Example.cpp
```diff
@@ -6,20 +6,14 @@
 #include "Example.hpp"
```
kennyweiss: Is the ``.clang-format`` file missing? I didn't see it.

cmake/SetupCodeChecks.cmake
```diff
@@ -412,15 +453,96 @@ macro(blt_add_astyle_target)
...
+        # ClangFormat does not support --dry-run until version 10 which isn't on many machines.
+        # For now, use run-clang-format for dry running purposes.
+        if(${arg_MODIFY_FILES})
+            add_custom_target(${arg_NAME}
+                    COMMAND  ${CLANGFORMAT_EXECUTABLE} ${arg_PREPEND_FLAGS}
+                        --style=file -i ${arg_SRC_FILES} ${arg_APPEND_FLAGS}
+                    WORKING_DIRECTORY ${_wd} 
+                    COMMENT "${arg_COMMENT}Running ClangFormat source code formatting checks.")
+        else()
+            set(_run_clangformat "${BLT_ROOT_DIR}/cmake/run-clang-format.py" --clang-format-executable ${CLANGFORMAT_EXECUTABLE})
```
kennyweiss: I think ``run-clang-format`` might have also been left out.

cmake/SetupCodeChecks.cmake
```diff
@@ -412,15 +453,96 @@ macro(blt_add_astyle_target)
...
+        set(_wd ${CMAKE_CURRENT_SOURCE_DIR})
+    endif()
+
+    set(_generate_target TRUE)
+
+    # Copy config file to given working directory since ClangFormat doesn't support pointing to one
+    configure_file(${arg_CFG_FILE} ${arg_WORKING_DIRECTORY}/.clang-format COPYONLY)
+
+    if(_generate_target)
+        # ClangFormat does not support --dry-run until version 10 which isn't on many machines.
```
kennyweiss: Looks like we are always using ``run-clang-format``, even if the clang version is ``10+``

cmake/SetupCodeChecks.cmake
```diff
@@ -412,15 +453,96 @@ macro(blt_add_astyle_target)
...
+        if(${arg_MODIFY_FILES})
+            add_custom_target(${arg_NAME}
+                    COMMAND  ${CLANGFORMAT_EXECUTABLE} ${arg_PREPEND_FLAGS}
+                        --style=file -i ${arg_SRC_FILES} ${arg_APPEND_FLAGS}
+                    WORKING_DIRECTORY ${_wd} 
+                    COMMENT "${arg_COMMENT}Running ClangFormat source code formatting checks.")
+        else()
+            set(_run_clangformat "${BLT_ROOT_DIR}/cmake/run-clang-format.py" --clang-format-executable ${CLANGFORMAT_EXECUTABLE})
+            add_custom_target(${arg_NAME}
+                    COMMAND ${_run_clangformat} -j1 ${arg_SRC_FILES}
```
kennyweiss: Is the ``j`` flag necessary?

In other cases, with ``make``-based generators, I'm pretty sure I was able to run
```
make -j16 format
```
and it would run the formatting in parallel.
(I could be wrong though).

README.md
```diff
@@ -29,13 +29,15 @@ operating systems and technologies:
...
+      [Doxygen](http://www.doxygen.nl/),
       [Sphinx](http://www.sphinx-doc.org)
- * Code style and health:
-      [Uncrustify](http://uncrustify.sourceforge.net), 
-      [AStyle](http://astyle.sourceforge.net), 
-      [Cppcheck](http://cppcheck.sourceforge.net),
-      [clang-query](http://clang.llvm.org/docs/LibASTMatchers.html)
+ * Code style:
+      [AStyle](http://astyle.sourceforge.net),
+      [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html),
```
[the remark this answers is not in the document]
white238: I was going what i found most popular via googling.  ClangFormat is in their documentation but i dont see a ClangQuery anywhere, in fact google autocorrects you.

docs/api/code_check.rst
```diff
@@ -226,29 +240,87 @@ MODIFY_FILES
...
+ClangFormat can be found `here <https://clang.llvm.org/docs/ClangFormat.html>`_.
+
+When MODIFY_FILES is set to TRUE, modifies the files in place and adds the created build
+target to the parent `style` build target.  Otherwise the files are not modified and the
+created target is added to the parent `check` build target. This target will notify you
+which files do not conform to your style guide.
+
+.. note::
+  ClangFormat does not support a command line option for config files.  To work around this,
+  we copy the given config file to the given working directory. We recommend using the build
```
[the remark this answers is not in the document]
white238: Done.

docs/api/code_check.rst
```diff
@@ -226,29 +240,87 @@ MODIFY_FILES
...
+  directory.
+
+.. note::
+  ClangFormat does not support a command line option for check (--dry-run) until version 10.
+  This version is not widely used or available at this time. To work around this, we use an 
+  included script called run-clang-format.py that does not use PREPEND_FLAGS or APPEND_FLAGS.
```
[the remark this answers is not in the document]
white238: Done.

tests/internal/CMakeLists.txt
```diff
@@ -327,15 +327,20 @@ if(ASTYLE_FOUND)
...
+      message(FATAL_ERROR "Could not find ClangFormat style file: ${_cfg_file}")
+  endif()
+
   blt_add_code_checks(
-      PREFIX          smoke_tests
-      SOURCES         ${smoke_tests_srcs}
-      ASTYLE_CFG_FILE ${CMAKE_CURRENT_SOURCE_DIR}/astyle.cfg )
+      PREFIX               smoke_tests
+      SOURCES              ${smoke_tests_srcs}
+      CLANGFORMAT_CFG_FILE ${_cfg_file} )
```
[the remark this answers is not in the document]
white238: This macro defaults to a bunch of things.  The purpose is to be the default case and they should only need to call the lower ones if this one doesnt work.  I will add a comment about working directory in the docs.

tests/smoke/blt_hcc_smoke.cpp
```diff
@@ -16,14 +16,12 @@
 #include <hc.hpp>
 #include <hc_printf.hpp>
 
-int main()
-{
+int main() {
+  // clang-format off
```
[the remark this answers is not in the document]
white238: reverted all code formatting and will attempt to use the settings you gave me yesterday to see if its better.

tests/internal/src/Example.cpp
white238: I was using the one given by googlebenchmark.  Couple things went wrong: moving includes and the lambda's were ugly.

cmake/SetupCodeChecks.cmake
white238: Fixed.

cmake/SetupCodeChecks.cmake
white238: I made a comment about it in the docs.  run-clang-format does a pretty nice job cleaning the output.

cmake/SetupCodeChecks.cmake
white238: It defaults the CPU +1 for the j.. I found that to be kinda scary.

tests/internal/CMakeLists.txt
```diff
@@ -266,9 +266,9 @@ if(ENABLE_CLANGQUERY)
 endif()
 
 #------------------------------------------------------------------------------
-# Format the testing code using AStyle
+# Format the testing code using ClangFormat
```
kennyweiss: Since this PR doesn't currently run ``clang-format`` and doesn't have a ``.clang-format`` file, should we revert this file back to the ``astyle`` version?

white238: well it technically does have a .clang-format file.. i was using the one included in google benchmark. But i will split the PR into two.

PR #351, opened 2020-03-19
Fix docs slightly

white238:
Fix formatting and add note to another macro.

### DIFF
--- a/docs/api/code_check.rst
+++ b/docs/api/code_check.rst
@@ -44,8 +44,9 @@ all code checks from the working directory `CMAKE_BINARY_DIR`.  If you need more
 functionality you will need to call the individual code check macros yourself.
 
 .. note::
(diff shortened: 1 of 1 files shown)


docs/api/code_check.rst
```diff
@@ -43,6 +43,11 @@ The purpose of this macro is to enable all code checks in the default manner.  I
 all code checks from the working directory `CMAKE_BINARY_DIR`.  If you need more specific
 functionality you will need to call the individual code check macros yourself.
 
+.. note::
+  We recommend guarding your code checks against someone directly including your CMake project
```
kennyweiss: Suggestion: 

How about something like: 
> For library projects that may be included as a subproject of another code via CMake's ``add_subproject()``, we recommend guarding "code check" targets against being included in other codes.   ...

PR #356, opened 2020-04-21
Fix duplicate MPI link flag

white238:
The link flags were duplicating if you enabled Fortran and in CMake 3.14.5.  Also there was clearly a typo that no one was affected by.  This is a pretty obscure bug because most of the time these flags are the same between languages but for some reason in 3.14.5, MPI_C_LINK_FLAGS was empty while the other two were filled and that caused me to add it twice.


white238:
Nice, Azure tested this branch before I had finished writing the description...

PR #359, opened 2020-04-27
Add NVIDIA SDK path to the CUDA link flags

gzagaris:
# Summary

This commit adds the path to the NVDIA SDK libraries to the
CMAKE_CUDA_LINK_FLAGS. Moreover, it ensures that these
link flags are specified to the registered CUDA library that
is used with BLT.

This allows applications to link additional libraries that are supplied
with the NVIDIA SDK by listing them in the list of dependencies,
e.g., nvToolsExt, cublas, etc. In addition, CMAKE_CUDA_LINK_FLAGS
may also be specified in a host-config or at the command line and
those will now be correctly propagated and associated with
BLT's registered CUDA library.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,7 @@ strategy:
       Compiler_ImageName: 'axom/tpls:gcc-7'
       C_COMPILER: '/usr/bin/gcc'
       CXX_COMPILER: '/usr/bin/g++'
+      CMAKE_BIN_DIR: '/home/axom/axom_tpls/gcc-7.3.0/cmake-3.9.6/bin'
       MPI_DIR: '/home/axom/axom_tpls/gcc-7.3.0/mpich-3.2.1'
       MPI_FLAGS: '-DENABLE_MPI=ON -DMPI_C_COMPILER=$(MPI_DIR)/bin/mpicc -DMPI_CXX_COMPILER=$(MPI_DIR)/bin/mpicxx -DMPIEXEC=$(MPI_DIR)/bin/mpiexec -DMPIEXEC_NUMPROC_FLAG=-n'
       CMAKE_FLAGS: '-DCMAKE_C_COMPILER=$(C_COMPILER) -DCMAKE_CXX_COMPILER=$(CXX_COMPILER) -DENABLE_GTEST_DEATH_TESTS=OFF $(MPI_FLAGS) -DENABLE_OPENMP=ON'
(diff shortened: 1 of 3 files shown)


gzagaris:
The `linux_gcc8` CI is failing because MPI is not found. Any ideas?

cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -7,6 +7,15 @@
 # Sanity Checks
 ################################
 
+# Ensure CUDA_TOOLKIT_ROOT_DIR is specified, since it is needed 
+# by the call to find_packag(CUDA) below.
```
kennyweiss, line 11: Typo: `find_packag` -> `find_package`

cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -7,6 +7,15 @@
...
+if (NOT DEFINED CUDA_TOOLKIT_ROOT_DIR)
+   message( FATAL_ERROR
+        "Please specifiy CUDA_TOOLKIT_ROOT_DIR to use CUDA" )
```
kennyweiss, line 14: Typo: `specifiy` -> `specify`

cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -7,6 +7,15 @@
 # Sanity Checks
 ################################
 
+# Ensure CUDA_TOOLKIT_ROOT_DIR is specified, since it is needed 
+# by the call to find_packag(CUDA) below.
+if (NOT DEFINED CUDA_TOOLKIT_ROOT_DIR)
```
kennyweiss, line 12: Is `CUDA_TOOLKIT_ROOT_DIR` strictly necessary for cuda support?

Or could/should it be optional, and if it is present, we add `${CUDA_TOOLKIT_ROOT_DIR}/lib[64]` ?

I.e. was this always required, and this PR is adding a new check, or is this a new requirement?

gzagaris: For `find_package()` the way it is currently used, it is required. 

We are also relying on this for `clang-cuda`. 

I believe we were always requiring this, but not explicitly stating it.

That said, I think all this is now deprecated and no longer required. However, we seem to be stuck in a hybrid state where we use both the deprecated stuff and new CUDA Language support available with CMake, which gets confusing.

gzagaris: I guess if it is not broken, don't fix it....But, I'll submit an issue on this and when the opportunity arises this can be revamped quite a bit...

cmake/thirdparty/SetupCUDA.cmake
gzagaris: Done - Resolve and amended previous commit.

cmake/thirdparty/SetupCUDA.cmake
gzagaris: Done - Resolved and amended previous commit.

cmake/thirdparty/SetupCUDA.cmake
gzagaris: See #360

gzagaris:
Ok, I am not sure what the issue is with `Linux_gcc8` CI. 

After chatting with @white238 and trying a few things, I am now getting the following error:
```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/home/axom/axom_tpls/gcc-8.1.0/cmake-3.9.6/bin/cmake\": stat /home/axom/axom_tpls/gcc-8.1.0/cmake-3.9.6/bin/cmake: no such file or directory": unknown.
time="2020-04-28T03:55:20Z" level=error msg="error waiting for container: context canceled"

##[error]Bash exited with code '127'.
```

So, it looks like Azure was not able to load the image? 

This also appears to be one of the most recently updated and largest images we have (see https://hub.docker.com/r/axom/tpls/tags)

Does Azure have a time limit on how long it takes to download an image?

azure-pipelines.yml
```diff
@@ -22,7 +23,8 @@ strategy:
       Compiler_ImageName: 'axom/tpls:gcc-8'
       C_COMPILER: '/usr/bin/gcc'
       CXX_COMPILER: '/usr/bin/g++'
-      MPI_DIR: '/home/axom/axom_tpls/gcc-8.1.0/mpich-3.2.1'
+      CMAKE_BIN_DIR: '/home/axom/axom_tpls/gcc-8.1.0/cmake-3.10.1/bin'
+      MPI_DIR: '/home/axom/axom_tpls/gcc-8.1.0/mpich-3.3.2'
```
kennyweiss, line 27: Thanks for tracking this down @gzagaris !

PR #368, opened 2020-06-29
Support rocm 3.5

### LINKED ISSUE #366: HIP 3.5.0 support
Hip code doesn't build with rocm 3.5.0 without updating FindHIP on ubuntu.
I installed rocm 3.5.0 on ubuntu and tried to build RAJA with it, but it wouldn't work until I updated FindHIP in BLT.
I am still having an issue linking the BLT hip smoke test but that looks like an internal linker error and everything except the smoke test seemed to link.

MrBurmark:
Update FindHIP cmake from rocm 3.5. Fixes #366.

I've only tested compilation and linking with rocm 3.5 on ubuntu 20.04.

The hip smoke test appears to cause an internal error in the linker.

### DIFF
--- a/cmake/thirdparty/FindHIP/run_hipcc.cmake
+++ b/cmake/thirdparty/FindHIP/run_hipcc.cmake
@@ -32,12 +32,16 @@ set(HIP_HOST_COMPILER "@HIP_HOST_COMPILER@") # path
 set(CMAKE_COMMAND "@CMAKE_COMMAND@") # path
 set(HIP_run_make2cmake "@HIP_run_make2cmake@") # path
 set(HCC_HOME "@HCC_HOME@") #path
+set(HIP_CLANG_PATH "@HIP_CLANG_PATH@") #path
+set(HIP_CLANG_PARALLEL_BUILD_COMPILE_OPTIONS "@HIP_CLANG_PARALLEL_BUILD_COMPILE_OPTIONS@")
 
 @HIP_HOST_FLAGS@
 @_HIP_HIPCC_FLAGS@
 @_HIP_HCC_FLAGS@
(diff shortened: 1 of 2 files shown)


white238:
Side question are you guys seeing the quoted HIP includes on the compile line? possibly only on newish CMake versions?

MrBurmark:
@white238 yes, the new FIndHIP from this PR shouldn't have that problem anymore.

cmake/thirdparty/FindHIP/run_hipcc.cmake
```diff
@@ -32,26 +32,40 @@ set(HIP_HOST_COMPILER "@HIP_HOST_COMPILER@") # path
...
 set(HCC_HOME "@HCC_HOME@") #path
+set(HIP_CLANG_PATH "@HIP_CLANG_PATH@") #path
+set(HIP_CLANG_PARALLEL_BUILD_COMPILE_OPTIONS "@HIP_CLANG_PARALLEL_BUILD_COMPILE_OPTIONS@")
 
 @HIP_HOST_FLAGS@
 @_HIP_HIPCC_FLAGS@
 @_HIP_HCC_FLAGS@
+@_HIP_CLANG_FLAGS@
 @_HIP_NVCC_FLAGS@
-set(HIP_HIPCC_INCLUDE_ARGS "@HIP_HIPCC_INCLUDE_ARGS@") # list (needs to be in quotes to handle spaces properly)
```
MrBurmark, line 40: I believe this line was the culprit for quoted includes.

white238: That would do it.  CMake now (helpfully though backwards breaking) expects a list.

PR #385, opened 2020-08-05
Add shell script snippet from spack to handle multiple Python versions

### LINKED ISSUE #384: run-clang-format.py doesn't support differently-named Python interpreters
An approach similar to [the one used by Spack](https://github.com/spack/spack/blob/develop/bin/spack#L9-L22) would provide compatibility with Python executables not named `python`, e.g. `python3`, `python2`, `python2.7`, etc

joshessman-llnl:
This code is mostly from Spack, which itself is mostly from the linked StackOverflow question.

Resolves #384

### DIFF
--- a/cmake/run-clang-format.py
+++ b/cmake/run-clang-format.py
@@ -1,4 +1,18 @@
-#!/usr/bin/env python
+#!/bin/sh
+
+# This file is bilingual. The following shell code finds our preferred python.
+# Following line is a shell no-op, and starts a multi-line Python comment.
+# See https://stackoverflow.com/a/47886254
+""":"
+# prefer python3, then python, then python2
+for cmd in python3 python python2; do
+   command -v > /dev/null $cmd && exec $cmd $0 "$@"
+done
+echo "==> Error: run-clang-format could not find a python interpreter!" >&2
+exit 1
+":"""
+# Line above is a shell no-op, and ends a python multi-line comment.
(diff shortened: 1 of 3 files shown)


cmake/run-clang-format.py
```diff
@@ -1,4 +1,18 @@
-#!/usr/bin/env python
+#!/bin/sh
+
+# This file is bilingual. The following shell code finds our preferred python.
+# Following line is a shell no-op, and starts a multi-line Python comment.
+# See https://stackoverflow.com/a/47886254
+""":"
+# prefer python3, then python, then python2
+for cmd in python3 python python2; do
```
kennyweiss, line 8: Cool!  Will this run the script (up to) 3 times? Or does it somehow stop after the first success?

kennyweiss: Also, it would be good to check if this (still) works on Windows.

(It's not clear if it used to though).

joshessman-llnl: If there's a success, the `exec` shell builtin will destroy/overwrite the `sh` process with the Python process.

Windows doesn't use shebangs so it should just treat the [entire block](https://github.com/LLNL/blt/pull/385/files#diff-e622cd37cc768918c8027879b8bcdf07R1-R15) as a comment and use whatever version of Python has a file association configured for `.py` files.

PR #389, opened 2020-08-17
Remove CUDA PIE flag for PGI

davidbeckingsale:


cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -121,6 +120,14 @@ else()
     set(CUDA_HOST_COMPILER ${CMAKE_C_COMPILER})
 endif()
 
+# Set PIE options to empty for PGI since it doesn't understand -fPIE This
+# option is set in the CUDA toolchain file so must be unset after
+# enable_language(CUDA)
```
white238, line 125: How would a user know to do this?  Can you instead remove the -fPIE from a flag from some variable instead?

davidbeckingsale: I couldn't find another way to do this. It's because the toolchain sets it that makes it hard to undo.

white238: We also call enable_language(CUDA) in this file... can this be handled without user intervention? I just have no idea how people are supposed to know to do this.

davidbeckingsale: I'm not sure I understand. This doesn't require user intervention - if they enable CUDA and are using PGI, this change ensures that option ends up empty.

white238: OH! i read that comment the wrong way.. i thought you were implying the user must unset this variable ....

davidbeckingsale:
I'll add a host-config in a bit

host-configs/llnl/blueos_3_ppc64le_ib_p9/pgi@20.4_nvcc.cmake
```diff
@@ -0,0 +1,23 @@
...
+# other BLT Project Developers. See the top-level COPYRIGHT file for details
+# 
+# SPDX-License-Identifier: (BSD-3-Clause)
+
+#------------------------------------------------------------------------------
+# Example host-config file for using PGI and CUDA
+#------------------------------------------------------------------------------
+set (CMAKE_CXX_COMPILER "/usr/tce/packages/pgi/pgi-20.4/bin/pgc++" CACHE PATH "")
+set (CMAKE_C_COMPILER "/usr/tce/packages/pgi/pgi-20.4/bin/pgcc" CACHE PATH "")
+
```
white238, line 11: 👍

PR #392, opened 2020-08-25
Add a hostconfig that uses CUDA + C++17

joshessman-llnl:
This hostconfig was generated for https://github.com/LLNL/serac using the updated BLT in #387 .


white238:
@joshessman-llnl  thanks for all the help today but we don't want to encourage using the use of mvapich on blueos.  I added a tweaked version of this to #393 . So I'm going to close this PR.

PR #394, opened 2020-09-03
blt_add_test: fix missing executable suffix

jschueller:
on win32 the test extension is missing

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -684,7 +684,7 @@ macro(blt_add_test)
     # Append the test_directory to the test argument, accounting for multi-config generators
     if(NOT CMAKE_CONFIGURATION_TYPES)
         if(NOT "${_test_directory}" STREQUAL "")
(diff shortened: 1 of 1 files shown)


white238:
Can you give me an example of how this was causing you a problem? We have windows testing and projects on windows that use this so it is a bit surprising this is broken.

jschueller:
I'm using the Makefile generator, so it goes in that IF instruction.
maybe you're using msvc generator so this is not tested ?

white238:
My concern is this will break more than it fixes.  COMMAND can have multiple items in it, for example command line parameters.  Just tacking on the executable suffix to the end will not work.

Can you handle this at the level of the blt_add_test macro call like the following?

```
blt_add_exectuable(NAME foo
                                 ...)

blt_add_test(NAME foo_test
                     COMMAND foo${CMAKE_EXECUTABLE_SUFFIX} --some-command-line-option)
```

white238:
We might be able to incorporate it in the logic above that splits the command and deals with the test executable

PR #395, opened 2020-09-03
Add support for clang-tidy static analysis

joshessman-llnl:
In addition to adding a new macro for creating a clang-tidy target, this PR pulls in the LLVM project Python wrapper script (same approach as was used for clang-format).

### DIFF
--- a/cmake/SetupCodeChecks.cmake
+++ b/cmake/SetupCodeChecks.cmake
@@ -55,11 +55,23 @@ if(CLANGQUERY_FOUND)
     add_dependencies(${BLT_CODE_CHECK_TARGET_NAME} clang_query_check)
 endif()
 
+if(CLANGTIDY_FOUND)
+    # note: interactive_clang_tidy_check 
+    # is for the use of code developers who
+    # want to check specific attributes of
+    # specific targets, and does not make 
+    # sense as a dependency of check
+    add_custom_target(clang_tidy_check)
+    add_custom_target(interactive_clang-tidy_check)
+    add_dependencies(${BLT_CODE_CHECK_TARGET_NAME} clang_tidy_check)
+endif()
+
 # Code check targets should only be run on demand
 foreach(target 
         check uncrustify_check astyle_check clangformat_check cppcheck_check
         style uncrustify_style astyle_style clangformat_style
-        clang_query_check interactive_clang_query_check)
+        clang_query_check interactive_clang_query_check
+        clang_tidy_check interactive_clang_tidy_check)
     if(TARGET ${target})
         set_property(TARGET ${target} PROPERTY EXCLUDE_FROM_ALL TRUE)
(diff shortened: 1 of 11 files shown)


tests/internal/src/static_analysis/subtle_error_source.cpp
```diff
@@ -0,0 +1,25 @@
...
+ * See https://clang.llvm.org/extra/clang-tidy/checks/list.html
+ * for the full list of available checks
+ */
+
+#include <iostream>
+
+int main()
+{
+  int array[1] = {5};
+  int first_element = array[1];
```
joshessman-llnl, line 23: Interestingly enough, clang will issue a compiler warning by default, but GCC does not, even with `-Wall -Wextra -Wpedantic`.

cmake/SetupCodeChecks.cmake
```diff
@@ -279,6 +292,71 @@ macro(blt_add_clang_query_target)
...
 
+##-----------------------------------------------------------------------------
+## blt_add_clang_tidy_target( NAME              <Created Target Name>
+##                            WORKING_DIRECTORY <Working Directory>
+##                            COMMENT           <Additional Comment for Target Invocation>
+##                            CHECKS            <If specified, enables a specific set of checks>
+##                            FIX               <If true, apply fixes>
+##                            SRC_FILES         [FILE1 [FILE2 ...]] )
+##
+## Creates a new target with the given NAME for running clang=tidy over the given SRC_FILES
```
white238, line 303: clang-tidy

docs/api/code_check.rst
```diff
@@ -91,8 +91,12 @@ This macro supports the following static analysis tools with their requirements:
 
   * CLANGQUERY_EXECUTABLE is defined and found prior to calling this macro
 
+- clang-tidy
+
+  * CLANGTIDY_EXECUTABLE is defined and found prior to calling this macro
+
 These are added as children to the `check` build target and produce child build targets
-that follow the pattern `<PREFIX>_<cppcheck|clangquery>_check`.
+that follow the pattern `<PREFIX>_<cppcheck|clangquery|clang_tidy>_check`.
```
white238, line 99: Can you fix the wrong clangquery on this line... it should have been clang_query

white238 (approved):

kennyweiss (commented):
Thanks for the addition @joshessman-llnl -- Sorry for not getting to the review in time.

I had a quick question about how you intend users to use this new macro:
* Do you expect them to bake the actual clang-tidy `CHECKS` into their build system (i.e. `clang-analyzer-*`) as in the included `static_analysis` example?  
* Or do you envision allowing the user to specify a string for the clang-tidy checks and blt will help them apply it to its source files.

It seems to me that the latter would be more desirable and likely not too difficult.  Would it make sense to add this blt to demonstrate this use case?

PR #393, opened 2020-08-26
Various compatibility fixes and added tests against them

white238:
* Guard test that requires separable compilation
* Add new blueos c++17 host-config
* Remove BGQ host-config
* Guard system includes on PGI compilers
* Add ability to exclude CUDA implicit link dirs

I also reenabled (and hopefully fixed) our internal testing.  I am not sure who disabled it but it was on the wrong branch (main/master) and throwing alarming looking git errors.  I changed to a straight git repository.

PR #405, opened 2020-09-30
Add BLT_CMAKE_IMPLICIT_LINK_DIRECTORIES_EXCLUDE variable

davidbeckingsale:
This is used to filter the implicit link directories for all enabled languages

### DIFF
--- a/cmake/thirdparty/SetupCUDA.cmake
+++ b/cmake/thirdparty/SetupCUDA.cmake
@@ -159,13 +159,3 @@ blt_register_library(NAME          cuda
 blt_register_library(NAME      cuda_runtime
                      INCLUDES  ${CUDA_INCLUDE_DIRS}
                      LIBRARIES ${CUDA_LIBRARIES})
-
-# Allow the removal of implicitly found link directories
-# Note: This is a very specific fix for working around CMake adding implicit link
-# directories returned by the BlueOS compilers to link CUDA executables
-# They are added by logic found in the CMake source: Modules/CMakeTestCUDACompiler.cmake
-if(BLT_CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES_EXCLUDE)
-    message(STATUS "Removing implicit CUDA link directories: ${BLT_CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES_EXCLUDE}")
-    list(REMOVE_ITEM CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES ${BLT_CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES_EXCLUDE})
-    message(STATUS "Updated CUDA Implicit Link Directories: ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}")
-endif()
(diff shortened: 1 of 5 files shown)


cmake/thirdparty/SetupCUDA.cmake
```diff
@@ -158,14 +158,4 @@ blt_register_library(NAME          cuda
 # macros
 blt_register_library(NAME      cuda_runtime
                      INCLUDES  ${CUDA_INCLUDE_DIRS}
-                     LIBRARIES ${CUDA_LIBRARIES})
-
-# Allow the removal of implicitly found link directories
-# Note: This is a very specific fix for working around CMake adding implicit link
-# directories returned by the BlueOS compilers to link CUDA executables
-# They are added by logic found in the CMake source: Modules/CMakeTestCUDACompiler.cmake
-if(BLT_CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES_EXCLUDE)
```
cyrush, line 167: Folks may still be relying on the old version?

cyrush (commented):
More of this fun stuff huh? Overall looks good, I do wonder if we should still keep the old var.  Folks may have outstanding host configs with it in use, and if they update -- nothing will happen and they will run into a wall w/o warning?

cmake/thirdparty/SetupCUDA.cmake
kennyweiss: I agree -- it is subtle.  Maybe add an error message and fail the config?

kennyweiss (approved):
Thanks @davidbeckingsale 
Assuming this handles the original case and more, this looks good to me (but deferring to @white238 )

Please also update the `RELEASE-NOTES` to reflect this change.

cmake/thirdparty/SetupCUDA.cmake
white238: I'm not sure anyone actually used the old version.. old being very relative.  I used it in serac very briefly but then we got a new set of compilers that didn't rely on gcc 4.9.3.

kennyweiss: I think there was another usage of this variable.  See #393.

kennyweiss: Nevermind, this is what I was thinking about: 
https://github.com/LLNL/blt/blob/ae1b0c281fbca0dc3f6afab8d7e5f078f754c162/cmake/thirdparty/SetupCUDA.cmake#L115-L116

white238 (approved):

host-configs/llnl/blueos_3_ppc64le_ib_p9/clang@upstream_nvcc_c++17_no_separable.cmake
```diff
@@ -63,4 +63,4 @@ set(ENABLE_GTEST_DEATH_TESTS OFF CACHE BOOL "")
 
 # Very specific fix for working around CMake adding implicit link directories returned by the BlueOS
 # compilers to link CUDA executables 
-set(BLT_CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES_EXCLUDE "/usr/tce/packages/gcc/gcc-4.9.3/lib64/gcc/powerpc64le-unknown-linux-gnu/4.9.3;/usr/tce/packages/gcc/gcc-4.9.3/lib64" CACHE STRING "")
```
white238, line 66: https://github.com/LLNL/blt/blob/develop/host-configs/llnl/blueos_3_ppc64le_ib_p9/clang%40upstream_nvcc_c%2B%2B17.cmake

it is in this host-config too

davidbeckingsale: I'll fix that one.

davidbeckingsale:
Any other changes apart from release notes?

white238:
merge away sir! thanks for the fix

PR #410, opened 2020-10-14
Guard against clang-query turning itself on w/o explicit directories given

white238:
This was causing a problem where clang-query was being auto detected on the system but no checker directories were given.


kennyweiss (approved):
Thanks @white238 !

joshessman-llnl (approved):

PR #411, opened 2020-10-28
feat: check for multiple C format cfgs incl clang

goxberry:
This commit checks to see if a user has supplied more than one C-style
language formatter configuration file in blt_add_code_checks,
including CLANGFORMAT_CFG_FILE, which was previously excluded from
this check.

Prior to this commit, BLT returned a fatal error only if
ASTYLE_CFG_FILE and UNCRUSTIFY_CFG_FILE were supplied to
blt_add_code_checks. However, invocations supplying one of these
arguments plus CLANGFORMAT_CFG_FILE also seem undesirable, and the
documentation seems to imply that a check should be done to disallow
these combinations as well. Both of these observations motivate the
change in this commit.

I haven't yet added this change to the release notes because I wanted to make sure I wasn't missing some sort of use case for projects that intend to use `clang-format` plus `astyle` or `uncrustify`. Assuming the CMake changes proposed are reasonable, I'm happy to add this change to the release notes.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,6 +39,11 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   This caused a crash. blt_add_clang_query_target learned parameter CHECKER_DIRECTORIES
   and blt_add_code_checks learned parameter CLANGQUERY_CHECKER_DIRECTORIES.  Both still
   respect BLT_CLANGQUERY_CHECKER_DIRECTORIES.
+- It is now a fatal error to supply CLANGFORMAT_CFG_FILE plus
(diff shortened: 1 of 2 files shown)


kennyweiss (approved):
Thanks @goxberry. 

Please also update `RELEASE-NOTES.md`.

white238 (approved):
Thanks @goxberry

goxberry:
@kennyweiss Updated `RELEASE-NOTES.md`. Apologies for the slight trailing whitespace changes; I think that's due to my editor configuration. I can get rid of those changes if you'd like.

white238:
The trailing white-spaces are no problem.  Thanks for the contribution! I wil hit the merge button shortly.

PR #413, opened 2020-10-29
Loosen things that cause style version check to fail

white238:
* Guard against empty strings in version checks
* Change fatal-error to warning and disable style check when wrong version


cmake/SetupCodeChecks.cmake
```diff
@@ -509,9 +509,10 @@ macro(blt_add_astyle_target)
     if(BLT_REQUIRED_ASTYLE_VERSION)
         # The user may only specify a part of the version (e.g. just the maj ver)
         # so check for substring
-        string(FIND ${_astyle_version} ${BLT_REQUIRED_ASTYLE_VERSION} VERSION_POS)
+        string(FIND "${_astyle_version}" ${BLT_REQUIRED_ASTYLE_VERSION} VERSION_POS)
```
white238, line 512: This was an empty string which helpfully causes a CMake parsing error.

cmake/SetupCodeChecks.cmake
```diff
@@ -509,9 +509,10 @@ macro(blt_add_astyle_target)
...
         if (NOT VERSION_POS EQUAL 0)
-            message(FATAL_ERROR "blt_add_astyle_target: astyle ${BLT_REQUIRED_ASTYLE_VERSION} is required, found ${_astyle_version}")
+            set(_generate_target FALSE)
```
joshessman-llnl, line 514: 👍

joshessman-llnl (approved):

cmake/SetupCodeChecks.cmake
kennyweiss: Good catch! 

Does that mean that astyle was not emitting a version number? If so, was it an old version or a new version of astyle that was causing the issues?

kennyweiss (approved):
Thanks @white238 !

cmake/SetupCodeChecks.cmake
white238: This came up when Serac was built in WSL with an autodetected clang-format that returned an empty string (i guess).  I dont have access to the system.  So I'm guarding all future places this could happen. Then doing the chain of PRs so they can be unblocked.

kennyweiss: Got it -- thanks for the context.

white238:
This outputs too many warnings (one per child style target).  Fix incoming

goxberry:
Good catch! I also ran into this sort of error while testing the `yapf` formatter if `yapf` wasn't the right version or if it wasn't in the Python module search path.

PR #415, opened 2020-11-02
[feature/add-cmake-format-support] Add CMake beautifier support via cmake-format

### LINKED ISSUE #408: Add cmake-format checker/styler?
In the spirit of having `uncrustify`, `astyle` and `clang-format` linters, we should consider incorporating a CMake linter.
Perhaps: https://github.com/cheshirekow/cmake_format

goxberry:
Aims to resolve #408.

Has been tested with `cmake-format@0.6.13` on the files this PR adds to the test suite.


cmake/BLTPrivateMacros.cmake
```diff
@@ -589,6 +592,8 @@ macro(blt_split_source_list_by_language)
             message(FATAL_ERROR "blt_split_source_list_by_language given source file with no extension: ${_file}")
         endif()
 
+        get_filename_component(_name ${_file} NAME)
```
goxberry, line 595: Not sure if this syntax is worth flagging, but I noticed that `${_file}` isn't quoted in parts of `blt_split_source_by_language`, which gives me the impression this macro may not yield the desired results on paths that include spaces.

cmake/BLTPrivateMacros.cmake
```diff
@@ -599,8 +604,12 @@ macro(blt_split_source_list_by_language)
             if (DEFINED arg_Fortran_LIST)
                 list(APPEND ${arg_Fortran_LIST} ${_file})
             endif()
+        elseif(${_ext_lower} IN_LIST BLT_CMAKE_FILE_EXTS OR ${_name} STREQUAL "CMakeLists.txt")
```
goxberry, line 607: I elected to treat `CMakeLists.txt` here rather than place it in the list of extensions in `BLT_CMAKE_FILE_EXTS` because an exact filename match seemed like the simplest solution, and the extension-matching code seemed like a poor fit because it's case-insensitive.

tests/internal/cmake_format_style.py
```diff
@@ -0,0 +1,332 @@
+# Example Python-style cmake-format style file from cmake-format docs
```
goxberry, line 1: Possibly worth noting: this input file comes from `cmake-format@0.6.11`.

cmake/BLTPrivateMacros.cmake
white238: #422 added.  Will check later, thanks for the catch.

docs/api/code_check.rst
```diff
@@ -406,3 +418,52 @@ created target is added to the parent `check` build target. This target will not
...
   Setting MODIFY_FILES to FALSE is only supported in Uncrustify v0.61 or greater.
+
+
+blt_add_cmakeformat_target
+~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: cmake
+
+    blt_add_cmakeformat_target( NAME              <Created Target Name>
+                           MODIFY_FILES      [TRUE | FALSE (default)]
```
white238, line 429: nitpick: line these up please.

goxberry: Done in [00dc40c0](https://github.com/LLNL/blt/pull/415/commits/00dc40c0ae97d4b2c157fa633d13e5c58dcefa70).

white238 (approved):

tests/internal/src/test_cmake_format_conformant.cmake
```diff
@@ -0,0 +1,116 @@
+# Example from cmake-format documentation
```
kennyweiss, line 1: We should either add comments to attribute these files to their source, or develop our own test files instead.

kennyweiss (changes requested):
Thanks @goxberry !  Really excited to start using this (on blt and on other projects).

My only concern is about the test code, which seems like it is copied from the cmakeformat repo.  We should either attribute it properly according to their copyright or replace it with our own test code.

We'll develop a style file for BLT in a separate PR and apply it to BLT in a third.

Side question -- how does one get cmake format?  Is there a spack recipe for it?

cmake/BLTPrivateMacros.cmake
white238: Ah yes thanks for catching that @kennyweiss .  CMake parses strings first before evaluating if's.  We continually are bit by this.

If not quoted, empty strings do this and produce a very very hard to understand parsing error. The following:

```
if (${foo} STREQUAL "foo")
```

is evaluated as:

```
if ( STREQUAL "foo")
```

white238 (changes requested):
cmake-format is GPL 3.. Any source taken from them needs to be removed before it can be merged into our repository.  Also merged with a squash merge or removed from history.

cmake/BLTPrivateMacros.cmake
```diff
@@ -555,18 +555,19 @@ endmacro(blt_add_hip_executable)
...
+## and BLT_CMAKE_FILE_EXTS (global BLT variables). Files named
+## "CMakeLists.txt" are also filtered here. Files with no extension
+## or generator expressions that are not object libraries (of the form
 ## "$<TARGET_OBJECTS:nameofobjectlibrary>") will throw fatal errors.
 ## ------------------------------------------------------------------------------
 macro(blt_split_source_list_by_language)
 
     set(options)
-    set(singleValueArgs C_LIST Fortran_LIST Python_LIST)
+    set(singleValueArgs C_LIST Fortran_LIST Python_LIST CMAKE_LIST)
```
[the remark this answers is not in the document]
goxberry: @kennyweiss I thought about that possibility also. CMake has a lot of settings controlled by variables that start with the substring `CMAKE_`, so I chose a name consistent with that convention rather than one consistent with the convention established by `Fortran_LIST` and `Python_LIST`. 

CMake itself is inconsistent with respect to Python; compare the convention used in https://github.com/Kitware/CMake/blob/master/Modules/FindPythonInterp.cmake (Python is in all caps) to https://github.com/Kitware/CMake/blob/master/Modules/FindPython.cmake (Python is in the usual case).

goxberry:
* [390e6510](https://github.com/LLNL/blt/pull/415/commits/390e6510bb3ff17310558f4b9a42440a75138bf8) is a rebased commit that removes the files copied from the cmake-format documentation and replaces them as follows:
    - `tests/internal/cmake_format_style.py` contains the default settings for `py-cmake-format@0.6.13` generated via `cmake-format --dump-config python`
    - `tests/internal/src/test_cmake_format_nonconformant.cmake` is a copy of `cmake/SetupCodeChecks.cmake`
    - `tests/internal/src/test_cmake_format_conformant.cmake` is the result of running `cmake-format` on `tests/internal/src/cmake_format_nonconformant.cmake` using the style configuration in `tests/internal/cmake_format_style.py`

   The offending GPL'd files have been completely removed from the branch's history.
* [18ca7671](https://github.com/LLNL/blt/pull/415/commits/18ca7671ab3e7f61e9da31eec3fb06433916c70d) aims to fix the quoting issues noted by @kennyweiss in https://github.com/LLNL/blt/pull/415#discussion_r521857404 for the lines changed in `blt_split_source_list_by_language` in the process of adding logic for filtering CMake files into their own source list
* [7ccc22b6](https://github.com/LLNL/blt/pull/415/commits/7ccc22b6a5eaf128422af03a710b1cd8292c07ac) aims to fix the remaining quoting issues referenced by @white238 in #422; this commit could be moved into a separate pull request if desired (it's an easy cherry-pick)

white238:
The formatting it does is a bit interesting to say the least.  We will want to tweak that.  I think we need to add our macros to the `additional_commands` section at the beginning of `cmake_format_style.py`.

goxberry:
If you guys are anti-merge-commit for conflict resolution, let me know, and I can replace the merge with a rebase, squashing merge resolution changes into appropriate commits in the existing branch history.

white238 (approved):

white238:
Merging is fine. I was just contemplating the formatting it produces but I think we can clean that up in a follow up PR.

white238:
I ran into some problems trying to install cmake-format 0.6.13.  I'd like to tweak the format file before this goes in but maybe we should just merge this and fix it post merge

kennyweiss (approved):
Thanks for shepherding this through @goxberry !

PR #418, opened 2020-11-02
split HEADERS from SOURCES on blt_add_executable

white238:
This makes `blt_add_executable` learn `HEADERS` like `blt_add_library` already does.  This came up when in Axom's CUDA separable compilation build, where adding headers to the `SOURCES` list caused them to be compiled separately from the source file they were included on.

It is important to associate headers with an executable for dependency checking and IDE folder support.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -549,6 +549,7 @@ endmacro(blt_add_library)
 ##------------------------------------------------------------------------------
 ## blt_add_executable( NAME        <name>
 ##                     SOURCES     [source1 [source2 ...]]
+##                     HEADERS     [header1 [header2 ...]]
 ##                     INCLUDES    [dir1 [dir2 ...]]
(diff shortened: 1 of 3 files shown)


davidbeckingsale:
I understand that you don't want the files compiled as CUDA, but do we still want to set properties on headers that depend on CUDA? Only the source list gets passed out to `blt_setup_cuda_target`...

white238:
> I understand that you don't want the files compiled as CUDA, but do we still want to set properties on headers that depend on CUDA? Only the source list gets passed out to `blt_setup_cuda_target`...

This mimics what is already happening in `blt_add_library`, there are no properties you would set on a header in regards to CUDA (`LINKER_LANGUAGE`, `CUDA_SEPARABLE_COMPILATION`, or `LANGUAGE` to `CUDA`).

davidbeckingsale (approved):

kennyweiss (approved):

PR #420, opened 2020-11-03
[bugfix/replace-preprend-with-prepend-in-astyle-wrapper] Fix what might be bugs in the spelling & positions of astyle wrapper script arguments

goxberry:
This pull request makes two changes to fix what look like possible bugs in the astyle wrapper script:

* the substitution macro `@arg_PREPREND_FLAGS@` seems like it should be `@arg_PREPEND_FLAGS@`

* the line `set(ASTYLE_APPEND_FLAGS       @arg_PREPREND_FLAGS@)` seems like it should read `set(ASTYLE_APPEND_FLAGS       @arg_APPEND_FLAGS@)`


white238 (approved):
Good catch! I'm pretty sure we don't have any customers that use the specific flags part of this macro.  Only one I know of uses uncrustify.

PR #421, opened 2020-11-09
Add an (almost) drop-in replacement for blt_register_library that creates real CMake targets

joshessman-llnl:
This has been tested on axom (https://github.com/LLNL/axom/tree/feature/essman/blt_import) and serac (https://github.com/LLNL/serac/tree/feature/essman/blt_import).  

BLT's built-in third-party features have been switched to this new `blt_import_library` function with the exception of HCC.

The only context I have identified in which `blt_import_library` cannot be used as a drop-in replacement is with Google Test/Mock, as `blt_register_library` is used to create a "fake" (registered) library with the same name as a real CMake target.


cmake/BLTMacros.cmake
```diff
@@ -270,16 +270,25 @@ macro(blt_add_target_link_flags)
             endif()
         else()
             # In CMake <= 3.12, there is no target_link_flags or target_link_options command
-            get_target_property(_link_flags ${arg_TO} LINK_FLAGS)
-            if(NOT _link_flags)
-                set(_link_flags "")
+            get_target_property(target_type ${arg_TO} TYPE)
+            if (${target_type} STREQUAL "INTERFACE_LIBRARY")
+                # If it's an interface library, we add the flag via link_libraries
+                target_link_libraries(${arg_TO} INTERFACE ${_flags})
```
white238, line 276: This brings in all INTERFACE properties not just link flags.  I worry about this overreaching and grabbing more than intended.  Though I can't think of anything off hand a case where this would be a problem.  What was the cause of this change?

cmake/BLTPrivateMacros.cmake
```diff
@@ -433,6 +459,102 @@ macro(blt_setup_cuda_target)
...
+    if( arg_INCLUDES )
+        if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.11.0" )
+            target_include_directories(${arg_NAME} INTERFACE ${arg_INCLUDES})
+            # PGI does not support -isystem
+            if( (${arg_TREAT_INCLUDES_AS_SYSTEM}) AND (NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI"))
+                target_include_directories(${arg_NAME} SYSTEM INTERFACE ${arg_INCLUDES})
+            endif()
+        else()
+            # Interface include directories need to be set manually
+            SET_PROPERTY(TARGET ${arg_NAME}
```
white238, line 521: nitpick on style: we don't use all uppercase to macros/functions.

cmake/BLTMacros.cmake
joshessman-llnl: This conditional was added specifically for the `INTERFACE` libraries created by `blt_import_library` - per https://gitlab.kitware.com/cmake/cmake/-/issues/18606, this is how link flags can be propagated through interface libraries.

Interface libraries don't get built, so the only allowable/reasonable access specifier for `target_link_libraries` is `INTERFACE`.

white238: https://github.com/LLNL/blt/blob/106e47aeac960c4fac5c5876297b0bba722c6e2d/cmake/BLTMacros.cmake#L481

Header-only libraries also creates an `INTERFACE_LIBRARY`.  I just worry that this macro is only supposed to add link flags and is now adding all `INTERFACE` properties if _flags ends  up having a CMake target in it.

white238:
@rrsettgast this should solve a lot of your inheritance problems you were having.  Do you want @bmhan12 to try it out before this is merged?  It has been tested on Serac and Axom.

cmake/BLTMacros.cmake
joshessman-llnl: To my knowledge, this is the only way to add link options to an interface library pre-3.13 (I think use of this macro before this change would have failed for interface targets with the error message described in the CMake issue I linked).

I'm trying to think of a situation in which a link flag could also be a target, but link flags typically start with `-` or `/` which can't be in the name of a CMake target.  We could always tokenize the flags and look for existing target names, though.

white238: no its fine. lets not be crazy of a random thought i had.

cmake/BLTPrivateMacros.cmake
```diff
@@ -433,6 +459,99 @@ macro(blt_setup_cuda_target)
...
+    # TODO: This won't expand BLT-registered libraries
+    if( arg_DEPENDS_ON )
+        target_link_libraries(${arg_NAME} INTERFACE ${arg_DEPENDS_ON})
+    endif()
+
+    if( arg_INCLUDES )
+        if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.11.0" )
+            target_include_directories(${arg_NAME} INTERFACE ${arg_INCLUDES})
+            # PGI does not support -isystem
+            if( (${arg_TREAT_INCLUDES_AS_SYSTEM}) AND (NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI"))
```
white238, line 513: This use-case would be very helpful:

```
blt_patch_target(NAME axom TREAT_INCLUDES_AS_SYSTEM ON)
```

Where it would just copy the includes to the system includes variable.  We do this a lot in various projects. https://github.com/LLNL/axom/blob/f609592e34c5ce08facebc3c393e3aec68617ea9/src/cmake/thirdparty/SetupAxomThirdParty.cmake#L93

docs/api/target.rst
```diff
@@ -355,3 +355,36 @@ Internally created variables (NAME = "foo"):
...
+This behaves similarly to :ref:`blt_register_library`, but because it defines
+a native CMake target, no internal variable names are defined. This means that it
+is usable in contexts other than those that specifically accept BLT-registered
+libraries, and can take full advantage of CMake's transitive dependency resolution.
+
+As with BLT-registered libraries, it can be added to the DEPENDS_ON parameter
+when building another target or to ``target_link_libraries`` to transitively add in
+all includes, libraries, flags, and definitions associated with the imported library.
+
+See :ref:`blt_register_library` and :ref:`_blt_add_library` for a full description
```
white238, line 386: Please copy the parameters from the register library call.  I'd prefer people not have to pop back and forth between macros.

docs/tutorial/external_dependencies.rst
```diff
@@ -36,9 +36,15 @@ Then *axom* is available to be used in the DEPENDS_ON list in the following
 
 This is especially helpful for external libraries that are not built with CMake
 and don't provide CMake-friendly imported targets. Our ultimate goal is to use ``blt_register_library()`` 
-to import all external dependencies as first-class imported CMake targets to take full advanced of CMake's dependency lattice. 
+to import all external dependencies as first-class imported CMake targets to take full advantage of CMake's dependency lattice.
```
white238, line 39: This whole section should be reworked to use blt_import_library.  We should start discouraging the use of registered libraries except for use cases where we just cant get import libraries to work.

cmake/BLTPrivateMacros.cmake
joshessman-llnl: So in addition to exposing `blt_patch_target` as a public macro, you'd prefer to have TREAT_INCLUDES_AS_SYSTEM apply to all include directories and not just the ones specified by the INCLUDES parameter?

docs/tutorial/external_dependencies.rst
```diff
@@ -12,33 +12,42 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
+    blt_import_library(NAME      axom
+                       TREAT_INCLUDES_AS_SYSTEM ON
+                       DEFINES   HAVE_AXOM=1
+                       INCLUDES  ${AXOM_INCLUDES}
+                       LIBRARIES ${AXOM_LIBRARIES})
 
 Then *axom* is available to be used in the DEPENDS_ON list in the following
-``blt_add_executable()`` or ``blt_add_library()`` calls.
+``blt_add_executable()`` or ``blt_add_library()`` calls, or in any CMake command that accepts a target.
+CMake targets created by ``blt_import_library()`` are ``INTERFACE`` libraries that can be installed
```
kennyweiss, line 36: > CMake targets created by ``blt_import_library()`` are ``INTERFACE`` libraries that can be installed and exported

In that case, would it make sense for the new macro to be `blt_interface_library` instead of `blt_import_library`?
Specifically, are there different semantics in CMake for import libraries vs. interface libraries? And does the new macro better match one of these?

cmake/BLTPrivateMacros.cmake
white238: Private vs public: I would say yes, I think this is a very helpful macro that should be exposed and documented.

SYSTEM_INCLUDE: Yes, all customers I have talked to have wanted a target's includes to stop producing warnings want all of them to.

docs/tutorial/external_dependencies.rst
```diff
@@ -12,35 +12,38 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
 
-The ``blt_register_library()`` macro allows you to reuse all information needed
+The ``blt_import_library()`` macro allows you to reuse all information needed
 for an external dependency under a single name.  This includes any include
 directories, libraries, compile flags, link flags, defines, etc.  You can also
 hide any warnings created by their headers by setting the
 ``TREAT_INCLUDES_AS_SYSTEM`` argument.
 
-For example, to find and register the external dependency *axom* as a BLT registered library, you can simply use:
+For example, to find and register the external dependency *axom* as a CMake target, you can simply use:
```
white238, line 21: register -> create maybe import?

docs/tutorial/external_dependencies.rst
```diff
@@ -12,33 +12,42 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
 
-This is especially helpful for external libraries that are not built with CMake
-and don't provide CMake-friendly imported targets. Our ultimate goal is to use ``blt_register_library()`` 
-to import all external dependencies as first-class imported CMake targets to take full advanced of CMake's dependency lattice. 
+This is especially helpful for "converting" external libraries that are not built with CMake
+into CMake-friendly imported targets.
 
-MPI, CUDA, and OpenMP are all registered via ``blt_register_library()``. 
+This first-class importing functionality provided by ``blt_import_library()`` should be preferred, but ``blt_register_library()`` is
+still available for compatibility.  It is generally usable as a drop-in replacement, though it does not support the creation
```
kennyweiss, line 43: > __It__ is generally usable as a drop-in replacement...


`It` might be ambiguous here.  Consider replacing `it` with the explicit `blt_import_library`, or perhaps, `"the former"`

cmake/BLTPrivateMacros.cmake
davidbeckingsale: I agree re the system include stuff.

docs/tutorial/external_dependencies.rst
```diff
@@ -12,35 +12,38 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
+into CMake-friendly imported targets.
 
-This first-class importing functionality is provided by ``blt_import_library()``, but ``blt_register_library()`` is
-still available for compatibility.  It is generally usable as a drop-in replacement, though it does not currently support
-the FORTRAN_MODULES option.  Because CMake targets are only accessible from within the directory they were defined (including
+This first-class importing functionality provided by ``blt_import_library()`` should be preferred, but ``blt_register_library()`` is
+still available for compatibility.  It is generally usable as a drop-in replacement, though it does not support the creation
+of targets with the same name as a target that already exists.  
+
+Because CMake targets are only accessible from within the directory they were defined (including
```
white238, line 46: I would make this a "note" so it stands out.

docs/tutorial/external_dependencies.rst
```diff
@@ -12,33 +12,42 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
-to import all external dependencies as first-class imported CMake targets to take full advanced of CMake's dependency lattice. 
+This is especially helpful for "converting" external libraries that are not built with CMake
+into CMake-friendly imported targets.
 
-MPI, CUDA, and OpenMP are all registered via ``blt_register_library()``. 
+This first-class importing functionality provided by ``blt_import_library()`` should be preferred, but ``blt_register_library()`` is
+still available for compatibility.  It is generally usable as a drop-in replacement, though it does not support the creation
+of targets with the same name as a target that already exists.  
+
+Because CMake targets are only accessible from within the directory they were defined (including
```
kennyweiss, line 46: Consider making this a markdown "Note" admonition

docs/tutorial/external_dependencies.rst
```diff
@@ -12,33 +12,42 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
+subdirectories), the ``include()`` command should be preferred to the ``add_subdirectory()`` command for adding CMake files
+that create imported library targets needed in other directories. 
+
+MPI, CUDA, and OpenMP are all registered via ``blt_import_library()``.
```
kennyweiss, line 50: Suggestion:

```diff
- MPI, CUDA, and OpenMP are all registered via ``blt_import_library()``. 
+ Blt's `mpi`, `cuda`, and `openmp` targets are all defined via ``blt_import_library()``. 
```

docs/tutorial/external_dependencies.rst
```diff
@@ -12,33 +12,42 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
+of targets with the same name as a target that already exists.  
+
+Because CMake targets are only accessible from within the directory they were defined (including
+subdirectories), the ``include()`` command should be preferred to the ``add_subdirectory()`` command for adding CMake files
+that create imported library targets needed in other directories. 
+
+MPI, CUDA, and OpenMP are all registered via ``blt_import_library()``. 
 You can see how in ``blt/thirdparty_builtin/CMakelists.txt``.
 
 BLT also supports using ``blt_register_library()`` to provide additional options for existing CMake targets.
```
kennyweiss, line 53: > BLT also supports using ``blt_register_library()`` to provide additional options for existing CMake targets. 

It might be clearer if this line was placed above "This first-class importing functionality..."   ( a few lines up)

I might be missing something, but with the rewrite, I don't think we otherwise introduce `blt_register_library` anywhere else.

docs/api/target.rst
```diff
@@ -355,3 +355,62 @@ Internally created variables (NAME = "foo"):
...
+                        DEPENDS_ON               [dep1 [dep2 ...]]
+                        INCLUDES                 [include1 [include2 ...]]
+                        TREAT_INCLUDES_AS_SYSTEM [ON|OFF]
+                        FORTRAN_MODULES          [path1 [path2 ..]]
+                        LIBRARIES                [lib1 [lib2 ...]]
+                        COMPILE_FLAGS            [flag1 [flag2 ..]]
+                        LINK_FLAGS               [flag1 [flag2 ..]]
+                        DEFINES                  [def1 [def2 ...]] )
+
+Creates a CMake library target from a list of existing library files or targets.
```
kennyweiss, line 375: >  Creates a CMake library target from \*a list of existing library files or targets. \*

Please clarify the end of the above sentence.

The following might be a bit better, but is still not perfect:
> Creates a CMake target from build artifacts and system files generated outside of this build system.

docs/api/target.rst
```diff
@@ -355,3 +355,62 @@ Internally created variables (NAME = "foo"):
...
+COMPILE_FLAGS
+  List of compiler flags to be inherited by dependent targets
+
+LINK_FLAGS
+  List of linker flags to be inherited by dependent targets
+
+DEFINES
+  List of compiler defines to be inherited by dependent targets
+
+This behaves similarly to :ref:`blt_register_library`, but because it defines
```
kennyweiss, line 406: I think the following could benefit from a rewrite since it focuses more on implementation details than what it actually provides and when one would want to use it.

Since we want `blt_import_library` to be the new standard, we should not be defining it in terms of the old version.
Rather we should define what `blt_import_library` does and modify the `blt_register_library` writeup to indicate when one would need/prefer to use the latter.

docs/api/target.rst
```diff
@@ -355,3 +355,62 @@ Internally created variables (NAME = "foo"):
...
 
+.. _blt_import_library:
+
+blt_import_library
+~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: cmake
+
+    blt_import_library( NAME                     <libname>
+                        DEPENDS_ON               [dep1 [dep2 ...]]
```
kennyweiss, line 366: Are we planning to add a `GLOBAL` parameter to `blt_import_library`?

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -177,10 +177,9 @@ if (ENABLE_FORTRAN)
...
 endif()
 
-# Create the registered library
-blt_register_library(NAME          mpi
-                     INCLUDES      ${_mpi_includes}
-                     TREAT_INCLUDES_AS_SYSTEM ON
-                     LIBRARIES     ${_mpi_libraries}
-                     COMPILE_FLAGS ${_mpi_compile_flags}
-                     LINK_FLAGS    ${_mpi_link_flags} )
+blt_import_library(NAME          mpi
```
kennyweiss, line 180: 👍 -- it is actually just a drop-in replacement in all of these cases!

kennyweiss (commented):
Thanks @joshessman-llnl -- this is great! I'm really looking forward to using this.

Please update the RELEASE-NOTES and the documentation and then I think this is ready to go.

cmake/BLTMacros.cmake
```diff
@@ -380,6 +389,59 @@ macro(blt_register_library)
...
+
+    # Input checks
+    if( "${arg_NAME}" STREQUAL "" )
+        message(FATAL_ERROR "blt_import_library() must be called with argument NAME <name>")
+    endif()
+
+    # Add all imported targets to a single interface target
+    # Introducing OBJECT targets for each library file causes issues when exporting as 
+    # dependencies are not transitively included in exports
+    add_library(${arg_NAME} INTERFACE)
```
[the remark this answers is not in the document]
white238: The CMake error that is output is pretty self explanatory.

docs/tutorial/external_dependencies.rst
white238: I view "import" as a verb not an adjective but the word is pretty overloaded in CMake.  Maybe clarify the docs with "are CMake `INTERFACE` libraries"

docs/tutorial/external_dependencies.rst
white238: `cuda_runtime` as well then.

cmake/BLTMacros.cmake
```diff
@@ -270,16 +270,25 @@ macro(blt_add_target_link_flags)
             endif()
         else()
             # In CMake <= 3.12, there is no target_link_flags or target_link_options command
```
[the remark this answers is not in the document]
white238: BLT's various CI different CMake versions:

* axom/tpl docker container (with various compilers) for linux runs CMake 3.10.1
* Windows runs whatever the virtual machine provided by Azure and that is currently running CMake 3.18.4
* Bamboo runs 3.9.2

docs/tutorial/external_dependencies.rst
kennyweiss: Thanks @white238 -- to clarify, I'm happy with `blt_import_library` as the name as long as it is not unintentionally implying something CMake-related about the semantics of the generated targets.

docs/tutorial/external_dependencies.rst
```diff
@@ -12,39 +12,49 @@ To accomplish this BLT provides a ``DEPENDS_ON`` option for the
...
-MPI, CUDA, and OpenMP are all registered via ``blt_register_library()``. 
-You can see how in ``blt/thirdparty_builtin/CMakelists.txt``.
+This is especially helpful for "converting" external libraries that are not built with CMake
+into CMake-friendly imported targets.
 
 BLT also supports using ``blt_register_library()`` to provide additional options for existing CMake targets. 
 The implementation doesn't modify the properties of the existing targets, 
 it just exposes these options via BLT's support for  ``DEPENDS_ON``.
 
+This first-class importing functionality provided by ``blt_import_library()`` should be preferred, but ``blt_register_library()`` is
```
kennyweiss, line 46: 👍

RELEASE-NOTES.md
```diff
@@ -22,6 +22,9 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   ``BLT_REQUIRED_ASTYLE_VERSION``, ``BLT_REQUIRED_CLANGFORMAT_VERSION``, and ``BLT_REQUIRED_UNCRUSTIFY_VERSION``
 - Added ``HEADERS`` to ``blt_add_executable``.  This is important for build system dependency tracking
   and IDE folder support.
+- Added new ``blt_import_library`` macro that creates a real CMake target for imported libraries,
+  intended to be used instead of ``blt_register_library`` whenever possible
+- Added new ``blt_patch_target`` macro for modifying the properties of a CMake target in bulk
```
kennyweiss, line 27: For the second one "in bulk" seems a bit vague.

Perhaps something like the following would be clearer:
>Added new ``blt_patch_target`` macro to simplify modifying properties of an existing CMake target. This macro accounts for known differences in compilers and CMake releases.

kennyweiss (approved):
Thanks @joshessman-llnl !

RELEASE-NOTES.md
joshessman-llnl: I think it also makes sense to mention that it takes into account differences in target types as some operations are only valid on certain types of CMake targets.

docs/api/target.rst
joshessman-llnl: I was able to add this by adding the `IMPORTED` qualifier to the created libraries (in addition to `INTERFACE`).  In addition to allowing for `GLOBAL` visibility to be set, the created libraries are automatically wrapped into the libraries that depend on them, meaning that the end result is identical to the registered libraries approach.  This means that Axom is not required to namespace its exports as I previously thought.

white238: 👍  👍

cmake/BLTMacros.cmake
```diff
@@ -380,6 +400,175 @@ macro(blt_register_library)
...
+    endif()
+
+    if (NOT TARGET ${arg_NAME})
+        message(FATAL_ERROR "blt_patch_target() NAME argument must be a native CMake target")
+    endif()
+
+    # Default to public scope, unless it's an interface library
+    set(_scope PUBLIC)
+    get_target_property(_target_type ${arg_NAME} TYPE)
+    if(${_target_type} STREQUAL "INTERFACE_LIBRARY")
```
white238, line 443: Please change this to `"${_target_type}"`, I have been bit too many times by CMake's evaluation of if statements with strings.

white238 (approved):

PR #424, opened 2020-11-12
fix: add missing brace to yapf nonconformant test

goxberry:
This commit adds a missing brace to the yapf nonconformant style test
so that a CMake variable is interpolated correctly. The current
version lacks this brace, which is a syntax error.


goxberry:
@white238 @kennyweiss A heads up: caught this syntax error while updating #415.

white238 (approved):

PR #427, opened 2020-11-17
Fix dependency generation issue encountered with blt_import_library + HIP + gtest

joshessman-llnl:
@davidbeckingsale discovered an issue with hip+gtest that appeared when HIP was added via the new `blt_import_library`.

The generator expression in `FindHIP.cmake` intended to build the include-related compiler flags via the `INCLUDE_DIRECTORIES` property for a target fails (is empty) when HIP is added as a "real" CMake target.  An identical generator expression for the `INTERFACE_INCLUDE_DIRECTORIES` property was appended to "solve" the problem, though it is not obvious to me why method used to import the HIP library affects unrelated include directories in an executable target that depends on it.

To mitigate similar issues in the future, smoketests were added for cuda+gtest and hip+gtest.


davidbeckingsale:
CC @kab163

davidbeckingsale (approved):
Awesome, thanks for tracking this down so quickly @joshessman-llnl

white238 (approved):

PR #429, opened 2020-11-19
Refresh code coverage integration

joshessman-llnl:
This PR exposes the code coverage target macro to the user and allows source directories to be specified instead of using what I think is a hardcoded source directory.

It also allows for coverage analysis of failing tests, as it may be useful to evaluate test coverage even if not all tests pass.


docs/api/code_metric.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Code Metric Macros
+==================
+
+blt_add_code_coverage_target
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: cmake
+
+    blt_add_code_coverage_target( NAME              <Created Target Name>
```
white238, line 14: Nitpick: line up this up please.

docs/api/code_metric.rst
```diff
@@ -0,0 +1,43 @@
...
+Code coverage is the degree to which the tests for a piece of software "cover" functions
+and/or individual lines of code.  It can be used to identify gaps in testing, namely, code
+that is not tested.  GCC's ``gcov`` tool is used to generate the coverage data, and its accompanying
+``lcov`` tool is used to generate an HTML report containing coverage percentage information and
+highlighted source code that indicates which code was or was not executed as part of the test suite.
+
+.. note::
+  Coverage analysis is only supported by GNU/Clang compilers.
+
+This functionality requires that BLT's ``ENABLE_COVERAGE`` option is enabled and that ``gcov``, ``lcov``,
```
white238, line 38: You should mention here that GCOV_EXECUTABLE, LCOV_EXECUTABLE, and GENHTML_EXECUTABLE can be used to point at a specific version.

white238 (approved):

white238:
@cyrush I am pretty sure you use this macro and the macro signature is changing.

white238:
@joshessman-llnl please update the RELEASE-NOTES.md

cmake/thirdparty/SetupThirdParty.cmake
```diff
@@ -115,3 +115,21 @@ if(CMAKE_GENERATOR STREQUAL "Unix Makefiles" OR CMAKE_GENERATOR STREQUAL "Ninja"
...
+    set(ENABLE_GCOV ON)
+    set(ENABLE_LCOV ON)
+    set(ENABLE_GENHTML ON)
+    blt_find_executable(NAME        gcov
+                        EXECUTABLES gcov)
+
+    blt_find_executable(NAME        lcov
+                        EXECUTABLES lcov)
+
+    blt_find_executable(NAME        genhtml
```
kennyweiss, line 133: Side comment -- could we please update the documentation for `blt_find_executable` to reflect that it only attempts to find `arg_NAME` when `ENABLE_${arg_NAME}` is `TRUE` ?

https://github.com/LLNL/blt/blob/05ebb179779494f600d9031737888981d7b566ce/cmake/BLTPrivateMacros.cmake#L133-L144

(perhaps in a separate PR if you feel it is off topic for this one).

cmake/thirdparty/SetupThirdParty.cmake
```diff
@@ -115,3 +115,21 @@ if(CMAKE_GENERATOR STREQUAL "Unix Makefiles" OR CMAKE_GENERATOR STREQUAL "Ninja"
     blt_find_executable(NAME        ClangTidy
                         EXECUTABLES clang-tidy)
 endif()
+
+#------------------------------------
+# Code coverage
+#------------------------------------
+if (ENABLE_COVERAGE)
+    # find_executable requires that the executables be enabled
```
kennyweiss, line 123: For clarity, this comment should probably be something like:
> Attempt to find the executables associated with `gcov`, `lcov` and `genhtml`.
    This requires that the associate features are enabled.

cmake/thirdparty/SetupThirdParty.cmake
```diff
@@ -115,3 +115,21 @@ if(CMAKE_GENERATOR STREQUAL "Unix Makefiles" OR CMAKE_GENERATOR STREQUAL "Ninja"
...
+    set(ENABLE_GCOV ON)
```
kennyweiss, line 124: Could you please clarify how this line interacts with the corresponding cache variable `ENABLE_GCOV` if it is present?

My understanding is that it locally overrides that value (e.g. for the `blt_find_executable` below), but will not modify the CACHE variable's value, e.g. for use elsewhere in the code.

cmake/SetupCodeCoverageReports.cmake
```diff
@@ -45,47 +47,71 @@ endif()
...
-        COMMAND ${_testrunner} ${ARGV2}
+        # Run tests - allow for failing tests
+        COMMAND ${arg_RUNNER} || (exit 0)
 
         # Capture lcov counters and generating report
-        COMMAND ${LCOV_EXECUTABLE} --no-external --gcov-tool ${GCOV_EXECUTABLE} --directory ${CMAKE_BINARY_DIR} --directory ${CMAKE_SOURCE_DIR}/components --capture --output-file ${_targetname}.info
-        COMMAND ${LCOV_EXECUTABLE} --no-external --gcov-tool ${GCOV_EXECUTABLE} --directory ${CMAKE_BINARY_DIR} --directory ${CMAKE_SOURCE_DIR}/components --remove ${_targetname}.info '/usr/include/*' --output-file ${_targetname}.info.cleaned
-        COMMAND ${GENHTML_EXECUTABLE} -o ${_targetname} ${_targetname}.info.cleaned
-        COMMAND ${CMAKE_COMMAND} -E remove ${_targetname}.info ${_targetname}.info.cleaned
+        COMMAND ${LCOV_EXECUTABLE} --no-external --gcov-tool ${GCOV_EXECUTABLE} ${_coverage_directories} --capture --output-file ${arg_NAME}.info
```
kennyweiss, line 96: Have you tested that this works with more than one directory?

kennyweiss (approved):
Thanks @joshessman-llnl  -- at a high level, this looks good.

Please also update the `RELEASE-NOTES`.

cmake/thirdparty/SetupThirdParty.cmake
joshessman-llnl: Right now these are only needed for compatibility with `blt_find_executable` as discussed above (and not used anywhere else), but I think it's reasonable to set/use the cached variable instead - if this coverage functionality is extended further or if the user just wants to disable a particular tool it might be unwise to unconditionally set the "local" variable.

cmake/SetupCodeCoverageReports.cmake
joshessman-llnl: Yes, the "default" is two - from my understanding it expects at least the binary directory and a source directory, so there should always be at least two.  CMake automatically expands the semicolon-delimited list into space-separated arguments in this case.

white238 (approved):

cmake/thirdparty/SetupThirdParty.cmake
```diff
@@ -119,3 +119,22 @@ if(CMAKE_GENERATOR STREQUAL "Unix Makefiles" OR CMAKE_GENERATOR STREQUAL "Ninja"
     blt_find_executable(NAME        ClangTidy
                         EXECUTABLES clang-tidy)
 endif()
+
+#------------------------------------
+# Code coverage
+#------------------------------------
+if (ENABLE_COVERAGE)
+    # Attempt to find the executables associated with gcov, lcov and genhtml.
+    # This requires that the associate features are enabled.
```
kennyweiss, line 128: Typo: 
>associate -> associated

kennyweiss (approved):
Thanks for the updates!

PR #431, opened 2020-11-23
Remove nonexistent includes/libraries from "hip" target

joshessman-llnl:
After some further investigation of the issue that motivated #427, it appears that the underlying issue was related to a nonexistent/empty variable being used to configure the include directories for the `hip` target.  This resulted in the include directories for a HIP-dependent target comparing false and thus "poisoning" the generator expression used by `hip_add_library`/`hip_add_executable`.  

This did not appear until `hip` became a real CMake target in #421 due to the different setup logic used for BLT-registered libraries.
 
Removing these empty variables and adding a check in ``blt_patch_target`` appears to be sufficient and allows the patch to `FindHIP` to be removed.

Also, I noticed that BLT does not distinguish between `hip` and `hip_runtime` in that it appears to use `hipcc` to compile targets depending on either one (unlike with `cuda` vs `cuda_runtime`): https://github.com/LLNL/blt/blob/995a058193b4ae55daf5d34f0ede1ee1f2d26668/cmake/BLTPrivateMacros.cmake#L558 versus https://github.com/LLNL/blt/blob/995a058193b4ae55daf5d34f0ede1ee1f2d26668/cmake/BLTPrivateMacros.cmake#L430

Is this intentional?

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -54,6 +54,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   blt_add_code_checks; previously, these combinations were implied to
   be errors in BLT documentation, but BLT would not return an error in
   those cases.
+- ``blt_patch_target`` no longer attempts to set system include directories when a target
+  has no include directories
 
 ## [Version 0.3.6] - Release date 2020-07-27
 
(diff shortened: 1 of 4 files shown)


davidbeckingsale:
@joshessman-llnl - I think the important thing is that anything that depends on `hip` or `hip_runtime` must be linked as hip, so must be passed into `hip_add_executable` or `hip_add_library`.

We could potentially guard this check:
https://github.com/LLNL/blt/blob/995a058193b4ae55daf5d34f0ede1ee1f2d26668/cmake/BLTPrivateMacros.cmake#L568
with `if (${_depends_on_hip})`?

davidbeckingsale (approved):
Awesome catch!

white238 (approved):

PR #433, opened 2020-11-23
Add support for header-only library dependencies

joshessman-llnl:
Currently header-only libraries cannot specify dependencies with `DEPENDS_ON` in `blt_add_library` because `blt_setup_target` "hardcodes" `PUBLIC`/`PRIVATE` access specifiers.  This PR forces the use of `INTERFACE` access specifiers in `blt_setup_target` when the target is an `INTERFACE_LIBRARY`.

### DIFF
--- a/tests/internal/src/HeaderOnly.hpp
+++ b/tests/internal/src/HeaderOnly.hpp
@@ -6,12 +6,19 @@
 #ifndef BLT_HEADER_ONLY_HPP
 #define BLT_HEADER_ONLY_HPP
 
+#include "Example.hpp"
+
+#ifndef BLT_EXAMPLE_LIB
+  #error Compile definitions were not propagated from "example" library
+#endif
+
 namespace blt
 {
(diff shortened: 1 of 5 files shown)


tests/internal/src/HeaderOnly.hpp
```diff
@@ -6,12 +6,19 @@
 #ifndef BLT_HEADER_ONLY_HPP
 #define BLT_HEADER_ONLY_HPP
 
+#include "Example.hpp"
+
+#ifndef BLT_EXAMPLE_LIB
+  #error Compile definitions were not propagated from "example" library
```
white238, line 12: 👍

white238 (approved):

davidbeckingsale (approved):
I love this!

kennyweiss (approved):
Thanks @joshessman-llnl -- this is something we've wanted for a while!

PR #434, opened 2020-11-23
fix for umpire hip-enabled build

dtaller:
This is a fix for the hip-enabled umpire build. We ran into an error

[ 23%] Building CXX object src/tpl/umpire/src/umpire/resource/CMakeFiles/umpire_resource.dir/HipDeviceResourceFactory.cpp.o
In file included from /g/g17/taller1/CARE_SPACK/CHAI/src/tpl/umpire/src/umpire/resource/HipDeviceResourceFactory.cpp:9:
In file included from /opt/rocm/hip/include/hip/hip_runtime.h:64:
In file included from /opt/rocm/hip/include/hip/hip_runtime_api.h:366:
/opt/rocm/hip/include/hip/hcc_detail/hip_runtime_api.h:48:10: fatal error: 'hsa/hsa.h' file not found
#include <hsa/hsa.h>
         ^~~~~~~~~~~

When trying to build umpire (NOT using spack, by hand) .

This fix was inspired by David Beckinsale's work here: https://github.com/LLNL/Umpire/blob/develop/scripts/uberenv/packages/umpire/package.py#L326 , where he sets HIP_RUNTIME_INCLUDE_DIRS . Thank you @davidbeckingsale  for your help and advice!

### DIFF
--- a/cmake/thirdparty/SetupHIP.cmake
+++ b/cmake/thirdparty/SetupHIP.cmake
@@ -25,7 +25,7 @@ endif()
 #        2. The header hsa.h is found redundantly in both /opt/rocm/include/hsa/hsa.h
 #           and /opt/rocm/hsa/include/hsa/hsa.h . I only listed ${HIP_ROOT_DIR}/../include
 #           (pointing us to /opt/rocm/include/hsa/) below.
-if ( IS_DIRECTORY "${HIP_ROOT_DIR}/hcc/include" ) # this path only exists on older rocm installs
+if ( IS_DIRECTORY "${HIP_ROOT_DIR}/hcc/include" )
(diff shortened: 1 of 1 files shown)


davidbeckingsale:
@dtaller I don't think these specific tweaks should go in BLT - you should override somewhere else (like the Umpire spack package). I think that the specific include directories you need actually end up depending on the ROCM version too...

dtaller:
> @dtaller I don't think these specific tweaks should go in BLT - you should override somewhere else (like the Umpire spack package). I think that the specific include directories you need actually end up depending on the ROCM version too...

Ok, no worries, I'll kill this pull request and make the change elsewhere

dtaller:
> @dtaller I don't think these specific tweaks should go in BLT - you should override somewhere else (like the Umpire spack package). I think that the specific include directories you need actually end up depending on the ROCM version too...

Though actually I do not see this error with spack, only when I built by hand. The appropriate path must already have been included there

davidbeckingsale:
😄 it's probably handled by the rocm_prefix stuff you added!

white238:
This should be closed right?

PR #436, opened 2020-11-30
Add an EXPORTABLE option to blt_import_library

joshessman-llnl:
Since CMake does not provide a convenient way to "re-export" imported targets (see [this mailing list thread](https://cmake.org/pipermail/cmake/2019-July/069810.html) for example), this PR provides a more convenient way of exporting imported targets.

For example, if a project `proj` uses MPI via BLT's `mpi` target and wants to make its mpi target available to projects that import it, it can now configure BLT with `BLT_EXPORT_THIRDPARTY=ON`, which allows it to be exported like any other target:
```cmake
install(TARGETS     mpi
        EXPORT      proj-targets
        DESTINATION lib)
# Can also be namespaced if needed
set_target_properties(mpi PROPERTIES EXPORT_NAME proj::mpi)
```

Note that targets marked `EXPORTABLE` cannot also be `GLOBAL` (due to an underlying CMake limitation) and *must* be added to all export sets that include targets dependent on the `EXPORTABLE` target.

### DIFF
--- a/docs/tutorial/exporting_blt_targets.rst
+++ b/docs/tutorial/exporting_blt_targets.rst
@@ -3,7 +3,7 @@
 .. # 
 .. # SPDX-License-Identifier: (BSD-3-Clause)
 
-Exporting BLT-created Targets
+Exporting BLT Created Targets
 =============================
 
 BLT provides several built-in targets for commonly used libraries:
--- a/docs/tutorial/exporting_blt_targets.rst
+++ b/docs/tutorial/exporting_blt_targets.rst
@@ -1,0 +1,74 @@
+.. # Copyright (c) 2017-2019, Lawrence Livermore National Security, LLC and
+.. # other BLT Project Developers. See the top-level COPYRIGHT file for details
+.. # 
+.. # SPDX-License-Identifier: (BSD-3-Clause)
+
+Exporting BLT-created Targets
+=============================
+
+BLT provides several built-in targets for commonly used libraries:
+
+ * ``mpi`` (when ``ENABLE_MPI`` is on)
+ * ``openmp`` (when ``ENABLE_OPENMP`` is on)
+ * ``cuda`` and ``cuda_runtime`` (when ``ENABLE_CUDA`` is on)
+ * ``hip`` and ``hip_runtime`` (when ``ENABLE_HIP`` is on)
+
+These targets can be made exportable in order to make them available to users of
+your project via CMake's ``install()`` command.  Setting BLT's ``BLT_EXPORT_THIRDPARTY``
+option to ``ON`` will mark all active targets in the above list as ``EXPORTABLE``
+(see the ``blt_import_library()`` documentation for more info).
+
+.. note:: If a target in your project is added to an export set, any of its dependencies
+  marked ``EXPORTABLE`` must be added to the same export set.  Failure to add them will
+  result in a CMake error in the exporting project.
+
+Typical usage of the ``BLT_EXPORT_THIRDPARTY`` option is as follows:
+
+.. code-block:: cmake
+
+    # BLT configuration - enable MPI
+    set(ENABLE_MPI ON CACHE BOOL "")
+    # and mark the subsequently created MPI target as exportable
+    set(BLT_EXPORT_THIRDPARTY ON CACHE BOOL "")
+
+    # Later, a project might mark a target as dependent on MPI
+    blt_add_executable( NAME    example_1
+                        SOURCES example_1.cpp
+                        DEPENDS_ON mpi )
+
+    # Add the example_1 target to the example-targets export set
+    install(TARGETS example_1 EXPORT example-targets)
+
+    # Add the MPI target to the same export set - this is required
+    # because the mpi target was marked exportable
+    install(TARGETS mpi EXPORT example-targets)
+
+    # To avoid collisions with projects that import "example-targets",
+    # one option is to prefix the name of the MPI target
+    # With this approach the exported name of example_1 is unchanged
+    set_target_properties(mpi PROPERTIES EXPORT_NAME example::mpi)
+
+    # Alternatively, 
+
+To avoid collisions with projects that import "example-targets", there are
+two options for adjusting the exported name of the ``mpi`` target.
+
+The first is to rename only the ``mpi`` target's exported name:
+
+.. code-block:: cmake
+
+    set_target_properties(mpi PROPERTIES EXPORT_NAME example::mpi)
+    install(EXPORT example-targets)
+
+With this approach the ``example_1`` target's exported name is unchanged - a 
+project that imports the ``example-targets`` export set will have ``example_1``
+and ``example::mpi`` targets made available.
+
+Another approach is to install all targets in the export set behind a namespace:
+
+.. code-block:: cmake
+
+    install(EXPORT example-targets NAMESPACE example::)
+
+With this approach all targets in the export set are prefixed, so an importing
+project will have ``example::example_1`` and ``example::mpi`` targets made available.
--- a/docs/tutorial/exporting_blt_targets.rst
+++ b/docs/tutorial/exporting_blt_targets.rst
@@ -30,6 +30,7 @@ Typical usage of the ``BLT_EXPORT_THIRDPARTY`` option is as follows:
     set(ENABLE_MPI ON CACHE BOOL "")
     # and mark the subsequently created MPI target as exportable
     set(BLT_EXPORT_THIRDPARTY ON CACHE BOOL "")
+    # Both of the above must happen before SetupBLT.cmake is included
 
     # Later, a project might mark a target as dependent on MPI
     blt_add_executable( NAME    example_1
@@ -48,8 +49,6 @@ Typical usage of the ``BLT_EXPORT_THIRDPARTY`` option is as follows:
     # With this approach the exported name of example_1 is unchanged
     set_target_properties(mpi PROPERTIES EXPORT_NAME example::mpi)
 
-    # Alternatively, 
-
 To avoid collisions with projects that import "example-targets", there are
 two options for adjusting the exported name of the ``mpi`` target.
 
@@ -62,7 +61,8 @@ The first is to rename only the ``mpi`` target's exported name:
 
 With this approach the ``example_1`` target's exported name is unchanged - a 
 project that imports the ``example-targets`` export set will have ``example_1``
-and ``example::mpi`` targets made available.
+and ``example::mpi`` targets made available.  The imported ``example_1`` will
+depend on ``example::mpi``.
 
 Another approach is to install all targets in the export set behind a namespace:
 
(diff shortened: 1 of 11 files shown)


cyrush:
I am not understanding why folks would need to digest a TPL via a different name via this construct?

Can you show a use case that this helps support?

joshessman-llnl:
> 
> 
> I am not understanding why folks would need to digest a TPL via a different name via this construct?
> 
> Can you show a use case that this helps support?

The renamed target (in this case `proj::mpi`) would likely not be used by the importing project; it's mostly to avoid collisions.  

Without the renaming the exported `proj-targets.cmake` would have something like:
```cmake
add_library(proj STATIC IMPORTED)

set_target_properties(proj PROPERTIES
  INTERFACE_LINK_LIBRARIES "mpi"
)
```
This is fine if the importing project uses BLT (as BLT defines an `mpi` target), but is otherwise problematic as no `mpi` target would exist.  Exporting the `mpi` target solves this problem but introduces a collision for importing projects that use BLT, so my thinking was that it made sense to hide the dependency behind a namespace:
```cmake
add_library(proj::mpi INTERFACE IMPORTED)
# add libraries/includes to proj::mpi target

set_target_properties(proj PROPERTIES
  INTERFACE_LINK_LIBRARIES "proj::mpi"
)
```

With this approach exported targets can "keep track" of their dependencies without "polluting" the global namespace of projects that import them, similar to how BLT-registered libraries worked (though they just pasted include/library paths into target properties).  

Unlike BLT-registered libraries, this approach also allows for the "promoting" of imported targets.  For example, if a project that imports Axom with `AXOM_USE_MFEM` wishes to use MFEM independently of axom, they don't need a copy of `FindMFEM.cmake` - they can just
```cmake
add_library(mfem ALIAS axom::mfem)
```
and have access to a native CMake target for MFEM.

cyrush:
we have folks using our libs (that use blt) with mpi support in a code that doesn't use blt, and there are no issues with imported targets.

Mixing import logic (blt promoted as an import vs using find_package vs a FindZZZ logic) could be a source of lots of head scratching in the future.

kennyweiss (approved):
As far as I can tell, this PR does two explicit things and allows for a third implicit thing:

1. It adds an `EXPORTABLE` keyword to `blt_import_library`.  This adds support for imported libraries to be exported.
2. It adds a new option `BLT_EXPORT_THIRDPARTY` that marks blt's built-in TPLs as `EXPORTABLE`
3. This one is implicit in the sense that it allows users to EXPORT their imported libraries as described in the PR description.

(1) and (2) seem harmless enough since they are optional.
(3) seems like it could get somewhat tricky since BLT does not have built-in support for this (and I think that's what @cyrush is concerned about).  

I'm going to approve this since it seems like (a) it is helpful in @joshessman-llnl 's application and (b) one would have to go out of their way to get into trouble with this.

Perhaps some additional documentation about how this feature is meant to be used is in order?

kennyweiss:
Further question -- is there a concern about what might happen if multiple BLT-based projects that a project depends on define `BLT_EXPORT_THIRDPARTY=ON`, or if a project and its dependent define this?

Or is there only a concern if multiple projects mark their TPLs as `EXPORTABLE` **and** also export them in a way that clash?

white238 (approved):

cyrush:
@kennyweiss  I think it's possible to have a clash with a downstream code that isn't using BLT.

I don't want to hold this up, but I still have concerns.  

Pseudo Example:

Downstream code uses FindMPI (with or w/o BLT), and imports an Axom build that uses exportable-ed MPI -- b/c of the unique target names, that might end up linking MPI twice (or provide compiler the flags twice, or in a different order than might be easily understandable) The result could be a hard to reason about issue.

If instead, if import logic uses find package mechanics in a known way -- there won't be a chance for double link to happen.

We have been using BLT built libs in downstream codes that use CMake but not BLT for a while with no issues. 
The change to use blt_import for MPI seems to have kicked off a domino effect.

I don't see the downside of Axom calling FindMPI in its package import logic if that is needed to keep transitive working?

I know of no sim code using CMake that isn't calling FindMPI when they are using MPI. So the FindMPI in Axom's package import logic will basically be a no-op. Seems safer all around to me.  The same strategy will work for non-mpi deps as well. 
We have experience with HDF5 and MFEM being used this way.


Thinking about this, we might want to provide an option for `blt_import` for key deps -- there are wrinkles that need to be understood and ironed out.

white238:
> @kennyweiss I think it's possible to have a clash with a downstream code that isn't using BLT.
> 
> I don't want to hold this up, but I still have concerns.
> 
> Pseudo Example:
> 
> Downstream code uses FindMPI (with or w/o BLT), and imports an Axom build that uses exportable-ed MPI -- b/c of the unique target names, that might end up linking MPI twice (or provide compiler the flags twice, or in a different order than might be easily understandable) The result could be a hard to reason about issue.
> 
> If instead, if import logic uses find package mechanics in a known way -- there won't be a chance for double link to happen.
> 
> We have been using BLT built libs in downstream codes that use CMake but not BLT for a while with no issues.
> The change to use blt_import for MPI seems to have kicked off a domino effect.
> 
> I don't see the downside of Axom calling FindMPI in its package import logic if that is needed to keep transitive working?
> 
> I know of no sim code using CMake that isn't calling FindMPI when they are using MPI. So the FindMPI in Axom's package import logic will basically be a no-op. Seems safer all around to me. The same strategy will work for non-mpi deps as well.
> We have experience with HDF5 and MFEM being used this way.
> 
> Thinking about this, we might want to provide an option for `blt_import` for key deps -- there are wrinkles that need to be understood and ironed out.

The current system adds the mpi include/link information to the `axom` target that is exported in our `axom-targets.cmake`. So the double linking you are worried about is there already.

This simply puts it inside of a named `mpi` target instead of adding it directly to the `axom` target.
This allows either people to cleanly fix the exported mpi targets, alias the target, and/or use them directly.

joshessman-llnl:
> 
> 
> Further question -- is there a concern about what might happen if multiple BLT-based projects that a project depends on define `BLT_EXPORT_THIRDPARTY=ON`, or if a project and its dependent define this?
> 
> Or is there only a concern if multiple projects mark their TPLs as `EXPORTABLE` **and** also export them in a way that clash?

Multiple projects in a dependency tree can set `BLT_EXPORT_THIRDPARTY=ON`, but there will be a name conflict if an imported target fails to export those TPLs behind a namespace.  I've updated the documentation both in the tutorial and in the API docs to caution users of the EXPORTABLE option to namespace their exported imported targets.

kennyweiss:
@joshessman-llnl wrote:
> Multiple projects in a dependency tree can set `BLT_EXPORT_THIRDPARTY=ON`, *but there will be a name conflict if an imported target fails to export* those TPLs behind a namespace.

Just to make sure I understand this, the name conflict problem is only if:
(a) imported targets are exported
(b) *and* they are exported with conflicting target names.

What I'm trying to get at is if an action is required on the part of the library to export the `EXPORTABLE` targets (bullet (a)), or if this will happen automatically without an extra explicit `install` command.

joshessman-llnl:
> 
> Just to make sure I understand this, the name conflict problem is only if:
> (a) imported targets are exported
> (b) _and_ they are exported with conflicting target names.

Yes, that's correct.

> 
> What I'm trying to get at is if an action is required on the part of the library to export the `EXPORTABLE` targets (bullet (a)), or if this will happen automatically without an extra explicit `install` command.

The project would have to `install` the `EXPORTABLE` targets to introduce the possibility of a conflict.  If a target is marked `EXPORTABLE` and is a dependency of an already exported target, CMake will raise an error if the imported target is not explicitly installed.
The intended installation would look something like:
https://github.com/LLNL/axom/blob/02bc9fce34ddca555d741a35f0dd593119fd4765/src/cmake/thirdparty/SetupAxomThirdParty.cmake#L212-L216

If a target is not marked `EXPORTABLE`, the target must be found/defined by an importing project (e.g., potentially through the imported project's package configuration file).

kennyweiss:
Thanks @joshessman-llnl, quick follow-up:
> If a target is marked EXPORTABLE and is a dependency of an already exported target, **CMake** will raise an error...

Which CMake will raise the error -- the exporting repo or the importing repo ?

joshessman-llnl:
> 
> Which CMake will raise the error -- the exporting repo or the importing repo ?

The exporting project:
```cmake
blt_import_target(mpi ... EXPORTABLE ON)
blt_add_executable(calc_pi DEPENDS_ON mpi)
install(TARGETS calc_pi EXPORT calc_pi-targets)
install(EXPORT calc_pi-targets)
```

```
CMake Error: INSTALL (EXPORT "calc_pi-targets" ...) includes target "calc_pi" which requires target "mpi" that is not in the export set.
```

kennyweiss:
Thanks @joshessman-llnl.  

@cyrush -- does the above conversation help?  This is an opt-in feature that will error out if used improperly. The only downside that I can see is that users need to remember to add the project namespace to avoid clashes.

I'm probably missing some of the nuance though...

joshessman-llnl:
> The only downside that I can see is that users need to remember to add the project namespace to avoid clashes.

Another option I had considered was changing `EXPORTABLE` from a boolean option to something like `EXPORT_PREFIX`, where setting `EXPORT_PREFIX` to `calc_pi::` would add `calc_pi::` to the target's EXPORT_NAME property.  The issue is that this will add a second layer of prefixing to a project that exports its targets with the `NAMESPACE` option, e.g,, `calc_pi::calc_pi::mpi`.

cyrush:
the thing I am concerned about:  does the change to use `blt_import` now imply we have to do something different for mpi? That seems to be the genesis of his new feature? Maybe I am confused?

A feature to fix a change, worries me. 

Given that I haven't seen this issue and have knowledge of other codes using things just fine with the prior setup, what gives here?

docs/tutorial/exporting_blt_targets.rst
```diff
@@ -0,0 +1,74 @@
...
+  result in a CMake error in the exporting project.
+
+Typical usage of the ``BLT_EXPORT_THIRDPARTY`` option is as follows:
+
+.. code-block:: cmake
+
+    # BLT configuration - enable MPI
+    set(ENABLE_MPI ON CACHE BOOL "")
+    # and mark the subsequently created MPI target as exportable
+    set(BLT_EXPORT_THIRDPARTY ON CACHE BOOL "")
```
white238, line 32: Add a comment that (and/or show here) that these need to be set before SetupBLT.cmake is called

docs/tutorial/exporting_blt_targets.rst
```diff
@@ -0,0 +1,74 @@
...
+The first is to rename only the ``mpi`` target's exported name:
+
+.. code-block:: cmake
+
+    set_target_properties(mpi PROPERTIES EXPORT_NAME example::mpi)
+    install(EXPORT example-targets)
+
+With this approach the ``example_1`` target's exported name is unchanged - a 
+project that imports the ``example-targets`` export set will have ``example_1``
+and ``example::mpi`` targets made available.
```
white238, line 65: "made available and `example_1` automatically depends on `example::mpi` in the importing project" or something better worded.

cyrush (approved):
We talked through a few scenarios related to my concerns. The upshot was that I shouldn't worry so much.

Thanks for @joshessman-llnl + @white238 for listening and discussing.

white238:
https://llnl-blt.readthedocs.io/en/feature-exportable_imported_targets/

PR #438, opened 2020-12-03
Fix typo in blt_git_tag macro

gzagaris:
The check for whether or not the caller supplied
a SOURCE_DIR when calling the macro had a typo
that was preventing calling the macro with a
specified SOURCE_DIR. Fixed that.


white238 (approved):
Thanks @gzagaris !

cyrush (approved):
thanks @gzagaris !

PR #440, opened 2020-12-07
Bugfix: Disable SYSTEM flag for IMPORTED targets on PGI

### LINKED ISSUE #439: Failing LLNL blues_os PGI host-config
The following LLNL host-configs has been failing for the past few weeks:
https://github.com/LLNL/blt/blob/develop/host-configs/llnl/blueos_3_ppc64le_ib_p9/pgi%4020.4_nvcc.cmake

When attempting to build our internal tests
```bash
>cmake -C ../host-configs/llnl/blueos_3_ppc64le_ib_p9/pgi@20.4_nvcc.cmake ../tests/internal/
```

I got the following error messages:
```make
pgc++-Error-Unknown switch: -isystem
make[2]: *** [blt/tests/smoke/CMakeFiles/blt_cuda_smoke.dir/build.make:63: blt/tests/smoke/CMakeFiles/blt_cuda_smoke.dir/blt_cuda_smoke.cpp.o] Error 1
make[2]: *** [src/test_cuda_device_call_from_kernel/CMakeFiles/t_cuda_device_call_from_kernel_lib.dir/build.make:76: src/test_cuda_device_call_from_kernel/CMakeFiles/t_cuda_device_call_from_kernel_lib.dir/Parent.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:1616: blt/tests/smoke/CMakeFiles/blt_cuda_smoke.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

pgc++-Error-Unknown switch: -isystem
make[2]: *** [src/test_cuda_device_call_from_kernel/CMakeFiles/t_cuda_device_call_from_kernel_lib.dir/build.make:63: src/test_cuda_device_call_from_kernel/CMakeFiles/t_cuda_device_call_from_kernel_lib.dir/Child.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2046: src/test_cuda_device_call_from_kernel/CMakeFiles/t_cuda_device_call_from_kernel_lib.dir/all] Error 2

pgc++-Error-Unknown switch: -isystem
make[2]: *** [calc_pi/CMakeFiles/calc_pi_cuda.dir/build.make:63: calc_pi/CMakeFiles/calc_pi_cuda.dir/calc_pi_cuda.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2589: calc_pi/CMakeFiles/calc_pi_cuda.dir/all] Error 2

...

[ 28%] Linking CXX static library ../lib/libcalc_pi.a
[ 28%] Built target example
"/usr/tce/packages/cuda/cuda-10.1.243/include/crt/host_defines.h", line 54:
          warning: unrecognized preprocessing directive
  #warning "crt/host_defines.h is an internal header file and must not be used directly.  Please use cuda_runtime_api.h or cuda_runtime.h instead."
   ^
```

kennyweiss:
PGI does not support the `isystem` flag, but CMake's default for
imported targets is to mark include directories as SYTEM for imported targets.
As a workaround, we disable this feature for PGI.

Resolves #439

### DIFF
--- a/SetupBLT.cmake
+++ b/SetupBLT.cmake
@@ -101,14 +101,6 @@ if (NOT BLT_LOADED)
     ################################
     set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-
-    ################################
-    # PGI doesn't support the SYSTEM flag, which is the default for imported targets.
-    ################################
-    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
-        set(CMAKE_NO_SYSTEM_FROM_IMPORTED TRUE)
-    endif()
-
     ################################
     # Macros
     ################################
--- a/SetupBLT.cmake
+++ b/SetupBLT.cmake
@@ -101,6 +101,14 @@ if (NOT BLT_LOADED)
     ################################
     set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+
+    ################################
+    # PGI doesn't support the SYSTEM flag, which is the default for imported targets.
+    ################################
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
+        set(CMAKE_NO_SYSTEM_FROM_IMPORTED TRUE)
+    endif()
+
     ################################
     # Macros
     ################################
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -61,6 +61,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 - ``blt_patch_target`` no longer attempts to set system include directories when a target
   has no include directories
 - Header-only libraries now can have dependencies via DEPENDS_ON in ``blt_add_library``
+- Added a workaround for include directories of imported targets on PGI. CMake was 
+  erroneously marking them as SYSTEM but this is not supported by PGI.
 
 ## [Version 0.3.6] - Release date 2020-07-27
 
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -84,6 +84,14 @@ else()
 endif()
 
 
+#################################################################################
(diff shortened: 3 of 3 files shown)


white238 (approved):
well that is certainly a clean solution.  good find @kennyweiss

SetupBLT.cmake
```diff
@@ -101,6 +101,14 @@ if (NOT BLT_LOADED)
     ################################
     set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+
+    ################################
+    # PGI doesn't support the SYSTEM flag, which is the default for imported targets.
+    ################################
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
+        set(CMAKE_NO_SYSTEM_FROM_IMPORTED TRUE)
+    endif()
```
kennyweiss, line 110: This resolves the issue, but I'm not thrilled with the solution.

I'd like this fix to go inside `cmake/SetupCompilerOptions` since it is a specific fix for the PGI compiler.
However, the latter has a not that it has to be included after MPI is set up:
https://github.com/LLNL/blt/blob/113f4debe85e03f7f37b9ecd4bc0735da619dc2c/SetupBLT.cmake#L192-L196
which would seem be too late to affect our import targets.

Perhaps we should refactor the code so `SetupCompilerOptions.cmake` can be included before `SetupThirdParty.cmake` ?

white238: That could be a good idea but we do set some global properties in this file.

kennyweiss: Good news -- Looks like I'm able to move the workaround to `SetupCompilerOptions.cmake` without any refactoring. Will push an update momentarily.

joshessman-llnl (approved):
Good catch, thanks for looking into this!

davidbeckingsale (approved):

PR #441, opened 2020-12-21
Bugfix hip fortran

kab163:


davidbeckingsale (approved):

cmake/BLTMacros.cmake
```diff
@@ -796,7 +796,8 @@ macro(blt_add_executable)
     list(GET arg_SOURCES 0 _first)
     get_source_file_property(_lang ${_first} LANGUAGE)
     if(_lang STREQUAL Fortran)
-        if (NOT CUDA_LINK_WITH_NVCC)
+        #Check to see if HIP is enabled before resetting 
+        if (NOT (CUDA_LINK_WITH_NVCC OR ENABLE_HIP))
```
joshessman-llnl, line 800: I think since this is after `blt_add_hip_executable` we can be a little more precise by checking `HIP_SOURCE_PROPERTY_FORMAT` on the source file.

cmake/BLTMacros.cmake
```diff
@@ -796,7 +796,8 @@ macro(blt_add_executable)
     list(GET arg_SOURCES 0 _first)
     get_source_file_property(_lang ${_first} LANGUAGE)
     if(_lang STREQUAL Fortran)
-        if (NOT CUDA_LINK_WITH_NVCC)
+        #Check to see if HIP is enabled before resetting
```
joshessman-llnl, line 799: Suggestion/optional: something like "Don't reset the linker if the NVCC or HIP linker is required" might be more clear here

kab163: Done!

cmake/BLTMacros.cmake
kab163: Ok, let me fix this really quick. Thanks!

PR #442, opened 2020-12-21
Bugfix hip fortran

kab163:
Follow up to PR #441 to add a more precise way of checking the HIP linker


white238 (approved):
Thanks for the quick fix, @kab163 !

joshessman-llnl (approved):
Looks good!

PR #445, opened 2020-12-22
Append -dirty to Git hashcode when working tree is dirty

joshessman-llnl:
This replaces `rev-parse` with `describe`, which was introduced in Git 2.4.12.  Do we need to support versions of Git older than that?

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -41,6 +41,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   user-specified source directories
 - Code coverage targets leave LCOV-generated files intact for later use; these files will
   still be removed by ``make clean``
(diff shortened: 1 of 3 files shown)


kennyweiss:
@joshessman-llnl  wrote:
> This replaces rev-parse with describe, which was introduced in Git 2.4.12 (May 2017). 
Do we need to support versions of Git older than that?

The change here is small enough that it seems like we should probably check the git version and use the appropriate command.

kennyweiss (changes requested):
Seems like a nice change @joshessman-llnl, but since we're not requiring a specific version of git and this is a known issue that is easy to workaround, it seems to me that we should check the git version and use the appropriate command.

OTOH, if others feel that we do not need to support older git versions, then we should document the minimum supported version somewhere.

PR #446, opened 2020-12-22
[feature/add-fruit-mpi] Add FRUIT's MPI unit test reporting features to BLT

### LINKED ISSUE #444: Add FRUIT MPI extensions for MPI-aware Fortran unit tests?
Would there be any interest if I filed a PR adding the `fruit_mpi.f90` source file from FRUIT 3.4.1, plus a smoke test? I use FRUIT's MPI extensions in unit tests of a modal analysis package I'm developing for some of the lab's legacy codes. I'd like to update the build system so it uses BLT's copy of FRUIT, but I need those MPI extensions in order to run those tests.

goxberry:
Resolves #444.

This pull request adds another source file from FRUIT 3.4.1 -- `fruit_mpi.f90` -- that adds some procedures reporting unit test successes and failures over ranks in a given communicator. These additional procedures don't conflict with the API in `fruit.f90`.

### DIFF
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -81,6 +81,7 @@ endif()
 option(ENABLE_GTEST        "Enable Google Test testing support (if ENABLE_TESTS=ON)" ${_CXX_enabled})
 option(ENABLE_GMOCK        "Enable Google Mock testing support (if ENABLE_TESTS=ON)" OFF)
 option(ENABLE_FRUIT        "Enable Fruit testing support (if ENABLE_TESTS=ON and ENABLE_FORTRAN=ON)" ON)
+option(ENABLE_FRUIT_MPI    "Enable Fruit MPI testing support (if ENABLE_TESTS=ON and ENABLE_FORTRAN=ON and ENABLE_FRUIT=ON and ENABLE_MPI=ON" OFF)
 option(ENABLE_GBENCHMARK   "Enable Google Benchmark support (if ENABLE_TESTS=ON)" ${ENABLE_BENCHMARKS})
 
 
--- a/tests/smoke/CMakeLists.txt
+++ b/tests/smoke/CMakeLists.txt
@@ -65,6 +65,19 @@ if (ENABLE_FORTRAN AND ENABLE_FRUIT)
                  COMMAND blt_fruit_smoke)
 endif()
 
+###############
+# fruit mpi test
+###############
+if (ENABLE_FORTRAN AND ENABLE_FRUIT AND ENABLE_MPI AND ENABLE_FRUIT_MPI)
+    blt_add_executable(NAME blt_fruit_mpi_smoke
+                       SOURCES blt_fruit_mpi_smoke.f90
+                       OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
+                       DEPENDS_ON fruit fruit_mpi MPI::MPI_Fortran
+                       FOLDER blt/tests )
+
+    blt_add_test(NAME blt_fruit_mpi_smoke
+                 COMMAND blt_fruit_mpi_smoke)
+endif()
 
 ################
 # OpenMP test
--- a/tests/smoke/CMakeLists.txt
+++ b/tests/smoke/CMakeLists.txt
@@ -72,7 +72,7 @@ if (ENABLE_FORTRAN AND ENABLE_FRUIT AND ENABLE_MPI AND ENABLE_FRUIT_MPI)
     blt_add_executable(NAME blt_fruit_mpi_smoke
                        SOURCES blt_fruit_mpi_smoke.f90
                        OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
-                       DEPENDS_ON fruit fruit_mpi MPI::MPI_Fortran
+                       DEPENDS_ON fruit_mpi
                        FOLDER blt/tests )
 
     blt_add_test(NAME blt_fruit_mpi_smoke
--- a/tests/smoke/fortran_mpi_driver.cpp
+++ b/tests/smoke/fortran_mpi_driver.cpp
@@ -1,0 +1,23 @@
+// Copyright (c) 2017-2019, Lawrence Livermore National Security, LLC and
+// other BLT Project Developers. See the top-level COPYRIGHT file for details
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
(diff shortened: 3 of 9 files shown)


goxberry:
I suspect I'll need to clean up this PR -- I made some guesses during the implementation -- and can clean up the commit history during that process.

goxberry:
Tested this on an internal Fortran project and it seems to work as-is. Will be making some minor interactive rebase cleanups. Let me know if there are any cleanup- or style-related issues that need work. I added a smoke test to illustrate the main differences between `fruit` and `fruit_mpi` -- the main differences between `tests/smoke/blt_fruit_smoke.f90` and `tests/smoke/blt_fruit_mpi_smoke.f90` are:

* `fruit_summary_mpi` replaces `fruit_summary`
* `fruit_summary_mpi` assumes an `MPI_Comm` of `MPI_COMM_WORLD` and takes in `rank` and `size` arguments, so `mpi_init` must be called prior to this subroutine
* `fruit_finalize_mpi` replaces `fruit_finalize`, and `fruit_finalize_mpi` must be called before `mpi_finalize`

Admittedly, the assumption FRUIT makes that the `comm` of interest is `MPI_COMM_WORLD` isn't great; this assumption could be relaxed fairly easily, but would require forking the library. I don't think FRUIT is actively maintained.

tests/smoke/CMakeLists.txt
```diff
@@ -65,6 +65,19 @@ if (ENABLE_FORTRAN AND ENABLE_FRUIT)
...
 endif()
 
+###############
+# fruit mpi test
+###############
+if (ENABLE_FORTRAN AND ENABLE_FRUIT AND ENABLE_MPI AND ENABLE_FRUIT_MPI)
+    blt_add_executable(NAME blt_fruit_mpi_smoke
+                       SOURCES blt_fruit_mpi_smoke.f90
+                       OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
+                       DEPENDS_ON fruit fruit_mpi MPI::MPI_Fortran
```
white238, line 75: `fruit_mpi` should be the only thing required here

tests/smoke/fortran_mpi_driver.cpp
```diff
@@ -0,0 +1,23 @@
...
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+//----------------------------------------------------------------------
+
+/* 
+   fortran_mpi_test manages MPI_Init & MPI_Finalize. Do not put MPI
+   calls in this driver file -- put these calls in fortran_mpi_test,
+   which should be defined in blt_fruit_mpi_test.f90.
+ */
+extern "C" int fortran_mpi_test();
```
kennyweiss, line 13: [Likely outside the scope of this PR] Should the function / module be named `blt_fruit_mpi_test` for easier discoverability?
This would also imply changing `fortran_test` to `blt_fortran_test` in `fortran_driver.cpp` (and the related `blt_fruit_smoke.f90`.

kennyweiss (approved):
Thanks @goxberry

tests/smoke/fortran_mpi_driver.cpp
ltaylor16, line 13: I'm unclear where this file is used. It is not listed in CMakeLists.txt. Likewise for the existing fortran_driver.cpp.  The thing that caught my eye is the `extern "C" int fortran_mpi_test();`. There should be a corresponding `function fortran_mpi_test() bind(C)` in the fortran source.
I'm guessing this is left over from trying to fit fortran into gtest.

kennyweiss: I was wondering the same.

I saw that it is added to the list of cpp smoke tests here: https://github.com/LLNL/blt/pull/446/files#diff-652a16eeed4c19ea49e4b96c8e5c62f3143fece609868b90c23c6fd2343c90f2
(and we do the same for `tests/smoke/fortran_driver.cpp` ), but these might just be for formatting (?)

I don't see where they get compiled into an actual executable or tested.

Perhaps both of these are artifacts and should be removed?

goxberry: @ltaylor16 Agreed, and I was confused by the lack of `bind(C)` code in the function called by `tests/smoke/fortran_driver.cpp` as well.

@kennyweiss I didn't see where they were compiled into an actual executable, so the purpose of the Fortran C++ driver code was unclear to me, and I blindly pattern-matched when submitting the modifications in this pull request. If these files are artifacts, maybe it's better to delete them. If not, it's worth clarifying their purpose.

tests/smoke/CMakeLists.txt
goxberry: Done in [99322a95](99322a95fd471f4d729202d9fd4260758a8d0b44).

ltaylor16 (approved):

white238 (approved):

white238:
Thanks @goxberry ! if you update your branch I'll push the merge button.

cmake/BLTOptions.cmake
```diff
@@ -81,6 +81,7 @@ endif()
 option(ENABLE_GTEST        "Enable Google Test testing support (if ENABLE_TESTS=ON)" ${_CXX_enabled})
 option(ENABLE_GMOCK        "Enable Google Mock testing support (if ENABLE_TESTS=ON)" OFF)
 option(ENABLE_FRUIT        "Enable Fruit testing support (if ENABLE_TESTS=ON and ENABLE_FORTRAN=ON)" ON)
+option(ENABLE_FRUIT_MPI    "Enable Fruit MPI testing support (if ENABLE_TESTS=ON and ENABLE_FORTRAN=ON and ENABLE_FRUIT=ON and ENABLE_MPI=ON" OFF)
```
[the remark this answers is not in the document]
goxberry: It could be. I chose not to make it a dependent option for the time being because I saw that `ENABLE_GMOCK` was not a dependent option exposed when `ENABLE_GTEST` is `ON`.

tests/smoke/fortran_mpi_driver.cpp
goxberry: I think there are reasonable arguments for doing so in another PR, though obviously, this change would depend on whether the `fortran_driver.cpp` and `fortran_mpi_driver.cpp` files still exist.

goxberry:
@white238 Branch has been updated and it looks like all tests pass.

PR #447, opened 2021-01-04
Propagate HIP/CUDA linker requirements up the DAG

joshessman-llnl:
This PR introduces a `BLT_LINKER_LANGUAGE_OVERRIDE` target property used to convey the requirement that an executable depending on (directly or transitively) `cuda`/`hip` or their respective `runtime` targets be linked with the NVCC/HIP linker.

This property is copied from dependencies as part of `blt_setup` target to ensure correct behavior when the cuda/hip dependency is transitive.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -67,9 +67,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   erroneously marking them as SYSTEM but this is not supported by PGI.
 - Check added to make sure that if HIP is enabled with fortran, the LINKER LANGUAGE
   is not changed back to Fortran.
-- If an executable needs to be linked with the HIP or CUDA (NVCC) linker, this information
-  is internally propagated up the dependency graph and used to override the LINKER_LANGUAGE
-  target property for executable targets
+- Executables that link to libraries that depend on `hip`/`cuda`(`_runtime`) will automatically
+  be linked with the HIP or CUDA (NVCC) linker
 
 ## [Version 0.3.6] - Release date 2020-07-27
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -67,8 +67,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   erroneously marking them as SYSTEM but this is not supported by PGI.
 - Check added to make sure that if HIP is enabled with fortran, the LINKER LANGUAGE
   is not changed back to Fortran.
-- Executables that link to libraries that depend on `hip`/`cuda`(`_runtime`) will automatically
-  be linked with the HIP or CUDA (NVCC) linker
+- Executables that link to libraries that depend on `hip`/`hip_runtime`/`cuda`/`cuda_runtime`
+  will automatically be linked with the HIP or CUDA (NVCC) linker
 
 ## [Version 0.3.6] - Release date 2020-07-27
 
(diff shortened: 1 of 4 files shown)


kab163 (commented):
Looks good, thank you!!

kab163 (approved):

davidbeckingsale (approved):
Awesome! 👍

cmake/BLTPrivateMacros.cmake
```diff
@@ -389,6 +389,18 @@ macro(blt_setup_target)
             blt_add_target_link_flags(TO ${arg_NAME}
                                       FLAGS ${_BLT_${uppercase_dependency}_LINK_FLAGS} )
         endif()
+        # Propagate the overridden linker language, if applicable
+        if(TARGET ${dependency})
+            # If it's an interface library CMake doesn't even allow us to query the property
+            get_target_property(_dep_type ${dependency} TYPE)
+            if(NOT "${_dep_type}" STREQUAL "INTERFACE_LIBRARY")
```
white238, line 396: This is a complete shot in the dark.. but did you try INTERFACE_BLT_LINKER_LANGUAGE_OVERRIDE? maybe they only whitelist based on INTERFACE_?

white238 (approved):

RELEASE-NOTES.md
```diff
@@ -67,6 +67,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   erroneously marking them as SYSTEM but this is not supported by PGI.
 - Check added to make sure that if HIP is enabled with fortran, the LINKER LANGUAGE
   is not changed back to Fortran.
+- Executables that link to libraries that depend on `hip`/`cuda`(`_runtime`) will automatically
```
kennyweiss, line 70: Suggestion: Since there are only two cases, it might be clearer to say:
>`hip_runtime` or  `cuda_runtime`

cmake/BLTMacros.cmake
```diff
@@ -796,18 +796,23 @@ macro(blt_add_executable)
...
     blt_setup_target(NAME       ${arg_NAME}
                      DEPENDS_ON ${arg_DEPENDS_ON} 
                      OBJECT     FALSE)
 
+    # Override the linker language with BLT_LINKER_LANGUAGE_OVERRIDE, if applicable
+    # Will have just been populated by blt_setup_target
+    get_target_property(_blt_link_lang ${arg_NAME} BLT_LINKER_LANGUAGE_OVERRIDE)
+    if(_blt_link_lang)
+        # This is the final link (b/c executable), so override the actual LINKER_LANGUAGE
+        # The override will only ever be HIP or CUDA
```
kennyweiss, line 812: > \# The override will only ever be HIP or CUDA 

This seems like a promise that will be difficult to keep with a generic variable name like `BLT_LINKER_LANGUAGE_OVERRIDE`.  

I'd suggest modifying this comment to something like:
> \# BLT currently has overrides for HIP and CUDA

kennyweiss (approved):
Thanks @joshessman-llnl

cmake/BLTPrivateMacros.cmake
joshessman-llnl: This does work (I was surprised!) but I'm having trouble thinking of a situation where we'd want to query interface libraries for this property, since it really only applies to library targets created by the current project.

white238: Does it also auto propagate via CMake? Even if it doesnt it does continue the current CMake convention that INTERFACE properties propagate to their parents...

joshessman-llnl: It does not, but I think that is still a better name for it.

PR #449, opened 2021-01-11
Add infrastructure to perform device links for object libraries

joshessman-llnl:
The [CUDA_RESOLVE_DEVICE_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_RESOLVE_DEVICE_SYMBOLS.html) option allows device symbols to be resolved in "regular" libraries, meaning that they can be used without requiring the end user to link with NVCC.

Since the option does not extend to object libraries, this PR adds a device link step for targets that depend on object libraries for which the `CUDA_RESOLVE_DEVICE_SYMBOLS` property is enabled, or when the global option, `CMAKE_CUDA_RESOLVE_DEVICE_SYMBOLS` , is enabled.

Specifically, this is intended to address the use case in which object libraries are added to a user-facing static or shared library target.  Currently, if CUDA separable compilation is used within the object library, the associated symbols are undefined in the "final" library target, meaning that the NVCC linker is required by any users of that library.  To allow for use of this shared/static library with host-only linkers, a device link object for each object library would be created and added to the library via one of the above options.


docs/tutorial/creating_execs_and_libs.rst
```diff
@@ -122,3 +122,22 @@ the linker removing unused symbols in the larger library.
   Due to record keeping on BLT's part to make object libraries as easy to use
   as possible, you need to define object libraries before you use them
   if you need their inheritable information to be correct.
+
+If you are using separable CUDA compilation (relocatable device code) in your
+object library, users of that library will be required to use NVCC to link their
+executables - in general, only NVCC can perform the "device link" step.  To remove
+this restriction, you can enable the ``CUDA_RESOLVE_DEVICE_SYMBOLS`` property on
```
white238, line 129: Should they set this on non-object libraries? Or does this happen automatically for them?

joshessman-llnl: CMake enables this automatically for shared libraries (and it can be enabled for static libraries as well); I've added this to the docs here.

cmake/BLTOptions.cmake
```diff
@@ -57,6 +57,9 @@ option(ENABLE_CLANG_CUDA   "Enable Clang's native CUDA support" OFF)
 mark_as_advanced(ENABLE_CLANG_CUDA)
 set(BLT_CLANG_CUDA_ARCH "sm_30" CACHE STRING "Compute architecture to use when generating CUDA code with Clang")
 mark_as_advanced(BLT_CLANG_CUDA_ARCH)
+cmake_dependent_option(BLT_CUDA_RESOLVE_DEVICE_SYMBOLS
```
white238, line 60: Is this needed still?

white238 (approved):

white238:
@davidbeckingsale and @kennyweiss would you mind looking this over?

davidbeckingsale (approved):

PR #452, opened 2021-01-26
Remove target-specific generator expression from HIP_HIPCC_FLAGS

joshessman-llnl:
When adding compile definitions to a HIP target, the inclusive scan of compile definitions is used:
```cmake
blt_add_executable(NAME foo DEPENDS_ON hip DEFINES FOO_ONLY) # compile cmd will have -DFOO_ONLY
blt_add_executable(NAME bar DEPENDS_ON hip DEFINES BAR_ONLY) # compile cmd will have -DFOO_ONLY -DBAR_ONLY
blt_add_executable(NAME baz DEPENDS_ON hip DEFINES BAZ_ONLY) # compile cmd will have -DFOO_ONLY -DBAR_ONLY -DBAZ_ONLY
```

The following code in `FindHIP.cmake` appears to be the reason:
https://github.com/LLNL/blt/blob/b7e2bdcae5bd9f8250d05581d57dcfdf608bacd0/cmake/thirdparty/FindHIP.cmake#L486-L488

Specifically, target-specific generator expressions (for the compile definitions) are added to the global `HIP_HIPCC_FLAGS` variable.
Because the flags are written (via `generate_file`) to a `run_hipcc` script immediately after the above snippet, the state of `HIP_HCC_FLAGS` is "preserved" when the target is built.  This is also why the definitions are an inclusive scan and not a union.

Because the flags are preserved, the problematic generator expression can just be removed from the global flags after the `run_hipcc` script is written.

The following snippet also looks like it exhibits the same behavior, but I am not sure if it should be fixed in this PR or if we should wait to see if it affects any users:
https://github.com/LLNL/blt/blob/b7e2bdcae5bd9f8250d05581d57dcfdf608bacd0/cmake/thirdparty/FindHIP.cmake#L468-L470

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -70,6 +70,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   is not changed back to Fortran.
 - Executables that link to libraries that depend on `hip`/`hip_runtime`/`cuda`/`cuda_runtime`
   will automatically be linked with the HIP or CUDA (NVCC) linker
+- Patched an issue with the FindHIP macros that added the inclusive scan of specified
+  DEFINEs to compile commands
 
 ## [Version 0.3.6] - Release date 2020-07-27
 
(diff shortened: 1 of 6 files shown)


white238 (approved):

davidbeckingsale (approved):
Nice fix 👍

PR #459, opened 2021-02-10
Re-add previous iteration of openMP patching logic

joshessman-llnl:
It looks like the new approach I added in #421 that looks for the `openmp` imported library doesn't work for targets that don't call `target_link_libraries(openmp)` - namely, the old-style registered libraries.  To fix this I've added the old logic back in.


joshessman-llnl:
LGTM

white238:
LGTM

white238 (approved):
openmp strikes again!

kennyweiss (approved):
Thanks @joshessman-llnl and @white238 !

cyrush:
The return of `-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL-lSHELL` haunting us?

PR #480, opened 2021-04-12
fix import TPL example

white238:
swapped two variables.

### DIFF
--- a/docs/tutorial/third_party_libraries.rst
+++ b/docs/tutorial/third_party_libraries.rst
@@ -114,10 +114,11 @@ new imported library's ``NAME`` in the rest of your project.
 
 .. code-block:: cmake
 
-    # FindLua.cmake takes in LUA_DIR, which the directory where Lua was installed to
-    # and fills variables: LUA_FOUND, LUA_LIBRARIES, and LUA_INCLUDE_DIR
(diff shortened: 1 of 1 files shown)


kennyweiss (approved):

docs/tutorial/third_party_libraries.rst
```diff
@@ -114,10 +114,11 @@ new imported library's ``NAME`` in the rest of your project.
...
 
-    # FindLua.cmake takes in LUA_DIR, which the directory where Lua was installed to
-    # and fills variables: LUA_FOUND, LUA_LIBRARIES, and LUA_INCLUDE_DIR
-    find_package(Lua NO_DEFAULT_PATH
-                     PATHS ${LUA_DIR})
+    # FindLua.cmake takes in LUA_DIR as an environment variable, which is the directory 
+    # where Lua was installed to and fills variables: LUA_FOUND, LUA_LIBRARIES, and LUA_INCLUDE_DIR
+    set(ENV{LUA_DIR} ${LUA_DIR})
+
+    find_package(Lua)
```
white238, line 121: FindLua.cmake is odd to say the least.  It doesn't respect the normal PATHS variable.  It only uses environment variables or system locations.

PR #482, opened 2021-04-14
add option to use mpi imported targets

cyrush:
Adds an option to use the modern imported targets provided by FindMPI. 

This solves an issue with BLT related to transitive mpi deps and `-pthreads` vs cuda.

The issue:
- You build a TPL that doesn't have cuda support.  In that case, BLT does not (cannot) apply any of the cuda pthreads protection logic. (As far as I know we can't use generator exprs that involve non-enabled languages)
- You now use this TPL in a lib or application that does need CUDA support.  The flags that are tied to MPI are already baked in to the first lib, so nvcc sees `pthreads` and gives up.

Using the targets imported by FindMPI solves this issue, b/c those targets expand to include the proper guards when FindMPI is called.

I don't know the exact point at where it changed, but pthreads is added by FindMPI in newer versions of CMake. It does not happen in CMake 3.14.  We may eventually want to make this the default when newer CMake is used. 

TODO:
- [x] Better name for the option
- [x] Add some docs to https://llnl-blt.readthedocs.io/en/develop/tutorial/common_hpc_dependencies.html#commonhpcdependencies

### DIFF
--- a/cmake/thirdparty/SetupMPI.cmake
+++ b/cmake/thirdparty/SetupMPI.cmake
@@ -33,6 +33,29 @@ set(_mpi_link_flags )
 
 message(STATUS "Enable FindMPI:  ${ENABLE_FIND_MPI}")
 
+#
+# TODO: should this be BLT_ENABLE_FIND_MPI?
+#
+if(ENABLE_FIND_MPI)
+    message(STATUS "FindMPI Enabled  (ENABLE_FIND_MPI == ON)")
+else()
+    message(STATUS "FindMPI Disabled (ENABLE_FIND_MPI == OFF) ")
+endif()
+
+
+if(BLT_USE_FIND_MPI_TARGETS)
+    message(STATUS "Using MPI CMake imported targets (BLT_USE_FIND_MPI_TARGETS == ON)")
+else()
+    message(STATUS "Not using MPI CMake imported targets (BLT_USE_FIND_MPI_TARGETS == OFF)")
+endif()
+
+
+
+if(NOT ENABLE_FIND_MPI AND BLT_USE_FIND_MPI_TARGETS)
+    message(FATAL_ERROR "ENABLE_FIND_MPI == false, but BLT_USE_FIND_MPI_TARGETS == true."
+                        "BLT_USE_FIND_MPI_TARGETS == true requires ENABLE_FIND_MPI == true.")
+endif()
+
 if (ENABLE_FIND_MPI)
     find_package(MPI REQUIRED)
 
@@ -123,16 +146,33 @@ if (ENABLE_FIND_MPI)
 endif()
 
 # Allow users to override CMake's FindMPI
+# Assuming we are not directly using targets it provides
 if (BLT_MPI_COMPILE_FLAGS)
+    if(BLT_USE_FIND_MPI_TARGETS)
+        message(FATAL_ERROR "Cannot override MPI Compile Flags when BLT_USE_FIND_MPI_TARGETS == true. "
+                            "(BLT_MPI_COMPILE_FLAGS provided, but BLT_USE_FIND_MPI_TARGETS == true)")
+    endif()
     set(_mpi_compile_flags ${BLT_MPI_COMPILE_FLAGS})
 endif()
 if (BLT_MPI_INCLUDES)
+    if(BLT_USE_FIND_MPI_TARGETS)
+        message(FATAL_ERROR "Cannot override MPI Includes when BLT_USE_FIND_MPI_TARGETS == true. "
+                            "(BLT_MPI_INCLUDES provided, but BLT_USE_FIND_MPI_TARGETS == true)")
+    endif()
     set(_mpi_includes ${BLT_MPI_INCLUDES})
 endif()
 if (BLT_MPI_LIBRARIES)
+    if(BLT_USE_FIND_MPI_TARGETS)
+        message(FATAL_ERROR "Cannot override MPI Libraries when BLT_USE_FIND_MPI_TARGETS == true. "
+                            "(BLT_MPI_LIBRARIES provided, but BLT_USE_FIND_MPI_TARGETS == true)")
+    endif()
     set(_mpi_libraries ${BLT_MPI_LIBRARIES})
 endif()
 if (BLT_MPI_LINK_FLAGS)
+    if(BLT_USE_FIND_MPI_TARGETS)
+        message(FATAL_ERROR "Cannot override MPI Link Flags when BLT_USE_FIND_MPI_TARGETS == true. "
+                            "(BLT_MPI_LINK_FLAGS provided, but BLT_USE_FIND_MPI_TARGETS == true)")
+    endif()
     set(_mpi_link_flags ${BLT_MPI_LINK_FLAGS})
 endif()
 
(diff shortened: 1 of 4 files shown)


cmake/thirdparty/SetupMPI.cmake
```diff
@@ -177,10 +177,24 @@ if (ENABLE_FORTRAN)
...
 
-blt_import_library(NAME          mpi
-                   INCLUDES      ${_mpi_includes}
-                   TREAT_INCLUDES_AS_SYSTEM ON
-                   LIBRARIES     ${_mpi_libraries}
-                   COMPILE_FLAGS ${_mpi_compile_flags}
-                   LINK_FLAGS    ${_mpi_link_flags} 
-                   EXPORTABLE    ${BLT_EXPORT_THIRDPARTY})
+
+if(BLT_USE_MPI_IMPORTED_TARGETS)
```
white238, line 181: imported is a bit overloaded in this context.

white238: maybe "BLT_USE_CMAKE_MPI_TARGETS"?

cyrush: the other MPI options are all have the `BLT_MPI` prefix, like `BLT_MPI_COMPILE_FLAGS` 

how about:
`BLT_MPI_USE_FINDMPI_CMAKE_TARGETS`

white238: That works.. after this is proven to solve you problem, we should handle the overriding case where the user provides includes/link flags/etc.

cyrush: or 
`BLT_MPI_USE_FIND_MPI_CMAKE_TARGETS`

white238: BLT_USE_FIND_MPI_TARGETS?

cyrush:
@white238 pushed docs, error checks and improvements. I also verified this does resolve the issue. 
Let me know if I should make more tweaks.

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -33,6 +33,29 @@ set(_mpi_link_flags )
 
 message(STATUS "Enable FindMPI:  ${ENABLE_FIND_MPI}")
 
+#
+# TODO: should this be BLT_ENABLE_FIND_MPI?
+#
+if(ENABLE_FIND_MPI)
```
white238, line 39: Can you remove line 34 if you want to add this?

Yea this was a very early option.  It should be prefixed but I've been hesitant to change it due to legacy.  But honestly no one uses that I know anymore.  The main use case was BGQ.

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -33,6 +33,29 @@ set(_mpi_link_flags )
 
 message(STATUS "Enable FindMPI:  ${ENABLE_FIND_MPI}")
 
+#
+# TODO: should this be BLT_ENABLE_FIND_MPI?
```
kennyweiss, line 37: > TODO: should this be `BLT_ENABLE_FIND_MPI`?

I agree that we should convert it to `BLT_ENABLE_FIND_MPI` and deprecate `ENABLE_FIND_MPI`, e.g. via a loud warning at first.

Perhaps we should do the same for many/most blt `ENABLE_` variables at the same time?

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -33,6 +33,29 @@ set(_mpi_link_flags )
...
+
+if(BLT_USE_FIND_MPI_TARGETS)
+    message(STATUS "Using MPI CMake imported targets (BLT_USE_FIND_MPI_TARGETS == ON)")
+else()
+    message(STATUS "Not using MPI CMake imported targets (BLT_USE_FIND_MPI_TARGETS == OFF)")
+endif()
+
+
+
+if(NOT ENABLE_FIND_MPI AND BLT_USE_FIND_MPI_TARGETS)
```
kennyweiss, line 54: Seems like we could handle this better by converting `BLT_USE_FIND_MPI_TARGETS` to a `cmake_dependent_option`

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -123,16 +146,33 @@ if (ENABLE_FIND_MPI)
 endif()
 
 # Allow users to override CMake's FindMPI
+# Assuming we are not directly using targets it provides
 if (BLT_MPI_COMPILE_FLAGS)
+    if(BLT_USE_FIND_MPI_TARGETS)
```
kennyweiss, line 151: Can you please convert this to a `foreach()` ?
That way it is easier to see that it is the same logic.

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -177,10 +217,24 @@ if (ENABLE_FORTRAN)
...
+else()
+    # non imported targets case
+    blt_import_library(NAME          mpi
+                       INCLUDES      ${_mpi_includes}
+                       TREAT_INCLUDES_AS_SYSTEM ON
+                       LIBRARIES     ${_mpi_libraries}
+                       COMPILE_FLAGS ${_mpi_compile_flags}
+                       LINK_FLAGS    ${_mpi_link_flags} 
+                       EXPORTABLE    ${BLT_EXPORT_THIRDPARTY})
+endif()
```
white238, line 240: Add a new line here please.

docs/tutorial/common_hpc_dependencies.rst
```diff
@@ -86,6 +86,11 @@ Test. ``test_2.cpp`` provides an example driver for MPI with GoogleTest.
   BLT also has the variable ``ENABLE_FIND_MPI`` which turns off all CMake's ``FindMPI``
   logic and then uses the MPI wrapper directly when you provide them as the default
   compilers.
+  
+  BLT also provides ``BLT_USE_FIND_MPI_TARGETS`` that will directly use the
+  ``MPI::MPI_C`` and ``MPI::MPI_CXX`` targets provided by CMake's ``FindMPI``.
```
white238, line 91: Please add that it those targets are able to be referenced under the unified name of `mpi`

kennyweiss (approved):
Looks good @cyrush 

Please also update the `RELEASE_NOTES`

cmake/thirdparty/SetupMPI.cmake
white238: It would be a good time to do it since we just did the 0.4.0 release.

cmake/thirdparty/SetupMPI.cmake
cyrush: I tried `cmake_dependent_option`. It will just flip `BLT_USE_FIND_MPI_TARGETS` to OFF, when `ENABLE_FIND_MPI` is OFF. I think an explicit error is a better approach here --  better to error when things aren't consistent than to drive on to surprise town.

cmake/thirdparty/SetupMPI.cmake
cyrush: Added: https://github.com/LLNL/blt/issues/483

cmake/thirdparty/SetupMPI.cmake
cyrush: Each error case here provides a detailed unique error message. It's not just the var names that are different, so var indirection in a for loop won't get you the same result.

cyrush: Looking at the full section in the file -- the logic seems easy to parse.  I feel a cmake for loop would not improve the readability. If you have an idea of how it could work, lets explore in a future PR.

cmake/thirdparty/SetupMPI.cmake
cyrush: done

docs/tutorial/common_hpc_dependencies.rst
cyrush: done

cmake/thirdparty/SetupMPI.cmake
cyrush: removed line 34 and the TODO

docs/tutorial/common_hpc_dependencies.rst
white238: two backticks

cyrush: two backticks times two.

PR #486, opened 2021-04-22
Revert "add option to use mpi imported targets"

cyrush:
Reverts LLNL/blt#482

Life is complex -- turns out we don't need these options!

### DIFF
--- a/cmake/thirdparty/SetupMPI.cmake
+++ b/cmake/thirdparty/SetupMPI.cmake
@@ -31,25 +31,7 @@ set(_mpi_includes )
 set(_mpi_libraries )
 set(_mpi_link_flags )
 
-
-if(ENABLE_FIND_MPI)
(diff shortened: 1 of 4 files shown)


white238 (approved):

cmake/thirdparty/SetupMPI.cmake
```diff
@@ -31,25 +31,7 @@ set(_mpi_includes )
 set(_mpi_libraries )
 set(_mpi_link_flags )
 
-
-if(ENABLE_FIND_MPI)
-    message(STATUS "FindMPI Enabled  (ENABLE_FIND_MPI == ON)")
-else()
-    message(STATUS "FindMPI Disabled (ENABLE_FIND_MPI == OFF) ")
-endif()
-
```
cyrush, line 40: this is one think we should keep, i'll see if that's possible.

PR #500, opened 2021-06-17
Simplify clangformat version check

### LINKED ISSUE #499: regex match of clang-format --version output hanging
on rznevada, some users noticed their cmake process hanging. I was able to track this down to the regex match in https://github.com/LLNL/blt/blob/develop/cmake/SetupCodeChecks.cmake trying to parse the output of `clang-format --version`. I think the issue is that the Cray version of clang-format adds some unexpected strings. Compare the vanilla output:

[lee218@rzgenie2:build]$ /usr/tce/packages/clang/clang-11.0.1/bin/clang-format --version
clang-format version 11.0.1

versus the Cray version on rznevada:

[lee218@rznevada15:build]$ /opt/cray/pe/cce/12.0.0/cce-clang/x86_64/bin/clang-format --version
Cray clang-format version 12.0.0  (59041842ced75ac192c0f1331989ae45f58d7caf)

Would it be possible to modify the regex match to properly parse this output?

white238:
Fixes #499

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,6 +9,9 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 
 ## [Unreleased] - Release date yyyy-mm-dd
 
+### Changed
+-  Simpified the clang-format version regex that was causing hangs on some version strings.
+
 ## [Version 0.4.0] - Release date 2021-04-09
 
 ## Added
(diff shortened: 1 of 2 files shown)


joshessman-llnl (approved):
Fingers crossed that we don't run into this again :)

kennyweiss (approved):
Thanks @white238

lee218llnl:
@white238 did you miss my comment about double escaping the dot '.' character? As is, your changes return this:

Cray clang-format v

white238:
> 
> 
> @white238 did you miss my comment about double escaping the dot '.' character? As is, your changes return this:
> 
> Cray clang-format v

I did not and I can't find it in an email/issue/pr anywhere.  On the plus side it doesn't hang.

PR #501, opened 2021-06-18
fix escape of . character in clangformat regex

lee218llnl:
slight modification of regex from #500 for #499.

The changes in #500 return `Cray clang-format v` for the clang-format --version output `Cray clang-format version 12.0.0 (59041842ced75ac192c0f1331989ae45f58d7caf)`. This PR will correctly return `12.0.0`


white238 (approved):
Thanks for taking care of this @lee218llnl !

kennyweiss (approved):
Thanks @lee218llnl !

PR #512, opened 2021-08-13
change COMPILER_FAMILY_IS flags to CACHE variables so users can have …

robinson96:
…control over detected family type.


white238:
I am curious what caused this to not work?

white238 (approved):

white238:
I am concerned these variables would be pushed from directory scoped to global and are not prefixed with "BLT_".  Do codes ever use these variables? Can we just prefix them now so they dont potentially conflict with project variables?  Thoughts @kennyweiss ?

kennyweiss:
I agree with @white238's concern -- it seems that if we're exposing these to the cache, we should prefix them with `BLT_`, 
(and give users some notice that we're doing this).

I am also interested in hearing about what prompted this.
Being able to modify this might lead to some unexpected behavior, so it would be good to understand the use case.

PR #517, opened 2021-09-10
Minimal support for cmake 3.21 hip targets

trws:
The hip support in cmake 3.21 requires that the hip architecture is set
for all targets.  This sets the default based on the value of the
BLT_ROCM_ARCH, which is enough to satisfy it by default for all
targets.

### DIFF
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -66,6 +66,7 @@ set(BLT_CLANG_HIP_ARCH "gfx906" CACHE STRING "AMDGPU ISA target to use when gene
 mark_as_advanced(BLT_CLANG_HIP_ARCH)
 option(ENABLE_HCC         "Enable HCC support" OFF)
 set(BLT_ROCM_ARCH "gfx900" CACHE STRING "gfx architecture to use when generating ROCm code")
+set(CMAKE_HIP_ARCHITECTURES "${BLT_ROCM_ARCH}" CACHE STRING "")
 
 #------------------------------------------------------------------------------
 # Test Options
--- a/cmake/BLTOptions.cmake
+++ b/cmake/BLTOptions.cmake
@@ -65,8 +65,7 @@ mark_as_advanced(ENABLE_CLANG_HIP)
 set(BLT_CLANG_HIP_ARCH "gfx906" CACHE STRING "AMDGPU ISA target to use when generating HIP code with Clang" )
 mark_as_advanced(BLT_CLANG_HIP_ARCH)
 option(ENABLE_HCC         "Enable HCC support" OFF)
-set(BLT_ROCM_ARCH "gfx900" CACHE STRING "gfx architecture to use when generating ROCm code")
-set(CMAKE_HIP_ARCHITECTURES "${BLT_ROCM_ARCH}" CACHE STRING "")
+set(CMAKE_HIP_ARCHITECTURES "gfx900" CACHE STRING "gfx architecture to use when generating HIP/ROCm code")
 
 #------------------------------------------------------------------------------
 # Test Options
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -11,6 +11,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 
 ### Added
 - Added support for IntelLLVM compiler family to blt_append_custom_compiler_flag
+- Added support for hip targets configured with cmake 3.21 native hip support
 
 ### Changed
 - `BLT_C_FILE_EXTS` updated to include `.cuh`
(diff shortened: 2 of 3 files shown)


cmake/BLTOptions.cmake
```diff
@@ -66,6 +66,7 @@ set(BLT_CLANG_HIP_ARCH "gfx906" CACHE STRING "AMDGPU ISA target to use when gene
 mark_as_advanced(BLT_CLANG_HIP_ARCH)
 option(ENABLE_HCC         "Enable HCC support" OFF)
 set(BLT_ROCM_ARCH "gfx900" CACHE STRING "gfx architecture to use when generating ROCm code")
+set(CMAKE_HIP_ARCHITECTURES "${BLT_ROCM_ARCH}" CACHE STRING "")
```
white238, line 69: Can you move this to https://github.com/LLNL/blt/blob/develop/cmake/thirdparty/SetupHIP.cmake

This file is for command line options and the rest of the HIP setup logic is in that file.

Thanks!

trws: Possibly, that's a good policy question.  If only the BLT one is set it's pretty clear what to do, if the CMAKE one is set, probably pull it out into the BLT one, what if they are both set?  I would think either we combine them, since it's a list, or error out.  Do you have a preference?

white238: I agree that if one or the other is set to just set them both to the same thing.

If both are set to different values, I would lean towards erroring out.  Thoughts, @kennyweiss ?

kennyweiss: I agree -- if there are two conflicting config arguments we should error out with a useful error message.

Are these options really equivalent? If so, do we still need both? 
e.g. is it:
* for backwards compatibility with different versions of CMake?
* for symmetry with how we handle other architectures?
* <something else>?

trws: Yeah, it's backwards compatibility.  I could always go in and replace the blt option with CMAKE_HIP_ARCHITECTURES inside of BLT for all versions, it would remove the duplication and make usage a bit more consistent across versions, but it would be an interface break for blt users.

trws: Come to find out, the BLT_ROCM_ARCH option wasn't used anywhere.  As in it was literally never referenced elsewhere in the repo, so I'm removing it and replacing it with the CMAKE standard option from 3.21+ and adding a reference to it so it gets propagated correctly in older versions.

Doing that, the behavior is identical between `ENABLE_HIP` and `ENABLE_CLANG_HIP` for this flag, hip takes the same flag as clang does to propagate this option, which makes `CMAKE_HIP_ARCHITECTURES` effectively a duplicate of `BLT_CLANG_HIP_ARCH`.
I'm thinking my preference is this:

* Allow users to specify either `-DCMAKE_HIP_ARCHITECTURES` or the `HCC_AMDGPU_TARGET` environment variable to tell cmake what to do
* Remove `BLT_CLANG_HIP_ARCH`, use the value from first bullet for this case

Honestly it would be nice to do the same thing for the cuda support and merge out the `BLT_CLANG_CUDA_ARCH` as well.

white238: Thanks for finding that out @trws !  As far as I know @davidbeckingsale is the only one that used the clang cuda support.  I do like having both options the act the same way and as close to what the CMake standard is.

trws:
Ok, current version has these changes:

* BLT_CLANG_HIP_ARCH is gone
* both clang and hipcc hip compilation use CMAKE_HIP_ARCHITECTURES on all cmake versions
* on cmake versions below 3.21 the --offload-arch options are added from there to emulate the native behavior
* The `rocminfo` binary is no longer included with a hip install (as of at least 4.3), no clue why since rocm-smi doesn't cover all the same use-cases but it's gone, so the find_path was failing.  Switched that to key off a directory named `hip` existing under the path for now.

As far as I can tell, this now works for HIP with hipcc or amdclang or clang for every rocm version I have locally with hip support across cmake versions from at least 3.14 to 3.21.

rhornung67:
@trws are you sure the rocminfo binary is no longer included, or has it been moved to a different location? I ran into issues with 4.2 because AMD had reorganized their directory structure. To resolve the problem, I put the link line args we have in the RAJA hip host-config files. I got these from John G.

trws:
I was thinking the same thing at first @rhornung67, but for whatever reason it just wasn't there.  The odd part is, there is a rocminfo package, it is part of rocm-dev, and after a reinstall it is now there.  I'm still tempted to keep the find_path looking for something else though, perhaps something more specific like the main runtime header at `hip/include/hip/hip_runtime.h` since rocminfo is distributed as a completely separate package now.

kennyweiss (approved):
Thanks @trws

trws:
No problem, I added in a note in the release notes about the variable 
change.  If this works for everyone, I can come back and do a similar 
tweak on the cuda side soon.

white238 (approved):
Thanks @trws .

I will create the PR from fork in just a second to run the gitlab tests

PR #519, opened 2021-09-14
actually remove BLT_CLANG_HIP_ARCH variable

trws:
Apologies for the partial PR, I accidentally let the variable stuck around with its effect removed this cleans that up.


kennyweiss (approved):
Thanks for double checking and cleaning this up @trws

white238 (approved):

PR #522, opened 2021-09-14
Fix hip

### LINKED ISSUE #520: hipcc doesn't honor CMAKE_CXX_STANDARD
I'm not sure how this didn't come up before, perhaps because hipcc adds -std=c++11 by default, but it seems like hipcc doesn't receive the CMAKE_CXX_STANDARD flags either from cmake or from BLT.  I'm looking into this to find a workaround.

trws:
This is meant to fix the issues in #520.

It also works around a bug in FindHIP that was requiring environment variables and cmake variables to be set to properly initialize the input to the linking helper.  This lets the BLT_CXX_STD variable work for hipcc compiles, and removes a kinda nasty bug related to path detection in the AMD perl-based wrappers.

fixes #520

### DIFF
--- a/cmake/thirdparty/FindHIP.cmake
+++ b/cmake/thirdparty/FindHIP.cmake
@@ -83,6 +83,23 @@ if(UNIX AND NOT APPLE AND NOT CYGWIN)
     endif()
     mark_as_advanced(HIP_HIPCC_EXECUTABLE)
 
+    # Find hip clang executable
+    if(NOT DEFINED HIP_CLANG_PATH)
+        find_path(
+            HIP_CLANG_PATH
+            NAMES clang++
+            PATHS
+            "$ENV{HIP_CLANG_PATH}"
+            "$ENV{ROCM_PATH}"
+            "$ENV{HIP_PATH}/../llvm/bin"
+            "${HIP_ROOT_DIR}/../llvm/bin"
+            "/opt/rocm/llvm/bin"
+            PATH_SUFFIXES bin
+            NO_DEFAULT_PATH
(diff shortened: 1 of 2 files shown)


davidbeckingsale (approved):

davidbeckingsale (approved):

cmake/thirdparty/FindHIP.cmake
```diff
@@ -643,17 +662,6 @@ macro(HIP_ADD_EXECUTABLE hip_target)
         endif()
         set(CMAKE_HIP_LINK_EXECUTABLE "${HIP_HIPCC_CMAKE_LINKER_HELPER} ${HCC_HOME} <FLAGS> <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
     elseif("${HIP_COMPILER}" STREQUAL "clang")
-        if("x${HIP_CLANG_PATH}" STREQUAL "x")
-            if(DEFINED ENV{HIP_CLANG_PATH})
-                set(HIP_CLANG_PATH $ENV{HIP_CLANG_PATH})
-            elseif(DEFINED ENV{ROCM_PATH})
-                set(HIP_CLANG_PATH "$ENV{ROCM_PATH}/llvm/bin")
```
white238, line 650: This path seems lost in the above list.

trws: Yup, I had the base but not the suffix.  Thanks.

white238 (approved):

PR #525, opened 2021-09-17
Don't build executables depending on hip_runtime as HIP

davidbeckingsale:
Prior to this change, any executable depending on `hip_runtime` would be built as if it was HIP. This caused problems with C executables after #520 was fixed.

This PR adds a smoke test and fixes the issue.

### DIFF
--- a/cmake/BLTPrivateMacros.cmake
+++ b/cmake/BLTPrivateMacros.cmake
@@ -728,7 +728,7 @@ macro(blt_add_hip_executable)
         endif()
     endforeach()
 
-    if (${_depends_on_hip} OR ${_depends_on_hip_runtime})
+    if (${_depends_on_hip})
         # if hip is in depends_on, flag each file's language as HIP
         # instead of leaving it up to CMake to decide
         # Note: we don't do this when depending on just 'hip_runtime'
--- a/tests/smoke/CMakeLists.txt
+++ b/tests/smoke/CMakeLists.txt
@@ -226,6 +226,15 @@ if (ENABLE_HIP)
     blt_add_test(NAME blt_hip_runtime_smoke
                  COMMAND blt_hip_runtime_smoke)
 
+    blt_add_executable(NAME blt_hip_runtime_c_smoke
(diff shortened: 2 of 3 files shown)


trws:
Everything in here looks good to me.  Do you think it's worth supporting C files people want to compile as hip, and propagating the C standard flag rather than the c++ one, or leave that unsupported?

davidbeckingsale:
I think right now we should leave that unsupported, and ultimately transition off FindHIP to the hip-config and native CMake suport.

trws:
Sounds right to me.

I think the big trick with the hip-config support is going to be making everything play nice, since their exported target uses the same name.  Maybe we can re-use it and have BLT add on what it needs to theirs? Not sure.

davidbeckingsale:
> Sounds right to me. I think the big trick with the hip-config support is going to be making everything play nice, since their exported target uses the same name. Maybe we can re-use it and have BLT add on what it needs to theirs? Not sure.

Yup, that's what I'm working on right now. I am probably 90% of the way there, so only 90% more to go :)

white238 (approved):

cmake/BLTPrivateMacros.cmake
```diff
@@ -728,7 +728,7 @@ macro(blt_add_hip_executable)
         endif()
     endforeach()
 
-    if (${_depends_on_hip} OR ${_depends_on_hip_runtime})
+    if (${_depends_on_hip})
         # if hip is in depends_on, flag each file's language as HIP
         # instead of leaving it up to CMake to decide
         # Note: we don't do this when depending on just 'hip_runtime'
```
joshessman-llnl, line 734: remove this comment?

davidbeckingsale: I think the change makes it so this comment is now accurate (whereas it was incorrect before)

joshessman-llnl: Ah you're right - I had misread it the first time :)

PR #528, opened 2021-09-18
Use regexes for source file language detection instead of `get_filename_component(... ... EXT)`

### LINKED ISSUE #515: Cannot recognize the extension when compiling with xxx.xxx.c file
When compiling with some old C projects, source files with names like 1d.cube.order2.c cannot be recognized as C file. Error message 

```
CMake Error at blt/cmake/BLTPrivateMacros.cmake:822 (message):
  blt_split_source_list_by_language given source file with unknown file
  extension.  Add the missing extension to the corresponding list
  (BLT_C_FILE_EXTS, BLT_Fortran_FILE_EXTS, BLT_Python_FILE_EXTS, or
  BLT_CMAKE_FILE_EXTS).

   Unknown file: xxx/1d.cube.order2.c
  blt/cmake/BLTPrivateMacros.cmake:547 (blt_split_source_list_by_language)
  blt/cmake/BLTMacros.cmake:787 (blt_setup_cuda_target)
  CMakeLists.txt:109 (blt_add_executable)
```

will show up. Is there easy way to fix this?

goxberry:
Resolves #515.

This pull request changes the language source code extension detection code to match extensions via regular expressions on filenames instead of matching by using the output of `get_filename_component(... ... EXT)` and checking to see if that output is in a list of filename extensions.

`get_filename_component(... ... EXT)` will output the string formed from the leftmost period in the filename (ignoring directory prefixes), plus all characters to the right of that period. A challenge with this approach occurs when files are named, e.g., `quadrature_rule.order2.c` because `get_filename_component(... ... EXT)` applied to those files will return, e.g., `.order2.c`, which will not match `.c`. This limitation forces users to add extensions to BLT variables like `BLT_C_FILE_EXTS`. While this approach is tractable if a small number of extensions must be added, for larger numbers of "custom" extensions, this approach becomes cumbersome for users.

The approach in this commit instead forms a regex of the form, e.g., for C-like extensions,
```regex
^.*(\.cpp|\.hpp|\.cxx|\.cxx|\.hxx|\.c|\.h|\.cc|\.hh|\.inl|\.cu|\.cuh)$
```
by looping over file extension settings, contatenating the list with pipes, and escaping the period in file extensions appropriately. This regex will match the `.c` part of `quadrature_rule.order2.c` and appropriately identify it as a C-like source file.

Edit (Adding reasoning here for posterity):

- Source code filename extensions are now detected using regular
  expressions instead of checking whether the output of
  `get_filename_component(... ... EXT)` is in a list of filename
  extensions for each programming language (e.g.,
  `BLT_C_FILE_EXTS`). CMake's `get_filename_component(... ... EXT)`
  returns from its filename input a string containing the leftmost
  period of the filename and all characters to its right (stripping
  all directory prefixes, assuming use of a left-to-right-script-based
  character set). This behavior causes problems in filename extension
  detection with files that have names like `1d.cube.order2.c` because
  the `get_filename_component` call will output `.cube.order2.c`
  instead of `.c`.

  While users could add extensions like `.cube.order2.c` to BLT
  filename extension lists like `BLT_C_FILE_EXTS`, if many such
  "custom" extensions exist, adding each such extension could be
  cumbersome. Though CMake 3.14 and later support using
  `get_filename_component(... ... LAST_EXT)`, earlier versions of
  CMake do not, and BLT aims to support CMake 3.8+. Instead, regular
  expressions are used to detect that the filename extension of
  `1d.cube.order2.c` is `.c`. The regular-expression-based
  extension-matching implementation continues to support custom
  filename extensions, and should be backward-compatible with earlier
  versions of BLT.


cmake/BLTPrivateMacros.cmake
```diff
@@ -798,29 +832,30 @@ macro(blt_split_source_list_by_language)
             message(FATAL_ERROR "blt_split_source_list_by_language given source file with no extension: ${_file}")
         endif()
 
-        get_filename_component(_name "${_file}" NAME)  
-
-        string(TOLOWER "${_ext}" _ext_lower)
-
-        if("${_ext_lower}" IN_LIST BLT_C_FILE_EXTS)
-            if (DEFINED arg_C_LIST)
```
white238, line 806: Can you please indent by 4 spaces since that is what we do across the project? (or at least i fix it when i see it)

white238:
Awesome thanks @goxberry ! I was hoping to get to this soon but you beat me too it!

Can you add some tests for this in https://github.com/LLNL/blt/blob/develop/tests/internal/CMakeLists.txt?  I realize there were no tests for this before but I'm trying to fix that and regex's make me generally paranoid.

If you do not have time I will log an issue and write some soon.

cmake/BLTPrivateMacros.cmake
goxberry: Should be updated in [`f407f0c2`](https://github.com/LLNL/blt/pull/528/commits/f407f0c2be641477f37a9b7d5fb273168ebf6c8a)?

goxberry:
> Can you add some tests for this in https://github.com/LLNL/blt/blob/develop/tests/internal/CMakeLists.txt? I realize there were no tests for this before but I'm trying to fix that and regex's make me generally paranoid.

@white238 Same; I'll take a look.

cmake/BLTPrivateMacros.cmake
```diff
@@ -754,6 +754,24 @@ macro(blt_add_hip_executable)
 
 endmacro(blt_add_hip_executable)
 
+##-----------------------------------------------------------------------------
+## blt_error_if_target_exists(<extension_list> <regex_output_variable>)
```
kennyweiss, line 758:
```cmake
## blt_error_if_target_exists(<extension_list> <regex_output_variable>)
...
function(blt_make_regex_from_ext_list ext_list ext_regex)
```
Please make the docs consistent with the function name

```suggestion
## blt_make_regex_from_ext_list(<extension_list> <regex_output_variable>)
```

kennyweiss: Style note: BLT typically defines this sort of thing as a `macro` with named arguments instead of a `function`

cmake/BLTPrivateMacros.cmake
```diff
@@ -754,6 +754,24 @@ macro(blt_add_hip_executable)
...
+## This function converts the list of extensions in <extension_list>
+## to a regular expression in <regex_output_variable> that should match
+## any file name starting with a string and ending with any one of the
+## extensions in <extension_list>.
+##
+## Note that the <extension_list> argument has to be quoted, so that the
+## argument is passed as a string to this function.
+## -----------------------------------------------------------------------------
+function(blt_make_regex_from_ext_list ext_list ext_regex)
+  message(STATUS "ext_list = ${ext_list}")
```
kennyweiss, line 769: This would like be too verbose for typical usage. 

Once we're satisfied with the functionality, please consider removing/commenting this out, or adding a verbose flag to the function/macro to have it be quiet by default.

cmake/BLTPrivateMacros.cmake
```diff
@@ -783,6 +801,22 @@ macro(blt_split_source_list_by_language)
...
+    set(BLT_Python_FILE_REGEX "")
+    blt_make_regex_from_ext_list("${BLT_Python_FILE_EXTS}" BLT_Python_FILE_REGEX)
+    set(BLT_CMAKE_FILE_REGEX "")
+    blt_make_regex_from_ext_list("${BLT_CMAKE_FILE_EXTS}" BLT_CMAKE_FILE_REGEX)
+
+    # Normalize the case of the regexes
+    string(TOLOWER "${BLT_C_FILE_REGEX}" BLT_C_FILE_REGEX)
+    string(TOLOWER "${BLT_Fortran_FILE_REGEX}" BLT_Fortran_FILE_REGEX)
+    string(TOLOWER "${BLT_Python_FILE_REGEX}" BLT_Python_FILE_REGEX)
+    string(TOLOWER "${BLT_CMAKE_FILE_REGEX}" BLT_CMAKE_FILE_REGEX)
```
kennyweiss, line 818: Could this normalization be moved into `blt_make_regex_from_ext_list` ?

RELEASE-NOTES.md
```diff
@@ -16,6 +16,30 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
...
+  filename extension lists like `BLT_C_FILE_EXTS`, if many such
+  "custom" extensions exist, adding each such extension could be
+  cumbersome. Though CMake 3.14 and later support using
+  `get_filename_component(... ... LAST_EXT)`, earlier versions of
+  CMake do not, and BLT aims to support CMake 3.8+. Instead, regular
+  expressions are used to detect that the filename extension of
+  `1d.cube.order2.c` is `.c`. The regular-expression-based
+  extension-matching implementation continues to support custom
+  filename extensions, and should be backward-compatible with earlier
+  versions of BLT.
```
kennyweiss, line 42: This is a really nice overview of the problem @goxberry

Having said that, I think most of it should be moved elsewhere since the `RELEASE_NOTES` are typically a brief note about what changed.

Consider moving this to the `Fixed` section (since this is a bugfix/improvement)
and modify the writeup to something like the following:
```
- Source code filename extensions now only consider the substring containing the last period, e.g. `1d.cube.order2.c` is considered a `.c` file.
```

Perhaps the current writeup can be used for user docs (in an `.rst` file), or as dev docs as an extended cmake comment)

kennyweiss (changes requested):
Thanks @goxberry -- this is a nice improvement.

I agree with @white238 that we should add tests with this PR to ensure that it works in all the expected use cases, including:
* default file extensions for c/c++, Fortran, python, cmake, ....
* user-supplied/custom file extensions, e.g. lua, or different fortran extensions

Do we need to consider custom cases that have two periods? E.g. `file.foo.bar` with custom extension `.foo.bar` ?  I'd think not, but just putting it out there in case there are actual use cases of this form.

white238:
@ltaylor16 I have clarified in multiples places why case-insensitive file extensions is fine here.

.gitlab/build_and_test.sh
```diff
@@ -15,6 +15,7 @@ cd $BUILD_DIR
 
 cmake -C ../host-configs/llnl/$SYS_TYPE/$HOST_CONFIG ../tests/internal
 make -j8
+make style
```
kennyweiss, line 18: > make style

Does this do anything?

Is the intention that the CI should fail if the formatting is off?
I don't think that will happen by just calling `make style`.

RELEASE-NOTES.md
```diff
@@ -16,7 +16,13 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
...
 - `BLT_C_FILE_EXTS` updated to include `.cuh`
 - Fold `BLT_CLANG_HIP_ARCH` into the `CMAKE_HIP_ARCHITECTURES` variable
-- When using `ENABLE_ALL_WARNINGS`, append the flag to the beginning of `CMAKE_{C,CXX}_FLAGS` instead of the end
+- When using `ENABLE_ALL_WARNINGS`, append the flag to the beginning of `CMAKE_{C,CXX}_FLAGS` instead
+  of the end
+
+### Fixed
+- Source code filename extension filtering now utilize regular expressions to allow
+  greater user customization and improves handling of file names with multiple
+  periods, e.g. `1d.cube.order2.c` is considered a `.c` file.
```
kennyweiss, line 25: Suggestion -- slight rewording:
```suggestion
- Source code filename extension filtering now uses regular expressions to allow
  for more user customization and to improve handling of file names with multiple
  periods, e.g. `1d.cube.order2.c` is considered a `.c` file.
```

azure-pipelines.yml
```diff
@@ -53,7 +53,8 @@ steps:
   displayName: 'OSX/Windows CMake'
   condition: or( eq( variables['Agent.OS'], 'Windows_NT'), eq( variables['Agent.OS'], 'Darwin'))
 - script: |
-    cmake --build build --config Release 
+    cmake --build build --config Release
+    cmake --build build --config Release --target style
```
kennyweiss, line 57: Similarly here, what is the intent in adding these?

Perhaps we should add a CI job that checks the formatting via `make check` and fail if the code's not formatted

kennyweiss: Or, is this just meant to catch a situation where `make style` crashes ?

tests/internal/CMakeLists.txt
```diff
@@ -269,6 +269,132 @@ if(ENABLE_HIP)
   add_subdirectory(src/hip_defines_test)
 endif()
 
+#------------------------------------------------------------------------------
```
kennyweiss, line 272: 👍 
This is really nice! Thanks for adding it.

I wonder if it should be pulled out into its own file and included here instead of being inlined?

cmake/BLTPrivateMacros.cmake
```diff
@@ -754,19 +754,63 @@ macro(blt_add_hip_executable)
...
 ##                                    CMAKE_LIST <list name>)
 ##
 ## Filters source list by file extension into C/C++, Fortran, Python, and
 ## CMake source lists based on BLT_C_FILE_EXTS, BLT_Fortran_FILE_EXTS,
-## and BLT_CMAKE_FILE_EXTS (global BLT variables). Files named
-## "CMakeLists.txt" are also filtered here. Files with no extension
-## or generator expressions that are not object libraries (of the form
-## "$<TARGET_OBJECTS:nameofobjectlibrary>") will throw fatal errors.
+## and BLT_CMAKE_FILE_EXTS (global BLT variables). This filtering is
+## case-insensitive. Files named "CMakeLists.txt" are also filtered here.
```
kennyweiss, line 810: Should we document somewhere that `BLT_{C,Fortran,CMAKE,Python}_FILE_EXTS` are global variables that can be overridden?

E.g. by adding a page to the user docs listing the variables that support being overridden? (And noting that doing so is an advanced feature?)

kennyweiss (approved):
Thanks @goxberry  and @white238  -- this is a nice improvement

tests/internal/CMakeLists.txt
white238: we need a much larger followup PR that increases testing of BLT .. i can pull it into its own file

azure-pipelines.yml
white238: These actually don't do anything due to BLT not finding any formatters... made an issue ( #546) and will address this later.

cmake/BLTPrivateMacros.cmake
white238: They are documented here (https://llnl-blt.readthedocs.io/en/develop/api/code_check.html#blt-add-code-checks) but there could be a better place.

This goes along with previously discussed topic of documenting BLT's options better. Made an issue so we don't forget #547 and #548.

tests/internal/unit/CMakeLists.txt
```diff
@@ -0,0 +1,152 @@
...
+# 
+# SPDX-License-Identifier: (BSD-3-Clause)
+
+#------------------------------------------------------------------------------
+# BLT Internal Testing Project
+#------------------------------------------------------------------------------
+
+cmake_minimum_required(VERSION 3.8)
+
+project(blt-internal-tests LANGUAGES C CXX)
```
joshessman-llnl, line 12: Is it weird to have two `project` commands with the same name within a given build tree?

joshessman-llnl (approved):

tests/internal/unit/CMakeLists.txt
white238: copy/paste error... ill fix it. thanks for the catch!

PR #541, opened 2021-11-03
Patch hip::device target to allow compilation/running without FindHIP

joshessman-llnl:
```sh
$ ml cmake/3.21
$ cmake -C host-configs/llnl/toss_4_x86_64_ib_cray/cce@12.0.3_clang_hip.cmake tests/internal/
$ make blt_hip_smoke
$ ctest -R blt_hip_smoke -VV
...
test 11
    Start 11: blt_hip_smoke

11: Test command: /g/g16/essman1/axom/src/cmake/blt/tests/build/tests/blt_hip_smoke
11: Test timeout computed to be: 1500
1/1 Test #11: blt_hip_smoke ....................   Passed    0.03 sec

The following tests passed:
	blt_hip_smoke

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.05 sec
```


host-configs/llnl/toss_4_x86_64_ib_cray/cce@12.0.3_clang_hip.cmake
```diff
@@ -0,0 +1,52 @@
...
+
+# Fortran support
+set(ENABLE_FORTRAN ON CACHE BOOL "")
+set(CMAKE_Fortran_COMPILER "${CCE_HOME}/bin/crayftn" CACHE PATH "")
+# _blt_tutorial_compiler_config_end
+
+#------------------------------------------------------------------------------
+# MPI Support
+#------------------------------------------------------------------------------
+# _blt_tutorial_mpi_config_start
```
kennyweiss, line 30: >  _blt_tutorial_mpi_config_start

I'd suggest removing these start/end comments from this host-config.
They are related to our sphinx docs in other host-configs.

host-configs/llnl/toss_4_x86_64_ib_cray/cce@12.0.3_clang_hip.cmake
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2017-2021, Lawrence Livermore National Security, LLC and
```
kennyweiss, line 1: Since this compiler set is currently available on some machines, but not others (e.g. `rzwhamo` but not `rznevada`), I'm wondering if `rzwhamo` should be in the name of the file.

(It's in the comments and should work on `rznevada` in the near future, so perhaps that's enough).

cmake/thirdparty/SetupHIP.cmake
```diff
@@ -36,3 +36,26 @@ blt_import_library(NAME          blt::hip_runtime
                    DEPENDS_ON    hip::host
                    TREAT_INCLUDES_AS_SYSTEM ON
                    EXPORTABLE    ${BLT_EXPORT_THIRDPARTY})
+
+if(DEFINED AMDGPU_TARGETS)
```
kennyweiss, line 40: Where is `AMDGPU_TARGETS` defined? 
I don't see it in the host-configs.

Does it come from `find_package(hip )` ?

Would it make sense to add a comment to this file?

cmake/thirdparty/SetupHIP.cmake
```diff
@@ -36,3 +36,26 @@ blt_import_library(NAME          blt::hip_runtime
...
+    # If we haven't selected a particular architecture via CMAKE_HIP_ARCHITECTURES,
+    # we want to remove the unconditionally added compile/link flags from the hip::device target.
+    # FIXME: This may cause problems for targets whose HIP_ARCHITECTURES property differs
+    # from CMAKE_HIP_ARCHITECTURES
```
kennyweiss, line 44: >  FIXME: This may cause problems for targets whose HIP_ARCHITECTURES property differs
    from CMAKE_HIP_ARCHITECTURES

Could you please expand on this concern? Where is `HIP_ARCHITECTURES` defined?

kennyweiss (approved):
Thanks @joshessman-llnl !

Some notes:
* Did you also check this against `cmake@3.19`?
* This PR will be easier to review once the parent PR merges in the develop branch.

cmake/thirdparty/SetupHIP.cmake
kennyweiss: Found it: https://cmake.org/cmake/help/latest/prop_tgt/HIP_ARCHITECTURES.html

From the docs, it sounds like a user would have to modify the `HIP_ARCHITECTURES` property after it is initialized from the `CMAKE_HIP_ARCHITECTURES` variable.

host-configs/llnl/toss_4_x86_64_ib_cray/cce@12.0.3_clang_hip.cmake
kennyweiss: This comment is outdated -- `rznevada` now has `rocm@4.5`

davidbeckingsale (approved):

PR #542, opened 2021-11-04
add example host-config for intel-oneapi + c++17 on toss3

white238:
I will never remember the required "rh" so here it is so I don't keep forgetting....

### DIFF
--- a/host-configs/llnl/toss_3_x86_64_ib/intel-oneapi@2021.3-c++17.cmake
+++ b/host-configs/llnl/toss_3_x86_64_ib/intel-oneapi@2021.3-c++17.cmake
@@ -39,6 +39,7 @@ set(BLT_CXX_STD "c++17")
 
 # Use non-system default stdlib
 set(GCC_VERSION "gcc-9.3.1")
+# Note: gcc-8.3.1 and above need the "/rh" directory. Below that version do not.
 set(GCC_HOME "/usr/tce/packages/gcc/${GCC_VERSION}/rh")
 
 set(BLT_C_FLAGS "--gcc-toolchain=${GCC_HOME}" CACHE STRING "")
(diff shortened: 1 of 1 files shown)


host-configs/llnl/toss_3_x86_64_ib/intel-oneapi@2021.3-c++17.cmake
```diff
@@ -0,0 +1,60 @@
...
+
+#------------------------------------------------------------------------------
+# Extra flags
+#------------------------------------------------------------------------------
+
+set(BLT_CXX_STD "c++17")
+
+# Use non-system default stdlib
+set(GCC_VERSION "gcc-9.3.1")
+set(GCC_HOME "/usr/tce/packages/gcc/${GCC_VERSION}/rh")
```
kennyweiss, line 42: Please add a comment that gcc versions `8.3.1` and above require the `rh` suffix in their path

kennyweiss (approved):
Thanks @white238

host-configs/llnl/toss_3_x86_64_ib/intel-oneapi@2021.3-c++17.cmake
ltaylor16: What the difference?  There's so many links in the directories it's hard to see what's going on.

white238: From LC:

"The version up to and include 8.1.0 are standalone installations, while 8.3.1 and above actually point the build in /opt. Sorry for any confusion."

ltaylor16: OK, I remember. I ran into this problem earlier where scipy wanted a symbol from stdlibc++ but it could not be found.  scipy was compiled with 4.9.3 which had a newer version of stdlibc++ than the version 8.3.1 which wanted to use from /usr/lib64.  LC told me to load with the stdlibc++ from 8.1.0.  But if I use the rh directory, things should just work? I found a stdlibc++ directory under rh.

joshessman-llnl (approved):

PR #543, opened 2021-11-05
Map Flang Fortran COMPILER_ID to Clang family of compilers.

robinson96:
The amdflang compiler reports as Flang, would like to group this with Clang family of compilers.


adayton1 (approved):
LGTM

kennyweiss (approved):
Thanks @robinson96 -- assuming that `flang` flags are the same as `clang` flags, this seems like the right thing to do

white238 (approved):

PR #564, opened 2022-03-08
Add support for C++20

andrewcorrigan:
I'm a new user of `Blt` and the code I'm hoping to use it for requires certain features from C++20 (equivalent to `gcc-9` or higher).  Therefore, I went ahead and added C++20 support.  It'd be great to get this support merged into `main`.

- I see that you just branched version 0.5 a few hours ago (congrats!), but had not yet updated `develop` yet, therefore I based this commit on [49fb804b15d781b3a4ce57e8835e0d6addec715d].  Please let me know if you need me to rebase to a different commit.
- I don't have immediate access to `PGI` to test and could only find documentation online for the `--c++17` flags and therefore cautiously disabled it along with `XL`.

I'd be happy to address any feedback or requested changes.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,6 +9,10 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 
 ## [Unreleased] - Release date yyyy-mm-dd
 
+### Added
+- Added support for C++20. Note: XL does not support C++20.
+  While PGI has C++20 support, it is currently disabled (A BLT fatal error will occur).
+
 ## [Version 0.5.0] - Release date 2022-03-07
 
 ### Added
--- a/cmake/SetupCompilerOptions.cmake
+++ b/cmake/SetupCompilerOptions.cmake
@@ -263,7 +263,7 @@ endif ()
 SET( CMAKE_CXX_STANDARD_REQUIRED ON )
 
 set(BLT_CXX_STD "" CACHE STRING "Version of C++ standard")
-set_property(CACHE BLT_CXX_STD PROPERTY STRINGS c++98 c++11 c++14 c++17)
+set_property(CACHE BLT_CXX_STD PROPERTY STRINGS c++98 c++11 c++14 c++17 c++20)
 
 if (BLT_CXX_STD)
     if( BLT_CXX_STD STREQUAL c++98 ) 
@@ -296,9 +296,22 @@ if (BLT_CXX_STD)
             FLAGS_VAR CMAKE_CXX_FLAGS
             DEFAULT " "
             PGI "--c++17")
+    elseif( BLT_CXX_STD STREQUAL c++20)
+        # Error out on what does not support C++20
+        if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "XL")
+            message(FATAL_ERROR "XL does not support C++20.")
+        endif()
+        if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
+            message(FATAL_ERROR "PGI is not yet supported using C++20.")
+        endif()
+        if (ENABLE_CUDA AND (NOT DEFINED CMAKE_CUDA_COMPILE_FEATURES OR (NOT "cuda_std_20" IN_LIST CMAKE_CUDA_COMPILE_FEATURES)))
+            message(FATAL_ERROR "CMake's CUDA_STANDARD does not support C++20.")
+        endif()
+
+        set(CMAKE_CXX_STANDARD 20)
     else()
         message(FATAL_ERROR "${BLT_CXX_STD} is an invalid entry for BLT_CXX_STD. "
-                            "Valid Options are ( c++98, c++11, c++14, c++17 )")
+                            "Valid Options are ( c++98, c++11, c++14, c++17, c++20 )")
     endif()
 
     if (ENABLE_CUDA)


white238:
Thanks for the contribution but yes we definitely hit some poor timing.  For security reasons, we pull forks locally so they can run our normal CI checks.  I will attempt this, rebase, then make sure proper commit credit is still there.

kennyweiss:
Thanks @andrewcorrigan!

The following page appears to confirm your suspicions about XL's `C++20` support: https://en.cppreference.com/w/cpp/20
PGI 20.7 and 20.11 appear to support large swathes of `C++20`, but for the time-being, it seems safe for BLT to not support this until someone needs it.

Can you please respond with the compiler(s) that you tested this against?
From your description, it looks like gcc-9 (or newer).

We should add CI tests for one or more configurations with `C++20` support. It seems reasonable to add this as a follow-up PR.

It would probably also be a good idea for us to look into CMake's support for C++20 flags for various compilers...

white238:
This claims C++20 has been in CMake since 3.12 https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

white238:
I think it's safe to merge this as is and when we revamp BLT's CI (soon) we can add a lot more compilers and find the holes in the previous logic as well.

kennyweiss:
> I think it's safe to merge this as is and when we revamp BLT's CI (soon) we can add a lot more compilers and find the holes in the previous logic as well.

I agree.

kennyweiss (approved):
Thanks @andrewcorrigan

andrewcorrigan:
> Thanks @andrewcorrigan!
> 
> The following page appears to confirm your suspicions about XL's `C++20` support: https://en.cppreference.com/w/cpp/20 PGI 20.7 and 20.11 appear to support large swathes of `C++20`, but for the time-being, it seems safe for BLT to not support this until someone needs it.

I saw that too, but don't know if I have access to that compiler to test it or verify that the analogous compiler flag to what is set for C++17 would work.  Another reasonable solution I considered was to not check for `PGI` at all rely on `cmake` to do the right thing.

> 
> Can you please respond with the compiler(s) that you tested this against? From your description, it looks like gcc-9 (or newer).

Since I'm still working on adding BLT my project, so far I tried `blt/tests/internal` with `-DBLT_CXX_STD=c++20`

on Mac:
- `AppleClang 13.0.0.13000029`:  `make VERBOSE=1` built with `-std=c++2a`

on Linux:
- `Clang 13.0.1`:  `make VERBOSE=1` built with `-std=c++20`
- `Clang 9.0.1`:  `make VERBOSE=1` built with `-std=c++2a`
- `GNU 11.2.0``make VERBOSE=1` built with `-std=c++20`
- `GNU 9.3.0`:`make VERBOSE=1` built with `-std=c++2a`

In all cases `make test` reported `100% tests passed`.  This was with with `cmake version 3.22.2` in case it matters.

> It would probably also be a good idea for us to look into CMake's support for C++20 flags for various compilers...

FWIW, prior to adopting BLT, I've specified `CMAKE_CXX_STANDARD`, `CMAKE_CUDA_STANDARD`, and `CMAKE_HIP_STANDARD` directly across various versions of Clang, GCC, MSVC (2019 and 2022), and recent versions of IntelLLVM (`intel/oneapi.2021.4` on `rzslic`), and it has worked reliably, at least with recent versions of `cmake`

andrewcorrigan:
I just tracked down a recent version of PGI... Apparently it's part of "NVHPC":

```
% pgc++ --version

pgc++ (aka nvc++) 21.7-0 64-bit target on x86-64 Linux -tp skylake 
PGI Compilers and Tools
Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```

```
% cmake -DBLT_CXX_STD=c++20 ../ && make -j VERBOSE=1 && make test
-- The C compiler identification is NVHPC 21.7.0
-- The CXX compiler identification is NVHPC 21.7.0
```

Unfortunately both `-DBLT_CXX_STD=c++17` and `-DBLT_CXX_STD=c++20` fail when running the above test with linker errors:
```
CMakeFiles/t_example_compile_definitions_test_variant_2.dir/src/t_example_compile_definitions.cpp.o: In function `blt_compile_definitions_check_BLT_A_defined_Test::~blt_compile_definitions_check_BLT_A_defined_Test()':
nvc++aaazazi1TyAS.ll:(.text+0x94): undefined reference to `operator delete(void*, unsigned long)'
```

I don't think the issue is unique to this PR.

andrewcorrigan:
OK, PGI seems to be working on a different machine, but I'm getting

```
CMake Error in src/object_library_test/CMakeLists.txt:
  Target "object" requires the language dialect "CXX20" , but CMake does not
  know the compile flags to use to enable it.
```

andrewcorrigan:
I also tested as above with:
```
module load intel/oneapi.2021.4
export CXX=icpx
export CC=icx
export CFLAGS="--gcc-toolchain=/usr/tce/packages/gcc/gcc-10.2.1/rh/usr/"
export CXXFLAGS="${CFLAGS}"
```
BLT/CMake added the correct `-std=c++20` flag, all tests built and passed

andrewcorrigan:
> OK, PGI seems to be working on a different machine, but I'm getting
> 
> ```
> CMake Error in src/object_library_test/CMakeLists.txt:
>   Target "object" requires the language dialect "CXX20" , but CMake does not
>   know the compile flags to use to enable it.
> ```

I looked into this a bit more.  There are two possibilities depending on what approach is taken:

1. If we rely on `cmake` to choose the appropriate flag, then `cmake` complains:
```
  Target "test_1" requires the language dialect "CXX20" , but CMake does not
  know the compile flags to use to enable it.
```
If I proceed to build anyway, the tests all build without specifying any `c++20`-related flag and all tests pass, but that's not acceptable behavior.

2. If we specify as is done for C++17
```
            FLAGS_VAR CMAKE_CXX_FLAGS
            DEFAULT " "
            PGI "--c++20")
```
`cmake` still complains that:
``` 
 Target "test_1" requires the language dialect "CXX20" , but CMake does not
  know the compile flags to use to enable it.
```
However, if I proceed to build anyway, then the tests compiles with flags:
```
/usr/tce/packages/pgi/pgi-22.1/bin/pgc++   -Minform=warn   --c++20 
```

However, I'm seeing similar the same linker errors as above, this time on an LC system:
```
CMakeFiles/t_header_only_smoke.dir/src/t_header_only_smoke.cpp.o: In function `std::default_delete<std::string>::operator()(std::string*) const':
/usr/tce/packages/gcc/gcc-4.9.3/include/c++/4.9.3/bits/unique_ptr.h:76: undefined reference to `operator delete(void*, unsigned long)'
```

It seems to me like CMake isn't ready for PGI + C++20

RELEASE-NOTES.md
```diff
@@ -9,6 +9,10 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 
 ## [Unreleased] - Release date yyyy-mm-dd
 
+### Added
+- Added support for C++20. Note: XL does not support C++20.
+  While PGI has C++20 support, it is currently disabled (A BLT fatal error will occur).
```
kennyweiss, line 14: Thanks @andrewcorrigan !

PR #565, opened 2022-03-08
Add support for C++20

white238:
Local PR for #564

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,6 +9,10 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 
 ## [Unreleased] - Release date yyyy-mm-dd
 
+### Added
+- Added support for C++20. Note: Neither XL or PGI support
+  C++20 (A BLT fatal error will occur).
+
 ## [Version 0.5.0] - Release date 2022-03-07
 
 ### Added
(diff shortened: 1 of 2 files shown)


white238:
Thanks @andrewcorrigan !  Made one small change.  After @kennyweiss approves it I will merge it.

RELEASE-NOTES.md
```diff
@@ -9,6 +9,10 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 
 ## [Unreleased] - Release date yyyy-mm-dd
 
+### Added
+- Added support for C++20. Note: Neither XL or PGI support
+  C++20 (A BLT fatal error will occur).
```
kennyweiss, line 14: We might want to clarify this line a bit, since it looks like PGI@20.7+ support `C++20`, but (with this PR), BLT does not yet support `C++20` with PGI.

See: https://en.cppreference.com/w/cpp/20

kennyweiss (approved):

RELEASE-NOTES.md
andrewcorrigan: Updated here: https://github.com/andrewcorrigan/blt/commit/14a6e5e8793bdb165e1b36d4831c971de6f6d1d4

white238: I think that is safe to do in this case considering we don't actually have customers using PGI + C++20 support.  We can turn it on when that happens.

PR #568, opened 2022-03-11
added build and config type to build output fixes #567

### LINKED ISSUE #567: Output build type for debug purposes
This isn't output by CMake during the build output.  This would be helpful information.

Something like:

```
Build Type: <Debug|Release|RelWithDebug|etc>
```

Should also check if it is building with multi-configuration.

chapman39:
fixes #567


SetupBLT.cmake
```diff
@@ -212,3 +214,13 @@ if (NOT BLT_LOADED)
     endif()
 
 endif() # only load BLT once!
+
```
white238, line 217: Please remove all these extra lines

SetupBLT.cmake
```diff
@@ -33,6 +33,8 @@ if (NOT BLT_LOADED)
     # Show CMake info right out of the gate
     ################################
     message(STATUS "CMake Version: ${CMAKE_VERSION}")
+    message(STATUS "CMake Build Type: ${CMAKE_BUILD_TYPE}")
+    message(STATUS "CMake Configuration Types: ${CMAKE_CONFIGURATION_TYPE}")
```
white238, line 37:
```suggestion
    if(DEFINED CMAKE_BUILD_TYPE)
        message(STATUS "CMake Build Type: ${CMAKE_BUILD_TYPE}")
    endif()
    if(DEFINED CMAKE_CONFIGURATION_TYPES)
         message(STATUS "CMake Configuration Types: ${CMAKE_CONFIGURATION_TYPES}")
    endif()
```

I think these toggle between each other depending if it is a regular vs multi configuration build.  Also the latter variable is plural.

Can you try this out to make sure it looks correct in either case?

white238 (approved):

SetupBLT.cmake
```diff
@@ -33,6 +33,12 @@ if (NOT BLT_LOADED)
     # Show CMake info right out of the gate
     ################################
     message(STATUS "CMake Version: ${CMAKE_VERSION}")
+    if (DEFINED CMAKE_BUILD_TYPE)
+        message(STATUS "CMake Build Type: ${CMAKE_BUILD_TYPE}")
+    endif()
+    if (DEFINED CMAKE_CONFIGURATION_TYPES)
+        message(STATUS "CMake Configuration Types: ${CMAKE_CONFIGURATION_TYPES}")
+    endif()
```
kennyweiss, line 41: Suggestion: Would it make sense to check the `GENERATOR_IS_MULTI_CONFIG` property to see which one is expected? Looks like this has been available since version CMake@3.9.

E.g. something like:
```suggestion
    get_property(_is_multi_config GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
    if(_is_multi_config)
        message(STATUS "CMake Configuration Types: ${CMAKE_CONFIGURATION_TYPES}")
    else()
        message(STATUS "CMake Build Type: ${CMAKE_BUILD_TYPE}")
    endif()
    unset(_is_multi_config)
```

See: https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html

kennyweiss (approved):
Thanks @chapman39 !

SetupBLT.cmake
chapman39: good to know! just changed it and tested it (for single target generator only though). seems to work fine.

kennyweiss: Thanks @chapman39 !

PR #578, opened 2022-03-23
Support CMAKE_HIP_ARCHITECTURES

### LINKED ISSUE #575: Support HIP standards like CUDA standards
https://cmake.org/cmake/help/git-stage/prop_tgt/HIP_STANDARD.html

Logic in `cmake/SetupCompilerStandards.cmake`

white238:
Fixes #575 

* Removes a rather loud warning in one of our hip smoke tests
* Different versions of FindHIP added flags as generator expressions which was a case we were not handling properly.  This fixes that as well (rznevada, CMake 3.21.1, rocm 4.5.2)


tests/smoke/blt_hip_gtest_smoke.cpp
```diff
@@ -30,6 +30,9 @@ TEST(blt_hip_gtest_smoke,basic_assert_example)
   int num_threads = STR_LENGTH;
   int num_blocks = 1;
   hipLaunchKernelGGL((hello), dim3(num_blocks), dim3(num_threads),0,0);
-  hipDeviceSynchronize();
+  if(hipSuccess != hipDeviceSynchronize())
```
kennyweiss, line 33: Does `hipSuccess` come from `hip/hip_runtime.h` ?

kennyweiss (approved):
Thanks @white238!

tests/smoke/blt_hip_gtest_smoke.cpp
white238: yes

davidbeckingsale (approved):

cyrush (approved):

PR #584, opened 2022-04-20
Add utility macro for converting to system includes

adayton1:


cmake/BLTMacros.cmake
```diff
@@ -1401,3 +1401,21 @@ macro(blt_export_tpl_targets)
         endif()
     endforeach()
 endmacro()
+
+macro(blt_convert_to_system_includes)
```
white238, line 1405:
```suggestion
##------------------------------------------------------------------------------
## blt_convert_to_system_includes(TARGET <target>)
##
## Converts existing interface includes as system interface includes.
##------------------------------------------------------------------------------
macro(blt_convert_to_system_includes)
```

docs/api/utility.rst
```diff
@@ -167,3 +167,21 @@ command but doesn't throw an error if the list is empty or not defined.
...
+.. _blt_convert_to_system_includes:
+
+blt_convert_to_system_includes
+~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: cmake
+
+    blt_convert_to_system_includes(TARGET <target>)
+
+Converts the interface includes to system interface includes.
```
white238, line 179:
```suggestion
Converts existing interface includes to system interface includes.
```

cmake/BLTMacros.cmake
```diff
@@ -1401,3 +1401,21 @@ macro(blt_export_tpl_targets)
...
+    ## parse the arguments to the macro
+    cmake_parse_arguments(arg
+         "${options}" "${singleValuedArgs}" "${multiValuedArgs}" ${ARGN})
+
+    if(NOT DEFINED arg_TARGET)
+       message(FATAL_ERROR "TARGET is a required parameter for the blt_convert_to_system_includes macro.")
+    endif()
+
+    get_target_property(_include_dirs ${arg_TARGET} INTERFACE_INCLUDE_DIRECTORIES)
+    set_property(TARGET ${arg_TARGET} APPEND PROPERTY INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${_include_dirs}")
```
kennyweiss, line 1419: Should the macro remove the original includes before appending the system ones?

This might help users better debug their compilation lines.

kennyweiss (approved):
Thanks @adayton1 -- I like this!

Do you think it would make sense to extend this to take a regex to filter which includes get converted to system?

cmake/BLTMacros.cmake
adayton1: See my latest change and let me know if that's what you had in mind.

white238: Can you try this? I was under the impression the directories had to be duplicated in both places:

https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html

kennyweiss: Thanks @white238 -- Looks like the direct answer to my question is 
"No, we should not remove them"  
(thanks for digging into that).

On the other hand, combining [your link](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html) and [this page](https://cmake.org/cmake/help/latest/command/target_include_directories.html#command:target_include_directories) (linked to from your link might) seems to suggest that we might want to use `target_include_directories(SYSTEM)` in the proposed macro.

Having said that, since the current version works, that can be something we investigate in the future

white238: I've always set it manually... for example:

https://github.com/LLNL/serac/blob/10635cd995fa88ebbc58794aeab38c97c07e055b/cmake/thirdparty/SetupSeracThirdParty.cmake#L17-L20

Also I just confirmed that:

```
target_include_directories(foo SYSTEM PUBLIC /some/path)
blt_print_target_properties(TARGET foo)
```

does populate both `INTERFACE_INCLUDE_DIRECTORES` and `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES`. 

This is because the SYSTEM property marks the path for `-isystem` while `INCLUDE_DIRECTORIES` or `INTERFACE_INCLUDE_DIRECTORIES` tells it to use the path for this target or inherit.

adayton1:
> Thanks @adayton1 -- I like this!
> 
> Do you think it would make sense to extend this to take a regex to filter which includes get converted to system?

I don't have any use case for that, so I'm inclined to wait for someone to ask before going down that route.

adayton1:
@white238, @kennyweiss, is there a place where I can add a test for this macro? It looks like you do test some macros, but I can't see a clear pattern.

cmake/BLTMacros.cmake
adayton1: I changed it to `target_include_directories(foo SYSTEM **INTERFACE** /some/path)`

adayton1: Another question, do I need to recursively iterate through the INTERFACE_LINK_LIBRARIES and do the same thing on each of them?

white238: No.. this level is fine.

white238:
> @white238, @kennyweiss, is there a place where I can add a test for this macro? It looks like you do test some macros, but I can't see a clear pattern.

It's because there isnt a clear pattern.  This year I am going to revamp the BLT testing that grew into a mess.

cmake/BLTMacros.cmake
```diff
@@ -1401,3 +1401,26 @@ macro(blt_export_tpl_targets)
...
+    ## parse the arguments to the macro
+    cmake_parse_arguments(arg
+         "${options}" "${singleValuedArgs}" "${multiValuedArgs}" ${ARGN})
+
+    if(NOT DEFINED arg_TARGET)
+       message(FATAL_ERROR "TARGET is a required parameter for the blt_convert_to_system_includes macro.")
+    endif()
+
+    get_target_property(_include_dirs ${arg_TARGET} INTERFACE_INCLUDE_DIRECTORIES)
+    target_include_directories(${arg_TARGET} SYSTEM INTERFACE ${_include_dirs}))
```
white238, line 1424: This will duplicate the paths in the `INTERFACE_INCLUDE_DIRECTORIES`... sorry.

adayton1: OK, I've gone back to the original implementation.

white238 (approved):
Thanks @adayton1 !

white238:
Ack! don't merge until you add a release note about  this new macro!

PR #588, opened 2022-07-24
Improve error checking in `blt_list_append` macro

### LINKED ISSUE #587: `blt_list_append` macro should only require `ELEMENTS` to be present/valud when `IF` condition is satisfied
The `blt_list_append` macro requires its `ELEMENTS` parameter to be defined, but this should not be required in cases where nothing will be added.

I propose moving the check that `arg_ELEMENTS` is defined into the `if(${_shouldAdd})` clause.

Here is the entire macro:
https://github.com/LLNL/blt/blob/49971b88e631aba76a9990c7702ceee47a7c3216/cmake/BLTMacros.cmake#L52-L93

This tripped me up for the following use on axom's macos CI test:
```cmake
blt_list_append(TO var ELEMENTS ${CMAKE_BUILD_TYPES} IF CMAKE_BUILD_TYPES)
```
(apparently `CMAKE_BUILD_TYPES` is an empty string or undefined for that configuration)

kennyweiss:
Only check if `ELEMENTS` was provided and non-empty when appending, i.e. when `IF` condition evaluates to `TRUE`

Fixes #587.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -24,6 +24,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   CMAKE_CUDA_STANDARD if set otherwise falls back on BLT_CXX_STD or CMAKE_CXX_STANDARD.
 - Removed extra HIP offload flags that were being added as generator expressions as opposed to simple
   flags.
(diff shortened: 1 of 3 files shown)


white238 (approved):
Thanks @kennyweiss !

PR #590, opened 2022-08-03
Print target properties recursively and with regular expressions

chapman39:
Adds three more arguments to blt_print_target_properties
- CHILDREN (true/ false) whether or not you want to print the target's children's properties as well (recursively)
- PROPERTY_NAME_REGEX (regular expression string) reduce which properties to print by name
- PROPERTY_VALUE_REGEX (regular expression string) reduce which properties to print by value

Each target will only have its properties printed once

Also adds .gitignore file to ignore a VSCode file


cmake/BLTMacros.cmake
```diff
@@ -1274,96 +1274,92 @@ endmacro(blt_combine_static_libraries)
 
 
 ##------------------------------------------------------------------------------
-## blt_print_target_properties(TARGET <target> )
+## blt_print_target_properties(TARGET <target> CHILDREN <bool> PROP_REGEX <regular_expression_string>)
```
white238, line 1277:
```suggestion
## blt_print_target_properties(TARGET          <target name>
##                                             CHILDREN     <TRUE|FALSE>
##                                             PROP_REGEX <regular_expression_string>)
```

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,145 @@ macro(blt_clean_target)
...
+    unset(_is_cmake_target)
+endmacro(blt_print_target_properties_private)
+
+##------------------------------------------------------------------------------
+## blt_find_all_targets_recursive(TARGET <target> TLIST tlist)
+##
+## Store all target's dependancies (link libraries and interface link libraries)
+## recursively in TLIST.
+##
+## NOTE: TLIST should be an empty string at the start. TLIST must be called tlist
```
white238, line 935: Can you remove this requirement?

`if(NOT DEFINED tlist)` -> `if(NOT DEFINED ${arg_TLIST})`
`list(APPEND tlist ${t})` -> `list(APPEND ${arg_TLIST} ${t})`

cmake/BLTMacros.cmake
```diff
@@ -1274,96 +1274,92 @@ endmacro(blt_combine_static_libraries)
...
-## Prints out all properties of the given target.
+## Prints out properties of the given target.
+##
+## Has the options to print target's link libraries and interface link libraries
+## with the CHILDREN argument, as well as specific properties using a regular
+## expression.
+## 
+## Defaults:
+## CHILDREN = false (non recursive)
+## PROP_REGEX = .* (print every property)
```
white238, line 1287: Can you change this to default to an empty string and skip the regex matching completely if its empty?  Regex's are heavy handed and should be avoided if possible.  Not that speed super matters in this case but we should still optimize when its easy.

white238:
@chapman39 don't forget to update the docs here:  docs/api/target_properties.rst

As well as adding a message in the RELEASE-NOTES.md

cmake/BLTPrivateMacros.cmake
chapman39: yes that sounds much better

cmake/BLTMacros.cmake
white238: I did several tests and it ended up being exactly the same.. so you can leave it in.

chapman39:
Ready for review

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,149 @@ macro(blt_clean_target)
...
+    if(NOT DEFINED arg_TARGET)
+        message(FATAL_ERROR "TARGET is a required parameter for the blt_find_all_targets_recursive macro")
+    endif()
+
+    if(NOT DEFINED ${arg_TLIST})
+        message(FATAL_ERROR "TLIST is a required parameter for the blt_find_all_targets_recursive macro")
+    endif()
+
+    # check if this is a valid cmake target
+    if(TARGET ${arg_TARGET})
```
white238, line 962: Add a case here to check if it is a BLT registered library and check the following property for dependencies:

for registered library named `core`, it would be `_BLT_CORE_DEPENDS_ON`

white238:
This currently allows filtering based on property names.  For example, `INCLUDE` will print both `INTERFACE_INCLUDE_DIRECTORIES` and `INCLUDE_DIRECTORIES`.  @chapman39 is going to add the ability to filter based on property values as well.

@kennyweiss , do you like which set of parameter names:

* `PROP_NAME_REGEX` and `PROP_VALUE_REGEX`
* `PROPERTY_NAME_REGEX` and `PROPERTY_VALUE_REGEX`

The latter is verbose but at least I wont have to remember what word we truncated. Thoughts?

cmake/BLTPrivateMacros.cmake
white238: This should expose the problem:

```
blt_register_library(NAME bar)
blt_register_library(NAME foo DEPENDS_ON BAR)
blt_print_target_properties(TARGET foo CHILDREN TRUE)
```

chapman39:
longer argument names, now can filter by property value.

chapman39:
decided to keep some extra code in `tests/internal` for the `blt_registered` case. if i should remove it, or replace with less generic names (foo and bar) lmk.

cmake/BLTPrivateMacros.cmake
chapman39: fixed

docs/api/target_properties.rst
```diff
@@ -155,13 +155,21 @@ blt_print_target_properties
...
 
+CHILDREN
+  Whether or not to print the properties of the target's children (false by default)
+
+PROP_REGEX
+  Limit the properties to print (all by default)
+
 The given target must be added via ``add_executable()`` or ``add_library()`` or
 with the corresponding :ref:`blt_add_executable`, :ref:`blt_add_library`, 
 :ref:`blt_import_library`, or :ref:`blt_register_library` macros.
```
kennyweiss, line 175: Do we include examples for other macros? If so, an example of how this is used might be helpful.

tests/internal/CMakeLists.txt
```diff
@@ -260,9 +260,14 @@ endif()
...
 message(STATUS "")
 foreach(_target gtest example t_example_smoke not-a-target blt_header_only mpi cuda cuda_runtime blt::hip blt::hip_runtime)
-    blt_print_target_properties(TARGET ${_target})
+    blt_print_target_properties(TARGET ${_target} CHILDREN true)
 endforeach()
 
+message(STATUS "Exercising blt_print_target_properties macro on some blt_registered target.")
+blt_register_library(NAME foo)
+blt_register_library(NAME bar DEPENDS_ON foo)
+blt_print_target_properties(TARGET bar CHILDREN true)
```
kennyweiss, line 269: It would be helpful to have some examples that exercise the other options, e.g. with `CHILDREN == false` and with different property regex values.

docs/api/target_properties.rst
```diff
@@ -155,13 +155,21 @@ blt_print_target_properties
...
 Prints out all properties of the given target.
 
 TARGET
   Name of CMake target
 
+CHILDREN
+  Whether or not to print the properties of the target's children (false by default)
+
+PROP_REGEX
+  Limit the properties to print (all by default)
```
kennyweiss, line 171: Before merging, please ensure that the arguments in the docs match those of the actual macro in the code.

cmake/BLTMacros.cmake
```diff
@@ -1274,96 +1274,92 @@ endmacro(blt_combine_static_libraries)
...
-        ## For interface targets, filter against whitelist of valid properties
-        get_property(_target_type TARGET ${arg_TARGET} PROPERTY TYPE)
-        if("${_target_type}" STREQUAL "INTERFACE_LIBRARY")
-            blt_filter_list(TO _property_list
-                            REGEX "^(INTERFACE_|IMPORTED_LIBNAME_|COMPATIBLE_INTERFACE_|MAP_IMPORTED_CONFIG_)|^(NAME|TYPE|EXPORT_NAME)$"
-                            OPERATION "include")
+    else()
+        # print properties
+        blt_print_target_properties_private(TARGET ${arg_TARGET}
+                                            PROPERTY_NAME_REGEX ${arg_PROPERTY_NAME_REGEX})
```
kennyweiss, line 1340: Note: The macro doesn't appear to use `PROPERTY_VALUE_REGEX`.

I'm guess this will be added in the future?

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,156 @@ macro(blt_clean_target)
...
+endmacro(blt_print_target_properties_private)
+
+##------------------------------------------------------------------------------
+## blt_find_all_targets_recursive(TARGET <target> TLIST tlist)
+##
+## Store all target's dependancies (link libraries and interface link libraries)
+## recursively in TLIST.
+##
+##------------------------------------------------------------------------------
+macro(blt_find_all_targets_recursive)
```
kennyweiss, line 937: Suggestion (clarity): Should this be called `blt_find_target_dependencies` ?

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,156 @@ macro(blt_clean_target)
...
+
+    unset(_target_upper)
+    unset(_is_blt_registered_target)
+    unset(_is_cmake_target)
+endmacro(blt_print_target_properties_private)
+
+##------------------------------------------------------------------------------
+## blt_find_all_targets_recursive(TARGET <target> TLIST tlist)
+##
+## Store all target's dependancies (link libraries and interface link libraries)
```
kennyweiss, line 933: Possible typo?
```suggestion
## Store all target's dependencies (link libraries and interface link libraries)
```

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,156 @@ macro(blt_clean_target)
...
+    string(TOUPPER ${arg_TARGET} _target_upper)
+    if(_BLT_${_target_upper}_IS_REGISTERED_LIBRARY)
+        # recursive call
+        set (_depends_on "${_BLT_${_target_upper}_DEPENDS_ON}")
+        foreach(t ${_depends_on})
+            list(APPEND ${arg_TLIST} ${t})
+            blt_find_all_targets_recursive(TARGET ${t} TLIST ${arg_TLIST})
+        endforeach()
+        unset(_depends_on)
+    endif()
```
kennyweiss, line 966: Should we be concerned about 
(a) infinite recursion, e.g. if there is a cycle in the dependencies, or 
(b) repeated dependencies -- e.g. we have a DAG of dependencies, and this function, as written will find the dependents of each of the dependents

It might be better to only call `blt_find_all_targets_recursive` once for each potential dependent, e.g. by checking against the list of targets that we've already checked.

cmake/BLTMacros.cmake
```diff
@@ -1274,96 +1274,92 @@ endmacro(blt_combine_static_libraries)
...
+
+        if(${arg_CHILDREN})
+            # find all targets from dependency tree
+            set(tlist "")
+            blt_find_all_targets_recursive(TARGET ${arg_TARGET} TLIST tlist)
+            blt_list_remove_duplicates(TO tlist)
```
kennyweiss, line 1346: 👍

kennyweiss (changes requested):
Thanks @chapman39 -- this is coming along nicely.

Some observations:
* Please add some checks to avoid infinite loops and excessive computations (see my comment)
* Please ensure that the arguments in the docs match those in the code. 
* It looks like some of the regex's are not being passed down to the private macro. Please fix this
* It would also be nice to have some examples of how to use the regex options. Possibly in the docs, but definitely in the unit tests.

kennyweiss:
> @kennyweiss , do you like which set of parameter names:
>     * `PROP_NAME_REGEX` and `PROP_VALUE_REGEX`
>     * `PROPERTY_NAME_REGEX` and `PROPERTY_VALUE_REGEX`
> The latter is verbose but at least I wont have to remember what word we truncated. Thoughts?

I more verbose options seem clearer to me -- `PROPERTY_NAME_REGEX` and `PROPERTY_VALUE_REGEX`

cmake/BLTMacros.cmake
chapman39: yes, this was a mistake. this mistake actually revealed more mistakes in `blt_print_target_properties_private`:

- my regex `MATCHES` was checking `_propval` before `get_target_property` ran in the cmake target section
- and i had`_propval` in the blt registered target section (when it should be `${${prop}}`)

i've fixed these and they'll be up on github soon.

cmake/BLTPrivateMacros.cmake
chapman39: After `blt_find_target_dependencies` is called, `blt_list_remove_duplicates` is called on the list (in the public `blt_print_target_properties`), so (b) should not be an issue. I'll test infinite recursion

chapman39: Yes when I do

```
blt_register_library(NAME foo DEPENDS_ON bar)
blt_register_library(NAME bar DEPENDS_ON foo)
blt_print_target_properties(TARGET bar CHILDREN true)
```

There is an infinite loop. I can try an solve this issue

chapman39: I now check if a dependency is already in the list before adding it to `TLIST`.

RELEASE-NOTES.md
```diff
@@ -12,6 +12,10 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 ### Added
 - Added 'blt_convert_to_system_includes' macro to convert existing interface includes to system interface includes.
 
+### Changed
+- Added extra options to 'blt_print_target_properties' macro to print properties of
+target's children as well as limit the properties printed with a regular expression.
```
white238, line 17: Add what the new parameter names are here please.

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,168 @@ macro(blt_clean_target)
...
+    unset(_target_upper)
+    unset(_is_blt_registered_target)
+    unset(_is_cmake_target)
+endmacro(blt_print_target_properties_private)
+
+##------------------------------------------------------------------------------
+## blt_find_target_dependencies(TARGET <target> TLIST tlist)
+##
+## Store all target's dependencies (link libraries and interface link libraries)
+## recursively in TLIST.
```
bmhan12, line 944:
```suggestion
## recursively in tlist.
```

docs/api/target_properties.rst
```diff
@@ -155,12 +155,24 @@ blt_print_target_properties
...
+                                PROPERTY_VALUE_REGEX <regular_expression_string>)
 
 Prints out all properties of the given target.
 
 TARGET
-  Name of CMake target
+  Name of CMake target (required)
+
+CHILDREN
+  Whether or not to print the properties of the target's children (false by default)
```
bmhan12, line 169: Indicate that this will be recursive when true.

docs/api/target_properties.rst
```diff
@@ -169,6 +181,14 @@ with the corresponding :ref:`blt_add_executable`, :ref:`blt_add_library`,
...
  | [<target> property] <property>: <value>
 
+.. code-block:: cmake
+    :caption: **Example**
+    :linenos:
+
+    blt_print_target_properties(TARGET foo)
+    blt_print_target_properties(TARGET foo CHILDREN TRUE)
+    blt_print_target_properties(TARGET foo CHILDREN TRUE PROPERTY_NAME_REGEX "CXX")
+    blt_print_target_properties(TARGET foo PROPERTY_NAME_REGEX "CXX" PROPERTY_VALUE_REGEX "ON")
```
bmhan12, line 191: @white238  Can you activate this branch on ReadTheDocs to verify the formatting is alright?

bmhan12:
Also: Looks like the Mac Image version will need to be updated: https://dev.azure.com/llnl-blt/BLT/_build/results?buildId=1328&view=logs&j=8dab7c2a-824f-53c4-ff36-17cdfa82e5d3

https://github.com/LLNL/blt/blob/7b3b13dd55eb847ec2d17c87adfe16c872d634e8/azure-pipelines.yml#L53-L55

white238:
> Also: Looks like the Mac Image version will need to be updated: https://dev.azure.com/llnl-blt/BLT/_build/results?buildId=1328&view=logs&j=8dab7c2a-824f-53c4-ff36-17cdfa82e5d3
> 
> https://github.com/LLNL/blt/blob/7b3b13dd55eb847ec2d17c87adfe16c872d634e8/azure-pipelines.yml#L53-L55

Well i was originally going to say address this after but its blocking the builds. @chapman39 can you fix this before merging?

cmake/BLTPrivateMacros.cmake
chapman39: i think it should still be capital TLIST but ill change the wording to explain TLIST holds the name of the variable that has the list.

so its like:
- arg_TLIST = "my_list"
- ${arg_TLIST} = my_list = "target1;target2;target3"

chapman39:
Passing osx_gcc, but now I'm failing three of the gitlab tests since it's saying I'm an unauthorized user (or maybe it's referring to a service user?)

white238:
> Passing osx_gcc, but now I'm failing three of the gitlab tests since it's saying I'm an unauthorized user (or maybe it's referring to a service user?)

Not sure what the problem was. I retried and it passed.

RELEASE-NOTES.md
```diff
@@ -10,7 +10,14 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 ## [Unreleased] - Release date yyyy-mm-dd
 
 ### Added
-- Added 'blt_convert_to_system_includes' macro to convert existing interface includes to system interface includes.
+- Added `blt_convert_to_system_includes` macro to convert existing interface includes to system interface includes.
+
+### Changed
+- Added three extra options to `blt_print_target_properties` macro to print properties of
+target's children as well as limit the properties printed with regular expressions:
```
white238, line 17:
```suggestion
  target's children as well as limit the properties printed with regular expressions:
```

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,166 @@ macro(blt_clean_target)
...
+    ## parse the arguments to the macro
+    cmake_parse_arguments(arg
+         "${options}" "${singleValuedArgs}" "${multiValuedArgs}" ${ARGN})
+
+    ## check if this is a valid cmake target or blt_registered target
+    set(_is_cmake_target FALSE)
+    set(_target_type_str "")
+    if(TARGET ${arg_TARGET})
+        set(_is_cmake_target TRUE)
+        set(_target_type_str "${_target_type_str}cmake target")
```
white238, line 869:
```suggestion
        set(_target_type_str "CMake target")
```

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,166 @@ macro(blt_clean_target)
...
+    endif()
+
+    set(_is_blt_registered_target FALSE)
+    string(TOUPPER ${arg_TARGET} _target_upper)
+    if(_BLT_${_target_upper}_IS_REGISTERED_LIBRARY)
+        set(_is_blt_registered_target TRUE)
+        if (_is_cmake_target)
+            set(_target_type_str "${_target_type_str} and a ")
+        endif()
+        set(_target_type_str "${_target_type_str}blt_registered target")
```
white238, line 879:
```suggestion
        set(_target_type_str "${_target_type_str}BLT Registered target")
```

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,166 @@ macro(blt_clean_target)
...
+    endif()
+
+    if (_is_cmake_target OR _is_blt_registered_target)
+        message(STATUS "[${arg_TARGET} property] '${arg_TARGET}' is a ${_target_type_str}")
```
white238, line 883:
```suggestion
        message(STATUS "[${arg_TARGET}] '${arg_TARGET}' is a ${_target_type_str}")
```

cmake/BLTPrivateMacros.cmake
```diff
@@ -842,3 +842,166 @@ macro(blt_clean_target)
...
+##------------------------------------------------------------------------------
+## blt_find_target_dependencies(TARGET <target> TLIST <list name>)
+##
+## Store all target's dependencies (link libraries and interface link libraries)
+## recursively in the variable name TLIST holds.
+##------------------------------------------------------------------------------
+macro(blt_find_target_dependencies)
+
+    set(options)
+    set(singleValuedArgs TARGET TLIST CUR_TARGET)
```
white238, line 949:
```suggestion
    set(singleValuedArgs TARGET TLIST)
```

docs/api/target_properties.rst
```diff
@@ -155,12 +155,24 @@ blt_print_target_properties
 
 .. code-block:: cmake
 
-    blt_print_target_properties(TARGET <target>)
+    blt_print_target_properties(TARGET               <target name>
+                                CHILDREN             <TRUE|FALSE>
+                                PROPERTY_NAME_REGEX  <regular_expression_string>
+                                PROPERTY_VALUE_REGEX <regular_expression_string>)
 
 Prints out all properties of the given target.
```
white238, line 163:
```suggestion
Prints out all properties of the given target. Can also print target's dependencies with the 
`CHILDREN` argument, as well as specific filter properties using regular expressions with 
`PROPERTY_NAME_REGEX` and `PROPERTY_VALUE_REGEX`.
```

cmake/BLTPrivateMacros.cmake
chapman39: thanks for catching this. i was planning to use the variable to prevent the infinite loop issue but never ended up needing it- guess i forgot to remove it.

docs/api/target_properties.rst
chapman39: I feel like I already explain what the arguments are [below that line](https://github.com/LLNL/blt/pull/590/files/401f967a4e0e87b83ba09b914cbd88fdddbeefc9#diff-b74b6950ded9a2c4317c89df885955d3a1907c5dafbdee8fc288c1029cf24a5cR168-R175)

white238: well "of a given target" is not entirely accurate anymore for a brief description.

white238: It could also just be something like:

Prints out all, or filtered, properties of a given target with optionally printing dependencies.

white238 (approved):
Thanks @chapman39 ! This is a huge improvement!

tests/internal/CMakeLists.txt
```diff
@@ -257,20 +257,38 @@ if (ENABLE_CUDA AND ENABLE_MPI AND
...
+message(STATUS "Exercising blt_print_target_properties macro on dummy blt_registered target.")
+blt_register_library(NAME foo)
+blt_register_library(NAME bar DEPENDS_ON foo)
+blt_print_target_properties(TARGET bar CHILDREN true)
+
+message(STATUS "")
+message(STATUS "Exercising blt_print_target_properties macro. Testing infinite recursion")
+blt_register_library(NAME foo DEPENDS_ON bar)
+blt_register_library(NAME bar DEPENDS_ON foo)
+blt_print_target_properties(TARGET bar CHILDREN true)
```
kennyweiss, line 282: 👍

kennyweiss (approved):
Thanks @chapman39 -- I'm looking forward to using this!

PR #596, opened 2022-08-30
Feature/probinso/hipcc support

robinson96:
Add initial support for using CMAKE_HIP_COMPILER in cmake versions 3.21 and greater.

Differentiate between C_LIST and CXX_LIST for blt_split_source_list_by_language. (HIPCC will error if it's a c file, while cuda will just treat it as a C++ file).

SetupHIP now enables the hip language.

Add option BLT_HIP_REMOVE_HIP_COMPILE_OPTIONS to remove flags associated with pre-installed hip::device and hip-lang::device targets that are incompatible with hipcc.

### DIFF
--- a/SetupBLT.cmake
+++ b/SetupBLT.cmake
@@ -187,7 +187,9 @@ if (NOT BLT_LOADED)
     # File extension lists used to filter sources based on languages for code checks
     # and filtering Fortran sources out of cuda/hip source lists
     # Note: this filtering is case-insensitive
-    set(BLT_C_FILE_EXTS ".cpp" ".hpp" ".cxx" ".hxx" ".c" ".h" ".cc" ".hh" ".inl" ".cu" ".cuh"
+    set(BLT_C_FILE_EXTS ".c" ".h"
+               CACHE STRING "List of known file extensions used for C/CXX sources")
+    set(BLT_CXX_FILE_EXTS ".cpp" ".hpp" ".cxx" ".hxx"  ".cc" ".hh" ".inl" ".cu" ".cuh"
                CACHE STRING "List of known file extensions used for C/CXX sources")
     set(BLT_Fortran_FILE_EXTS ".f" ".f90"
                CACHE STRING "List of known file extensions used for Fortran sources")
--- a/cmake/BLTPrivateMacros.cmake
+++ b/cmake/BLTPrivateMacros.cmake
@@ -491,6 +491,87 @@ macro(blt_setup_target)
 
 endmacro(blt_setup_target)
 
+##------------------------------------------------------------------------------
+## blt_setup_hip_target(NAME <name of target>
+##                       SOURCES <list of sources>
+##                       DEPENDS_ON <list of dependencies>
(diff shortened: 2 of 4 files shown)


SetupBLT.cmake
```diff
@@ -187,7 +187,9 @@ if (NOT BLT_LOADED)
     # File extension lists used to filter sources based on languages for code checks
     # and filtering Fortran sources out of cuda/hip source lists
     # Note: this filtering is case-insensitive
-    set(BLT_C_FILE_EXTS ".cpp" ".hpp" ".cxx" ".hxx" ".c" ".h" ".cc" ".hh" ".inl" ".cu" ".cuh"
+    set(BLT_C_FILE_EXTS ".c" ".h"
+               CACHE STRING "List of known file extensions used for C/CXX sources")
```
adayton1, line 191: Update the comment to say C rather than C/CXX. And the following comment should have CXX instead of C/CXX.

cmake/BLTPrivateMacros.cmake
```diff
@@ -491,6 +491,87 @@ macro(blt_setup_target)
...
+    set(_depends_on_hip FALSE)
+    if(${_hip_index} GREATER -1)
+        set(_depends_on_hip TRUE)
+    endif()
+    list(FIND arg_DEPENDS_ON "blt::hip_runtime" _hip_runtime_index)
+    set(_depends_on_hip_runtime FALSE)
+    if(${_hip_runtime_index} GREATER -1)
+        set(_depends_on_hip_runtime TRUE)
+    endif()
+    message("DEPENDS ON HIP ${_depends_on_hip} DEPENDS_ON_HIP_RUNTIME ${_depends_on_hip_runtime}\n\t ${arg_DEPENDS_ON}")
```
adayton1, line 530: Is this a debug message? There are a few others like it as well.

cmake/BLTPrivateMacros.cmake
```diff
@@ -491,6 +491,87 @@ macro(blt_setup_target)
...
+        set(_non_hip_sources)
+        blt_split_source_list_by_language(SOURCES      ${arg_SOURCES}
+                                          CXX_LIST       _hip_sources
+                                          C_LIST       _non_hip_sources
+                                          Fortran_LIST _non_hip_sources)
+        message("ALL_SOURCES: ${arg_SOURCES}\n\t HIP SOURCES ${_hip_sources}")
+        set_source_files_properties( ${_hip_sources} PROPERTIES
+                                     LANGUAGE HIP)
+
+#        if (HIP_SEPARABLE_COMPILATION)
```
adayton1, line 554: Can you document why this isn't being used?

SetupBLT.cmake
white238: This will have pretty large user facing consequences and needs to be considered carefully.

white238: Might only be my problem though:

https://sourcegraph.com/search?q=context:global+repo:github.com/llnl/*+BLT_C_FILE_EXTS&patternType=standard

adayton1:
Closing in favor of #702.

PR #599, opened 2022-09-30
Explicitly set cxx standard for GNU compiler

mcfadden8:
Newer versions of the GNU compiler uses a may use newer c++ standards which could cause cmake compilation tests to yeild false-positive results when it says that a c++17 feature exists and is usable by when the target/using is actually building c++14 targets.

The problem caused the `check_cxx_source_compiles` check (https://github.com/LLNL/Umpire/blob/develop/cmake/SetupCompilerFlags.cmake#L14-L27) to produce a false-positve because it was using running `g++` without any `-std=c++14` argument.  The default c++ standard for 11.2.0 g++ is 17.  So, this check would succeed and the Umpire source code would fail to compile later since it was using `-std=c++14` 

~~NOTE: This PR contains one fix and one hack.  The hack needs to be resolved before this may be considered for merging.~~

NOTE 2: I've attempted to address the hack mentioned above by separating out the initialization of the compiler family variables to a different (new) file and am including that file earlier in BLTMacros.cmake.

This problem was reported by a spack user [here](https://github.com/spack/spack/pull/32504#issuecomment-1255601886) and is internally documented in an Umpire jira ticket [here](https://rzlc.llnl.gov/jira/browse/UM-1017).


mcfadden8:
Closing this since it will be addressed by https://github.com/LLNL/blt/pull/600

PR #613, opened 2023-01-10
Patch include doubling

white238:
`blt_patch_target()` was not only doubling the `INTERFACE_INCLUDE_DIRECTORIES` target property but also adding it to `INCLUDE_DIRECTORIES`. This doesn't make sense in any case.

Example:

```
add_library(foo)
target_include_directories(foo INTERFACE /test/path)
blt_patch_target(NAME foo TREAT_INCLUDES_AS_SYSTEM true)
```

Before:

```
-- [foo property] 'foo' is a CMake target
-- [foo property] INCLUDE_DIRECTORIES: /usr/workspace/white238/
-- [foo property] INTERFACE_INCLUDE_DIRECTORIES: /usr/workspace/white238/;/usr/workspace/white238/
-- [foo property] INTERFACE_SYSTEM_INCLUDE_DIRECTORIES: /usr/workspace/white238/
```

After:

```
-- [foo property] 'foo' is a CMake target
-- [foo property] INTERFACE_INCLUDE_DIRECTORIES: /usr/workspace/white238/
-- [foo property] INTERFACE_SYSTEM_INCLUDE_DIRECTORIES: /usr/workspace/white238/
```

Also there was some duplicated logic between `blt_patch_target()` and `blt_convert_to_system_includes()`.  I made `blt_patch_target()` call into `blt_convert_to_system_includes()`.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -1541,13 +1541,13 @@ macro(blt_convert_to_system_includes)
 
     # PGI does not support -isystem
     if(NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
-        get_target_property(_include_dirs ${arg_NAME} INTERFACE_INCLUDE_DIRECTORIES)
+        get_target_property(_include_dirs ${arg_TARGET} INTERFACE_INCLUDE_DIRECTORIES)
         # Don't copy if the target had no include directories
         if(_include_dirs)
             # Clear previous value in INTERFACE_INCLUDE_DIRECTORIES so it is not doubled
             # by target_include_directories
-            set_property(TARGET ${arg_NAME} PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
-            target_include_directories(${arg_NAME} SYSTEM INTERFACE ${_include_dirs})
+            set_property(TARGET ${arg_TARGET} PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
+            target_include_directories(${arg_TARGET} SYSTEM INTERFACE ${_include_dirs})
         endif()
     endif()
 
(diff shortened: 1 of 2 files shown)


cmake/BLTMacros.cmake
```diff
@@ -480,18 +480,8 @@ macro(blt_patch_target)
         endif()
     endif()
 
-    # PGI does not support -isystem
-    if( (${arg_TREAT_INCLUDES_AS_SYSTEM}) AND (NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI"))
-        get_target_property(_target_includes ${arg_NAME} INTERFACE_INCLUDE_DIRECTORIES)
-        # Don't copy if the target had no include directories
-        if(_target_includes)
-            if(_standard_lib_interface)
-                target_include_directories(${arg_NAME} SYSTEM ${_scope} ${_target_includes})
```
white238, line 489: The scope here doesn't make sense and so was removed.  `PUBLIC` is  `INTERFACE` (inheriting to downstream targets) and `PRIVATE` (used when building itself). The latter doesn't make sense for system includes.

cmake/BLTMacros.cmake
```diff
@@ -1548,8 +1539,18 @@ macro(blt_convert_to_system_includes)
        message(FATAL_ERROR "TARGET is a required parameter for the blt_convert_to_system_includes macro.")
     endif()
```
kennyweiss, line 1541: Should we add a check here to make sure that the target is not `IMPORTED`?

In https://github.com/LLNL/axom/pull/1017/commits/32ce9f28e95e847d553d08a0ad8a076917e17313, I had a workaround based on an error about that.

I suspect it's not necessary since I converted the code there to `blt_convert_to_system_includes` and no longer saw the error, but wanted to double check with you.

kennyweiss (approved):
Thanks @white238 !

cmake/BLTMacros.cmake
white238: My guess is the `target_include_directories` was causing that error. I say leave it until it causes another error.

PR #622, opened 2023-02-15
Remove tracking all project sources

white238:
This was done for historical reasons but is no longer used.


cyrush (approved):

kennyweiss (approved):
Thanks @white238 

Would it make sense to keep this as a macro, but make it opt in, e.g. via a debugging config variable? I can imagine it helping w/ debugging in some cases.

PR #625, opened 2023-03-07
Use regex to find python and cmake sources (small fix)

### LINKED ISSUE #624: Error when blt_add_code_checks calls blt_add_yapf_target
Having this issue in LiDO where when I call `blt_add_code_checks` I get the following error:

```
CMake Error at cmake/blt/cmake/SetupCodeChecks.cmake:873 (message):
  blt_add_yapf_target requires a SRC_FILES parameter
  cmake/blt/cmake/SetupCodeChecks.cmake:246 (blt_add_yapf_target)
  CMakeLists.txt:187 (blt_add_code_checks)
```

The `SRC_FILES` variable is set to `_py_sources` (retrieved from `blt_split_source_list_by_language`, see https://github.com/LLNL/blt/blob/develop/cmake/SetupCodeChecks.cmake#L156).

I guess when `blt_split_source_list_by_language` is called, `Python_LIST` is empty, which means `_py_sources` is empty, which means `SRC_FILES` is `""` when it gets down to `blt_add_yapf_target`.

When I commented out the fatal error I was able to make by LiDO build without issues. Maybe this check is not required?

chapman39:
After updating BLT on LiDO I realized `blt_split_source_list_by_language` was using EXT variables when is should have been using REGEX.

Fixes #624


white238 (approved):
Good catch! thanks @chapman39

PR #628, opened 2023-04-13
Update Docker images used by Azure Pipelines

white238:


rhornung67 (approved):
@white238 are pulling base images from the rse-ops project and building on top of those?

white238:
> @white238 are pulling base images from the rse-ops project and building on top of those?

This is merely moving to the same images that Axom is using and is not using the rse-ops ones. I will make an effort the move Axom and BLT forward soon. This PR is unblocking development for BLT.

Issue #635, opened 2023-05-17
blt_add_benchmark should take CONFIGURATIONS and OMP_NUM_THREADS arguments


arghdos:
E.g., here:

https://github.com/LLNL/blt/blob/553a90261dbe0e2c064330d9a129fa084267a69c/cmake/BLTMacros.cmake#L937

Not only is this consistent with `blt_add_test`, but this enables e.g., creating your own configurations that specific benchmarks should be run on.  For instance, I would like to create a "Multinode" configurations where only tests / benchmarks that require multiple nodes should be run (e.g., to benchmark inter-node bandwidth).  This way I can create a Jenkins job or the like that _only_ runs those tests when I allocate two+ nodes for a job.

Today, I have to do something like:

```
blt_add_test(...
                    CONFIGURATIONS Benchmark Multinode
)
```

I think this would be a minimally invasive PR, and I'd be happy to work on it if y'all agree.

arghdos:
Looking at it, it appears that NUM_OMP_THREADS is missing as well

white238:
That's a great idea. We appreciate any help if you have time, otherwise I will handle it shortly.

arghdos:
started in #636

white238:
Closed in #636

PR #636, opened 2023-05-17
Add OMP_NUM_THREADS and CONFIGURATIONS to blt_add_benchmarks (#635)

arghdos:


arghdos:
cc: @white238 -- I am not sure what adding a test for this would look like (not super familiar with BLT's test infra).  Feel free to point me in the right direction and I'll add something.

docs/api/target.rst
```diff
@@ -28,6 +30,14 @@ COMMAND
 NUM_MPI_TASKS
   Indicates this is an MPI test and how many MPI tasks to use.
 
+NUM_OMP_THREADS
+  Indicates this test requires the defined environment variable ``OMP_NUM_THREADS``
+  set to the given variable.
+
+CONFIGURATIONS
+  Set the CTest configuration for this test.  When not specified, the test
+  will be added to BLT's default ``Benchmark`` CTest configuration.
```
kennyweiss, line 39: The code is:
```cmake
CONFIGURATIONS Benchmark ${arg_CONFIGURATIONS})
```
If I'm understanding it correctly, the docs are a bit misleading since it is always adding the test to the `Benchmark` configuration.

Perhaps
```suggestion
  Optionally set additional CTest configuration(s) for this test. Benchmark tests
  will always be added to the ``Benchmark`` CTest configuration.
```

kennyweiss (approved):
Thanks @arghdos !

docs/api/target.rst
white238: I'll fix this in the other PR after I have fixed the CI due to the Toss4 conversion.

white238:
Github should have closed this but the last commit confused it. This was all inside the #637 PR.

Thanks for the contribution!

PR #644, opened 2023-06-17
Adds a `clang_tidy_style` target to allow clang-tidy to apply fixes inline

kennyweiss:
Note: This feature relies on the `clang-apply-replacements` executable, which should be
provided via a CMake variable: `CLANGAPPLYREPLACEMENTS_EXECUTABLE`.

### DIFF
--- a/cmake/SetupCodeChecks.cmake
+++ b/cmake/SetupCodeChecks.cmake
@@ -87,10 +87,10 @@ if(CLANGTIDY_FOUND)
     add_custom_target(clang_tidy_check)
     add_dependencies(${BLT_CODE_CHECK_TARGET_NAME} clang_tidy_check)
 
-    # targets for modifying code based via clang-tidy
+    # targets for modifying code based via clang-tidy; also require clang-apply-replacements
     if(CLANGAPPLYREPLACEMENTS_FOUND)
         add_custom_target(clang_tidy_style)
-        add_dependencies(${BLT_CODE_CHECK_TARGET_NAME} clang_tidy_style)
+        add_dependencies(${BLT_CODE_STYLE_TARGET_NAME} clang_tidy_style)
     endif()
 endif()
 
@@ -481,7 +481,11 @@ macro(blt_add_clang_tidy_target)
         endif()
 
         # hook our new target into the proper dependency chain
-        add_dependencies(clang_tidy_check ${arg_NAME})
+        if(arg_FIX)
+            add_dependencies(clang_tidy_style ${arg_NAME})
+        else()
+            add_dependencies(clang_tidy_check ${arg_NAME})
+        endif()
 
         # Code check targets should only be run on demand
         set_property(TARGET ${arg_NAME} PROPERTY EXCLUDE_FROM_ALL TRUE)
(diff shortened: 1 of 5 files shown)


rhornung67 (approved):
Thanks @kennyweiss. When we have access to clang 15 compilers, we won't need this, correct?

cmake/SetupCodeChecks.cmake
```diff
@@ -83,15 +83,22 @@ if(CLANGQUERY_FOUND)
...
 
 if(CLANGTIDY_FOUND)
+    # targets for verifying clang-tidy
     add_custom_target(clang_tidy_check)
     add_dependencies(${BLT_CODE_CHECK_TARGET_NAME} clang_tidy_check)
+
+    # targets for modifying code based via clang-tidy; also require clang-apply-replacements
+    if(CLANGAPPLYREPLACEMENTS_FOUND)
+        add_custom_target(clang_tidy_style)
+        add_dependencies(${BLT_CODE_STYLE_TARGET_NAME} clang_tidy_style)
```
kennyweiss, line 93: I think I should remove this line.

The changes from `clang_tidy_style` are not always safe, so it should not be added to the `style` target.
Rather, it should be opt-in, i.e. users would need to explicitly run `clang_tidy_style`.

kennyweiss:
> Thanks @kennyweiss. When we have access to clang 15 compilers, we won't need this, correct?

While I'm adding this so axom to add braces around one-line if/while/for statements (via the `readability-braces-around-statements` rule), I think it is generally useful to help blt users apply clang-tidy fixes.

cmake/SetupCodeChecks.cmake
white238: I think that sounds good. We should document it somewhere though, users will expect it to be under the `style` target.

cmake/SetupCodeChecks.cmake
```diff
@@ -83,15 +83,22 @@ if(CLANGQUERY_FOUND)
 endif()
 
 if(CLANGTIDY_FOUND)
+    # targets for verifying clang-tidy
     add_custom_target(clang_tidy_check)
     add_dependencies(${BLT_CODE_CHECK_TARGET_NAME} clang_tidy_check)
```
kennyweiss, line 88: Also wondering if `clang-tidy-check` should be run whenever we run `make check` since it can have a lot of false positives if you're not careful about reducing rules and possibly adding some `NOLINT*` lines to your code.

See: https://clang.llvm.org/extra/clang-tidy/index.html#suppressing-undesired-diagnostics

cmake/SetupCodeChecks.cmake
kennyweiss: We discussed this in an axom meeting on 6/26/2023 and decided that clang-tidy targets should not be added to `make check` and `make style` by default, and that we should document this.

cmake/SetupCodeChecks.cmake
kennyweiss: I've removed `clang_tidy_check` from the main `check` target, per group discussion. See comments elsewhere on this PR.

white238 (approved):

PR #647, opened 2023-07-05
Harden finding ROCM_PATH

white238:
A recent Spack change to `CachedCMakePackage` now adds a lot of paths to the `CMAKE_PREFIX_PATH`. This exposed that this `find_path` call was not as robust as it probably could be.

The failure was in Axom, where a combination of paths that seemed fine would confuse this `find_path` call to end up with `ROCM_PATH` set to `/opt/rocm-5.2.3/include` instead of `/opt/rocm-5.2.3`.

The idea here is that we search the user-provided paths first, then the CMake defaults if nothing is found.


kennyweiss (approved):
Thanks @white238 !

rhornung67 (approved):
Thank you @white238

davidbeckingsale (approved):

cyrush:
thanks @white238 !

PR #652, opened 2023-08-07
Update SetupHIP.cmake

CRobeck:
The Rocm CMake paths are changing and it looks like RAJA, along with some other applications are going to break. In current releases FindHIP.cmake is in /opt/rocm/lib/cmake/hip. However, for backwards compatibility there has been a sym link in /opt/rocm/hip/cmake/FindHIP.cmake. That link is getting removed in up coming Rocm releases. Preemptively attempt to avoid breaking RAJA builds.

### DIFF
--- a/cmake/thirdparty/SetupHIP.cmake
+++ b/cmake/thirdparty/SetupHIP.cmake
@@ -33,7 +33,7 @@ endif()
 
 # Update CMAKE_PREFIX_PATH to make sure all the configs that hip depends on are
 # found.
-set(CMAKE_PREFIX_PATH "${CMAKE_PREFIX_PATH};${ROCM_ROOT_DIR}/lib/cmake;${ROCM_PATH}")
+set(CMAKE_PREFIX_PATH "${CMAKE_PREFIX_PATH};${ROCM_PATH};${ROCM_ROOT_DIR}/lib/cmake")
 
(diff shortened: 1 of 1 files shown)


cyrush:
@CRobeck is this rocm version dependent or cmake version dependent?

In any case, best would be to detect the new case and support it, while still supporting the older install layouts (since folks may still be using old rocm and cmake installs)

CRobeck:
It should only be Rocm version dependent. I'm just tacking the new path on to the CMAKE_PREFIX_PATH so if Cmake doesn't find FindHIP.cmake it just cycles on to the next path in that list. So, this should support all the existing Rocm versions (5.0-5.6) and then just kick when the sym link gets removed in future releases. 

@MrBurmark what are your thoughts here?

cyrush:
Got it, since these are all just additional places to look this looks good.

CRobeck:
I'm using ${ROCM_ROOT_DIR} here because RAJA has historically set that with their build scripts:

https://github.com/LLNL/RAJA/blob/develop/scripts/lc-builds/toss4_amdclang.sh#L63

Depending on the other blt users and how they do their CMake configures we may need to add ${ROCM_PATH}/lib/cmake. So just keep this PR in mind if/when that comes up.

white238 (approved):

cyrush (approved):

PR #659, opened 2023-10-05
feature: Add tests/projects and execute tests as part of CI

johnbowen42:
This PR adds integration tests to the CI of BLT.  These tests contain toy repositories that use BLT to configure CMake.   As of this commit, the directory tests/projects contain tests of the BLT macro `blt_install_tpl_setups`.  These tests assert that generator expressions for compilation and linkage are passed correctly between projects.


johnbowen42:
please see [https://github.com/LLNL/blt_test/pull/5](https://github.com/LLNL/blt_test/pull/5) for more tests that this PR enables

kennyweiss:
Do we want the `blt` repo to have a submodule into `blt_test` just for testing? or do we want the dependency to go the other way? Perhaps this has already been discussed and decided.

johnbowen42:
> Do we want the `blt` repo to have a submodule into `blt_test` just for testing? or do we want the dependency to go the other way? Perhaps this has already been discussed and decided.

We two options with @adrienbernede via email.  One would be to clone and check out a `blt_test` commit as part of CI for blt.  The second, which this PR is currently proceeding with, is to add `blt_test` as a submodule.  If folks have objections to this I'd be interested in hearing them, and we can still switch to cloning if necessary.

johnbowen42:
> Do we want the `blt` repo to have a submodule into `blt_test` just for testing? or do we want the dependency to go the other way? Perhaps this has already been discussed and decided.

In an offline conversation with @white238 we decided to just merge the `blt_test` submodule directly into `blt`, because the size of the `blt_test` repo is not large enough to justify the complexity of adding a submodule.

.gitlab/run_blt_test.sh
```diff
@@ -0,0 +1,13 @@
...
+RUNNER_SCRIPT=`pwd`/blt_test/run_tests.py
+git submodule update --init --recursive
+cd `pwd`/blt_test/
+
+if [[ -z $HOST_CONFIG  ]]; then
+    HOST_CONFIG=`pwd`/blt_test/$HOST_CONFIG
+    python3 $RUNNER_SCRIPT --run-test $ENABLED_BLT_TESTS --verbose --clean --blt-source-dir $BLT_DIR
+else
+    python3 $RUNNER_SCRIPT --run-test $ENABLED_BLT_TESTS --host-config $HOST_CONFIG --verbose --clean --blt-source-dir $BLT_DIR
+fi
```
white238, line 13:
```suggestion
fi

```

adrienbernede:
@johnbowen42 @kennyweiss @white238 note that my suggestion was to for BLT to trigger the BLT_Test CI as a *subpipeline* in GitLab CI. Not using BLT_Test as a submodule of BLT. The latter would not have been straightforward because of GitLab limitations.

If BLT_Test can be merged into BLT, then I guess there is no debate.

.gitlab-ci.yml
```diff
@@ -30,10 +30,27 @@ stages:
...
+  artifacts:
+    reports:
+      junit: build/junit.xml
+
 
 # This is where jobs are included
 include:
   - local: .gitlab/build_lassen.yml
   - local: .gitlab/build_quartz.yml
 # white238: Turned off while tioga is down for maintenance and is fickle
```
white238, line 54: We should be able to turn this back on now.. Tioga has been behaving for a while now.

.gitlab/build_tioga.yml
```diff
@@ -17,10 +17,26 @@
     ALLOC_COMMAND: "flux mini run -t10m -n1"
   extends: [.build_script, .on_tioga]
 
+####
+# Template
+.run_project_integration_test_on_tioga:
+  stage: test
+  variables:
+    ALLOC_COMMAND: "flux mini run -t10m -n1"
```
white238, line 25:
```suggestion
    ALLOC_COMMAND: "flux run -t10m -n1"
```

tests/projects/share/blt/logo/blt_logo.png
white238, line 1: This doesn't need to be here

tests/projects/test/cuda-arch/downstream/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+if (expected_cuda_toolkit_dir EQUAL -1)
+    message(FATAL_ERROR "Expected downstream project config flags to inherit upstream project's when alternative is not specified.")
+endif()
+
+# Compile a basic example to test correctness of link and compile flags.
+blt_add_executable(
+  NAME hello-cuda
+  SOURCES cuda-clang-cpp-user.cpp
+  DEPENDS_ON cuda-clang-cpp cuda
```
white238, line 50:
```suggestion
  DEPENDS_ON cuda-clang-cpp blt::cuda
```

I put in the alias for all the BLT targets and would like to start socializing using the namespaced ones. (Also make sure they work in all the cases they should).

tests/projects/test/cuda-mpi-cpp/base/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+# Add third-party library setup files to the project's installation
+# directory.
+blt_install_tpl_setups(DESTINATION lib/cmake/${PROJECT_NAME})
+
+# Add the base library with its single source file and header file.
+blt_add_library(
+  NAME cuda-mpi-cpp
+  SOURCES cuda-mpi-cpp.cu
+  HEADERS cuda-mpi-cpp.cuh
+  DEPENDS_ON cuda)
```
white238, line 20:
```suggestion
  DEPENDS_ON blt::cuda)
```

tests/projects/test/cuda-arch/README.md
kennyweiss, line 1: Suggestion: Can we move everything up a directory?
Instead of `<blt>/tests/projects/test/<project_name>`, could it be `<blt>/tests/projects/<project_name>`?

adrienbernede:
@johnbowen42 This PR is a good place to start using the ci reservation on both machines (instead of pdebug).

white238 (approved):
Thanks @johnbowen42 !

tests/projects/cuda-arch/base/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+cmake_minimum_required(VERSION 3.14)
```
kennyweiss, line 1: Is this the right minimum version of cmake for cuda-based tests in blt?

@white238 -- in a follow-up PR, we might want to ensure that all the declared minimum versions throughout blt are consistent/accurate.

tests/projects/hip-config-test/downstream/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+if (NOT TARGET blt_hip)
+  message(FATAL_ERROR "Target blt_hip was not configured successfully during"
+                      "downstream project configuration.")
+endif()
+
+if (NOT TARGET blt_hip_runtime)
+  message(FATAL_ERROR "Target blt_hip_runtime was not configured successfully during"
+                      "downstream project configuration.")
+endif()
```
kennyweiss, line 23: Should these (also) be checking for the existence of the alias targets -- e.g. `blt::hip`

tests/projects/cuda-arch/base/CMakeLists.txt
kennyweiss: Also, is this note at the beginning of the BLT docs still accurate?

> BLT officially supports CMake 3.8 and above. However we only print a warning if you are below this version. Some features in earlier versions may or may not work. Use at your own risk.

kennyweiss (approved):
Thanks @johnbowen42 -- this looks great!

* I appreciate the `README` files in each directory indicating why the tests are there and when they'll need to be updated
* I noticed that the new files are missing copyright blocks? 
    * This file from axom might be a helpful starting point: https://github.com/LLNL/axom/blob/develop/scripts/copyrightPrepender.py  -- although it looks like it's only set up for C/C++ header files

.gitlab/build_tioga.yml
white238: This is not resolved. `flux mini` is a deprecated command. All that needs to happen is to remove `mini`.

tests/projects/run_tests.py
```diff
@@ -0,0 +1,217 @@
...
+    if not args["blt-source-dir"]:
+        print("[ERROR: Required command line argument, 'blt-source-dir', was not provided.]")
+        return None
+
+    if args["host-config"] is not None and not os.path.exists(args["host-config"]):
+        print("ERROR: Host config file {0} specified, but file does not exist.".format(args["host-config"]))
+        return None
+
+    if args["run-test"] is not None:
+        all_tests = set(os.listdir(os.path.relpath("test")))
```
white238, line 145:
```suggestion
        all_tests = set(os.listdir(os.path.relpath(".")))

```

I think this is right..

tests/projects/hip-config-test/downstream/CMakeLists.txt
white238: <shrugs> those are the underlying targets.. so this should fail if things go wrong.

tests/projects/cuda-arch/base/CMakeLists.txt
white238: This is actually a very hard question to answer. It is nearly impossible to know what is the minimum and really depends on the project and what they are using.

Also it is not just the CMake version minimum. Depending on what you put there CMake activates different policies that can change how CMake behaves.

.gitlab/build_tioga.yml
```diff
@@ -14,13 +14,29 @@
...
+  extends: [.run_project_integration_tests, .on_tioga]
+
 ####
 # Build jobs
 tioga-clang_14_0_0_hip:
   variables:
     HOST_CONFIG: "clang@14.0.0_hip.cmake"
   extends: .src_build_on_tioga
   allow_failure: true
```
white238, line 34:
```suggestion
```

PR #663, opened 2023-11-01
allow blt::cuda as target name for cuda

trws:
What it says on the tin.  This is the minimum I could figure out to allow `blt::cuda` to mirror `blt::hip`.


trws:
closing to move this from my fork to a branch here to unstick gitlab

PR #664, opened 2023-11-01
allow blt::cuda as target name for cuda

trws:
What it says on the tin. This is the minimum I could figure out to allow blt::cuda to mirror blt::hip.


davidbeckingsale (approved):
Looks reasonable. Do you also want to make the change for cuda_runtime?

rhornung67:
This change gives the same CMake error in RAJA when moving to the latest desul. I think we should continue to live with the inconsistent namespacing between CUDA and HIP in BLT, until the `blt::*` namespacing is fully fleshed out in BLT and available in a BLT release.

trws:
I tested it against the latest desul, and got no errors.  How did you test it @rhornung67?  If that's still broken there's no point, but it shouldn't be.

Happy to add one for cuda_runtime.

trws:
Actually, the tests on desul/desul#117 include this update, and are currently passing.

rhornung67:
@trws just looking at your RAJA PR https://github.com/LLNL/RAJA/pull/1563

CMake error for desul atomics build on lassen.

trws:
I just tested it, it works when not run with uberenv.  My current guess is that something in that stack is overriding the BLT being used so we get an old one.  Will probably have to dig to find it.  Here are the outputs for the clang and gcc versions, the one on the left _should_ match what's in CI if it were using the same BLT:

image

cmake/thirdparty/BLTSetupCUDA.cmake
```diff
@@ -173,3 +175,4 @@ blt_import_library(NAME       cuda_runtime
                    INCLUDES   ${CUDA_INCLUDE_DIRS}
                    LIBRARIES  ${CUDA_LIBRARIES}
                    EXPORTABLE ${BLT_EXPORT_THIRDPARTY})
+add_library(blt::cuda_runtime ALIAS cuda)
```
rhornung67, line 178:
```suggestion
add_library(blt::cuda_runtime ALIAS cuda_runtime)
```

Should it be this instead?

trws:
It most definitely should, thanks @rhornung67.

rhornung67:
> It most definitely should, thanks @rhornung67.

I think I'm hearing the Benny Hill theme song in my head....

white238:
I am going to merge this then immediately do a PR with namespacing the rest of the targets along with updating all documentation. I am planning on doing a breaking release for BLT this year and will remove the generic names at that point.

PR #667, opened 2023-12-08
Fix infinite loop in `blt_find_target_dependencies`

white238:
Also simplifies the macro.

### DIFF
--- a/cmake/BLTPrivateMacros.cmake
+++ b/cmake/BLTPrivateMacros.cmake
@@ -777,46 +777,39 @@ macro(blt_find_target_dependencies)
         message(FATAL_ERROR "TLIST is a required parameter for the blt_find_target_dependencies macro")
     endif()
 
+    set(_depends_on "")
+
+    # Get dependencies if BLT registered library
     string(TOUPPER ${arg_TARGET} _target_upper)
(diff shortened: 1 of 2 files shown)


cmake/BLTPrivateMacros.cmake
```diff
@@ -777,46 +777,39 @@ macro(blt_find_target_dependencies)
...
-    # check if this is a valid cmake target
+    # Get dependencies if CMake target
     if(TARGET ${arg_TARGET})
-        # get link libaries if whitelisted
-        set(_property_list "")
         get_property(_target_type TARGET ${arg_TARGET} PROPERTY TYPE)
         if(NOT "${_target_type}" STREQUAL "INTERFACE_LIBRARY")
-            get_property(_propval TARGET ${arg_TARGET} PROPERTY LINK_LIBRARIES SET)
             get_target_property(_propval ${arg_TARGET} LINK_LIBRARIES)
-            if (_propval AND NOT "${_propval}" IN_LIST ${arg_TLIST})
```
white238, line 801: The bug was here and the previous line 809. `_propval` was a list not a single value so it would basically never not be in the list.

cyrush (approved):

PR #671, opened 2024-01-10
Revamp blt_convert_to_system_includes

adayton1:
* Allow passing multiple targets (TARGETS [<target>...])
* Follow transitive dependencies (unless CHILDREN is set to FALSE)
* Warn if target does not exist or if using the PGI compiler (unless QUIET is specified)
* Warn about deprecated TARGET option


adayton1:
We frequently have multiple targets to handle (e.g. all the axom targets), so it's a definite upgrade to handle them all in one invocation.

Perhaps I'm thinking about the transitive dependencies incorrectly, but it seems like if we change a library's interface include directories to system include directories, but don't do that for one of its dependencies, a warning from its dependency could propagate up. If I'm wrong about that, please let me know.

The other case I'm thinking about is where some targets may or may not exist depending on the build configuration (e.g. if some of axom's components are turned off). It would be easier just to specify them all in the same invocation and have the function ignore them. For safety, it would warn by default if the target does not exist (only for the targets specified by the user). To turn off the warnings, you could specify QUIET. Another way to handle this would be to have both TARGETS and OPTIONAL_TARGETS as options (or REQUIRED_TARGETS and OPTIONAL_TARGETS).

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,80 @@ endmacro(blt_add_target_compile_flags)
...
-        get_target_property(_include_dirs ${arg_TARGET} INTERFACE_INCLUDE_DIRECTORIES)
-        # Don't copy if the target had no include directories
-        if(_include_dirs)
-            # Clear previous value in INTERFACE_INCLUDE_DIRECTORIES so it is not doubled
-            # by target_include_directories
-            set_property(TARGET ${arg_TARGET} PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
-            target_include_directories(${arg_TARGET} SYSTEM INTERFACE ${_include_dirs})
-        endif()
-    endif()
+        foreach(_target ${arg_TARGETS})
```
white238, line 465: We have a macro to find all the dependencies of a given target:

```
## blt_find_target_dependencies(TARGET <target> TLIST <list name>)
##
## Store all target's dependencies (link libraries and interface link libraries)
## recursively in the variable name TLIST holds.

```

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,80 @@ endmacro(blt_add_target_compile_flags)
...
+## blt_convert_to_system_includes(TARGETS [<target>...]
+##                                [QUIET]
+##                                [INTRANSITIVE])
 ##
 ## Converts existing interface includes to system interface includes.
 ##------------------------------------------------------------------------------
-macro(blt_convert_to_system_includes)
-    set(options)
+function(blt_convert_to_system_includes)
+    set(options INTRANSITIVE QUIET)
```
white238, line 438: Instead of `INTRANSITIVE`, I would suggest using the already BLT convention used in `blt_print_target_properties` of `CHILDREN   <TRUE|FALSE>`

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,80 @@ endmacro(blt_add_target_compile_flags)
 
 
 ##------------------------------------------------------------------------------
-## blt_convert_to_system_includes(TARGET <target>)
+## blt_convert_to_system_includes(TARGETS [<target>...]
```
white238, line 431: I would prefer not to break this interface. If you want to add `TARGETS` I would be okay with that but not remove `TARGET`

cmake/BLTInstallableMacros.cmake
adayton1: Don't we only want interface link libraries? I thought link libraries were dependencies whose headers are not exposed in the headers of the library in question.

cmake/BLTInstallableMacros.cmake
adayton1: Thinking about it more, does this even need to be an option? Is there ever a case where you'd want a top level library's headers to be treated as system includes but not the headers of its dependencies?

adayton1: I'm now using CHILDREN, but the quest still remains as to whether it is even necessary to make an option.

cmake/BLTInstallableMacros.cmake
adayton1: I left `TARGET` as an option, but marked it as deprecated so that people have a chance to change before a breaking release.

cmake/BLTInstallableMacros.cmake
adayton1: Now using blt_find_target_dependencies

cmake/BLTInstallableMacros.cmake
adayton1: There are some use cases for it, so I'm leaving in CHILDREN as a parameter.

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,79 @@ endmacro(blt_add_target_compile_flags)
...
-            set_property(TARGET ${arg_TARGET} PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
-            target_include_directories(${arg_TARGET} SYSTEM INTERFACE ${_include_dirs})
-        endif()
-    endif()
+        set(_target_list)
+
+        foreach(_target ${arg_TARGETS})
+            if(TARGET ${_target})
+                if(${arg_CHILDREN})
+                    blt_find_target_dependencies(TARGET ${_target} TLIST _target_list)
```
adayton1, line 466: Is it safe to call macros from BLTPrivateMacros.cmake in BLTInstallableMacros.cmake?

white238: no.. but you can move that macro into the installable file. I could actually use it in Axom's installed config.

adayton1: I moved it there. Should I add sphinx documentation for that macro now, too? Or will we still consider it to be a private macro?

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,143 @@ endmacro(blt_add_target_compile_flags)
...
+    if(NOT DEFINED arg_CHILDREN)
+        set(arg_CHILDREN TRUE)
+    endif()
+
+    # PGI does not support -isystem
+    if(NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
+        set(_target_list "")
+
+        foreach(_target ${arg_TARGETS})
+            if(TARGET ${_target})
```
adayton1, line 525: Is there an easy way to handle BLT registered targets as well? e.g. will this work? `if(_BLT_${_target_upper}_IS_REGISTERED_LIBRARY)`

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,143 @@ endmacro(blt_add_target_compile_flags)
...
+            elseif(NOT ${arg_QUIET)
+                message(WARNING "${_target} does not exist!")
+            endif()
+        endforeach()
+
+        blt_list_remove_duplicates(TO _target_list)
+
+        foreach(_target ${_target_list})
+            # Ignore entries that are not actual CMake targets
+            if(TARGET ${_target})
```
adayton1, line 540: Do you know how I would update include directories on a BLT registered target?

adayton1: Would it be as simple as `set(_BLT_${uppercase_dependency}_TREAT_INCLUDES_AS_SYSTEM TRUE)`?

cmake/BLTInstallableMacros.cmake
white238: That is how you know if it is a registered library.

cmake/BLTInstallableMacros.cmake
white238: that is correct

cmake/BLTInstallableMacros.cmake
adayton1: Just documented it.

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,157 @@ endmacro(blt_add_target_compile_flags)
...
+    endif()
+
+    # PGI does not support -isystem
+    if(NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
+        set(_target_list "")
+
+        foreach(_target ${arg_TARGETS})
+            string(TOUPPER ${_target} _target_upper)
+
+            if(TARGET ${_target} OR ${_BLT_${_target_upper}_IS_REGISTERED_LIBRARY})
```
white238, line 528:
```suggestion
            if(TARGET ${_target} OR _BLT_${_target_upper}_IS_REGISTERED_LIBRARY)
```

This is going to evaluate to an empty string which causes a parsing error in CMake. Pretty sure this fix is right.

adayton1: I'll trust you on this.

adayton1:
I think the tests just timed out rather than failed.

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,157 @@ endmacro(blt_add_target_compile_flags)
...
+                                               SYSTEM
+                                               INTERFACE
+                                               ${_interface_include_dirs})
+                endif()
+
+                unset(_interface_include_dirs)
+            else()
+                string(TOUPPER ${_target} _target_upper)
+
+                if(${_BLT_${_target_upper}_IS_REGISTERED_LIBRARY})
```
white238, line 568:
```suggestion
                if(_BLT_${_target_upper}_IS_REGISTERED_LIBRARY)
```

white238:
This needs a line added to RELEASE_NOTES

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,157 @@ endmacro(blt_add_target_compile_flags)
...
+        message(WARNING "TARGET is a deprecated parameter for the blt_convert_to_system_includes macro.")
+        list(APPEND arg_TARGETS ${arg_TARGET})
+    endif()
+
+    if(NOT DEFINED arg_TARGETS)
+        message(FATAL_ERROR "TARGETS is a required parameter for the blt_convert_to_system_includes macro.")
+    endif()
+
+    if(NOT DEFINED arg_CHILDREN)
+        set(arg_CHILDREN TRUE)
```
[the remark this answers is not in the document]
adayton1: Would you like me to change this?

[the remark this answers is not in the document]
adayton1: The default is now FALSE.

white238 (approved):
Thanks @adayton1 ! Sorry this took so long....

white238:
I removed the deprecated warning due to the impact it would have on current users. I marked this in the list of breaking changes we are planning on doing soon. At that time we will have a release that has deprecation warnings.

adayton1:
Thanks for pushing this through, @white238 !

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,161 @@ endmacro(blt_add_target_compile_flags)
 
 
 ##------------------------------------------------------------------------------
-## blt_convert_to_system_includes(TARGET <target>)
+## blt_find_target_dependencies(TARGET <target> TLIST <list name>)
 ##
-## Converts existing interface includes to system interface includes.
+## Store all target's dependencies (link libraries and interface link libraries)
+## recursively in the variable name TLIST holds.
 ##------------------------------------------------------------------------------
```
kennyweiss, line 435: Did anything in `blt_find_target_dependencies` change? Or was it just moved to a new file?

cmake/BLTInstallableMacros.cmake
```diff
@@ -428,37 +428,161 @@ endmacro(blt_add_target_compile_flags)
...
+endmacro(blt_find_target_dependencies)
+
+
+##------------------------------------------------------------------------------
+## blt_convert_to_system_includes(TARGET   <target>
+##                                TARGETS  [<target>...]
+##                                CHILDREN [TRUE|FALSE]
+##                                [QUIET])
+##
+## Converts existing interface includes to system interface includes.
```
kennyweiss, line 498: I see that there was a discussion about not marking `TARGET` as deprecated for now. 

Can we still include a comment that `TARGET` will be deprecated in the future, so users should prefer the `TARGETS` keyword? As is (and without a comment), the duplication of `TARGET` and `TARGETS` is confusing.

docs/api/utility.rst
```diff
@@ -215,21 +215,40 @@ command but doesn't throw an error if the list is empty or not defined.
...
 blt_convert_to_system_includes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code-block:: cmake
 
-    blt_convert_to_system_includes(TARGET <target>)
+    blt_convert_to_system_includes(TARGET   <target>
+                                   TARGETS  [<target>...]
+                                   CHILDREN [TRUE | FALSE (default)]
+                                   [QUIET])
```
kennyweiss, line 242: The docs should also have a comment/note that `TARGET` will be deprecated in the near future, so it would be better to use `TARGETS`

kennyweiss (approved):
Thanks @adayton1 

I was confused about both `TARGET` and `TARGETS`, so it would probably be a good idea to add a comment about this to the docs.

PR #678, opened 2024-01-22
resolve: gtest vs cmake vs python

cyrush:


cyrush:
here is the fix for gtest - let me know how you want to tackle.

cyrush:
closed in favor of: #678

PR #682, opened 2024-02-23
OpenMP and policy warning clean up

### LINKED ISSUE #611: Do we still need CMP0076 set to old?
Using newer cmake, we always get depreciation warning related to BLT's choice of old for CMP0076

```
CMake Deprecation Warning at blt/SetupBLT.cmake:87 (cmake_policy):
  The OLD behavior for policy CMP0076 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
  cmake/SetupBLT.cmake:40 (include)
  CMakeLists.txt:113 (include)
```

https://github.com/LLNL/blt/blob/8c229991e65e7a9603c621b47cb3ba158bb7468c/SetupBLT.cmake#L87

Seems any header only issues should be resolved with build vs install interface target settings?

white238:
This PR does the following:

* Utilizes the previous OpenMP compile flags with generator expressions in the link flags. This used to not be allowed but the target property `LINK_OPTIONS` has supported generator expressions since CMake 3.13. I ran into this on the clang+gfortran combination after I removed a hack inside of Axom. I was able to also remove a manual fortran flag replacement.
* Adds searching extra paths in various "find" calls in CMake. This came up with Umpire's honoring posix directories.
* Print out the implicit directories/libs that CMake keeps finding and plaguing us with.
* Update the CMake version that BLT prints a warning about from 3.8 to 3.15. Alias targets should be safe after this point.
* Fixes #611


adayton1 (approved):
LGTM

cmake/SetupCompilerOptions.cmake
```diff
@@ -358,12 +358,22 @@ endforeach()
...
-    list(REMOVE_ITEM CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES ${BLT_CMAKE_IMPLICIT_LINK_DIRECTORIES_EXCLUDE})
-    message(STATUS "Updated CXX implicit Link Directories: ${CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES}")
+
     list(REMOVE_ITEM CMAKE_C_IMPLICIT_LINK_DIRECTORIES ${BLT_CMAKE_IMPLICIT_LINK_DIRECTORIES_EXCLUDE})
     message(STATUS "Updated C implicit Link Directories: ${CMAKE_C_IMPLICIT_LINK_DIRECTORIES}")
+    list(REMOVE_ITEM CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES ${BLT_CMAKE_IMPLICIT_LINK_DIRECTORIES_EXCLUDE})
+    message(STATUS "Updated CXX implicit Link Directories: ${CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES}")
     if (ENABLE_FORTRAN)
         list(REMOVE_ITEM CMAKE_Fortran_IMPLICIT_LINK_DIRECTORIES ${BLT_CMAKE_IMPLICIT_LINK_DIRECTORIES_EXCLUDE})
         message(STATUS "Updated Fortran implicit Link Directories: ${CMAKE_Fortran_IMPLICIT_LINK_DIRECTORIES}")
```
kennyweiss, line 379: Suggestion (capitalization): 
`implicit Link Directories` -> `implicit link directories`  in our status messages (?)
(and similarly for libraries a few lines down)

kennyweiss (approved):
Thanks @white238 

Does this warrant a RELEASE-NOTES update?

cmake/SetupCompilerOptions.cmake
white238: fixed.

RELEASE-NOTES.md
```diff
@@ -9,6 +9,15 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
...
 ## [Unreleased] - Release date yyyy-mm-dd
 
+### Added
+- Added output for CMake's implicitly added link libraries and directories.
+
+### Changed
+- OpenMP target now uses a generator expression for Fortran flags instead of replacing flags in
+  Fortran targets created with BLT macros.
+- Remove setting CMP0076 to OLD which left relative paths in `target_sources()` instead of altering
+  them to absolute paths.
```
kennyweiss, line 19: 👍

cmake/BLTSetupTargets.cmake
```diff
@@ -16,30 +16,27 @@ if (NOT BLT_LOADED)
...
-  if (NOT DEFINED BLT_ENABLE_MPI OR NOT ${BLT_ENABLE_MPI})
-    set(BLT_ENABLE_MPI              ${ENABLE_MPI})
-  endif()
-  if (NOT DEFINED BLT_ENABLE_OPENMP OR NOT ${BLT_ENABLE_OPENMP})
-    set(BLT_ENABLE_OPENMP           ${ENABLE_OPENMP})
-  endif()
-  if (NOT DEFINED BLT_ENABLE_FIND_MPI OR NOT ${BLT_ENABLE_FIND_MPI})
-    set(BLT_ENABLE_FIND_MPI         ${ENABLE_FIND_MPI})
+  if (DEFINED BLT_ENABLE_HIP)
+    message(FATAL_ERROR "This variable should never exist at this point. Something has gone horribly wrong.")
```
adayton1, line 30: Why only a warning for this variable? I'm guessing this is for debugging?

cmake/BLTMacros.cmake
```diff
@@ -209,19 +209,14 @@ macro(blt_add_library)
...
-        add_library( ${arg_NAME} INTERFACE )
-        target_sources( ${arg_NAME} INTERFACE
-                        $<BUILD_INTERFACE:${_build_headers}>)
+        #  Note: Will not add header files to header-only interface libraries in older versions
+        #  of CMake when CMP0076 is set to NEW
+        #
+        if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.19.0" )
+            add_library( ${arg_NAME} INTERFACE ${arg_HEADERS} )
+        else()
+            add_library( ${arg_NAME} INTERFACE )
```
adayton1, line 218: Do we want a warning for this case?

cyrush (approved):

cmake/BLTSetupTargets.cmake
white238: This was to guard against a weird state of the TPL variables existing when they shouldn't. Basically when two BLT's are loaded in the same project. But we have BLT_LOADED for this reason. It should be safe to delete but I could also just put all the variables there.

cmake/BLTMacros.cmake
white238: I clarified the comment. Basically header only libraries don't need to add their sources but if you want them to show up in your IDE project then you have to. CMake is very fickle about this and depending on the version you have to jump through a lot of hoops. I went with the semi-sane answer to this in "newer" versions.

SetupBLT.cmake
```diff
@@ -41,10 +41,10 @@ if (NOT BLT_LOADED)
     endif()
     unset(_is_multi_config)
 
-    if(${CMAKE_VERSION} VERSION_LESS 3.8.0)
+    if(${CMAKE_VERSION} VERSION_LESS 3.15.0)
```
kennyweiss, line 44: The update of the min version should probably also be noted in the Release-notes due to potential changes to cmake policies.  (I get that this is only a warning from our side, but it's stating blt's expectations).

cmake/BLTMacros.cmake
kennyweiss: We should note this in the RELEASE-NOTES as well, to give upgraders a heads-up

cmake/BLTSetupTargets.cmake
kennyweiss: I agree w/ @adayton1 -- it should be ~all~ [a representative sample] or nothing

cmake/SetupCompilerOptions.cmake
```diff
@@ -358,31 +358,51 @@ endforeach()
...
-    list(REMOVE_ITEM CMAKE_CXX_IMPLICIT_LINK_LIBRARIES ${BLT_CMAKE_IMPLICIT_LINK_LIBRARIES_EXCLUDE})
-    message(STATUS "Updated CXX implicit Link Libraries: ${CMAKE_CXX_IMPLICIT_LINK_LIBRARIES}")
+
     list(REMOVE_ITEM CMAKE_C_IMPLICIT_LINK_LIBRARIES ${BLT_CMAKE_IMPLICIT_LINK_LIBRARIES_EXCLUDE})
     message(STATUS "Updated C implicit Link Libraries: ${CMAKE_C_IMPLICIT_LINK_LIBRARIES}")
+    list(REMOVE_ITEM CMAKE_CXX_IMPLICIT_LINK_LIBRARIES ${BLT_CMAKE_IMPLICIT_LINK_LIBRARIES_EXCLUDE})
+    message(STATUS "Updated CXX implicit Link Libraries: ${CMAKE_CXX_IMPLICIT_LINK_LIBRARIES}")
     if (ENABLE_FORTRAN)
         list(REMOVE_ITEM CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES ${BLT_CMAKE_IMPLICIT_LINK_LIBRARIES_EXCLUDE})
         message(STATUS "Updated Fortran implicit Link Libraries: ${CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES}")
```
kennyweiss, line 408: Nit: please make the case in the status messages to "implicit link libraries" to be consistent w/ "implicit link directories" above

kennyweiss (approved):
Thanks @white238 !

cmake/SetupCompilerOptions.cmake
white238: there is a chance im losing my mind at this point. I thought i fixed this.

RELEASE-NOTES.md
```diff
@@ -17,6 +17,9 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
   Fortran targets created with BLT macros.
 - Remove setting CMP0076 to OLD which left relative paths in `target_sources()` instead of altering
   them to absolute paths.
+- Header-only libraries headers now show up under their own target in IDE project views instead of
+  under downstream targes. This only works in CMake >= 3.19, otherwise they will not show up at all.
```
cyrush, line 21:
```suggestion
  under downstream targets. This only works in CMake >= 3.19, otherwise they will not show up at all.
```

cyrush (approved):

Issue #686, opened 2024-05-01
blt cuda setup issue with newer cmake?


cyrush:
```
cmake version 3.26.3
```
```
-- Creating BLT CUDA targets...
CMake Warning (dev) at /__w/1/s/install/lib/cmake/ascent/thirdparty/BLTSetupCUDA.cmake:25 (if):
  Policy CMP0057 is not set: Support new IN_LIST if() operator.  Run "cmake
  --help-policy CMP0057" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  IN_LIST will be interpreted as an operator when the policy is set to NEW.
  Since the policy is not set the OLD behavior will be used.
Call Stack (most recent call first):
  /__w/1/s/install/lib/cmake/ascent/BLTSetupTargets.cmake:97 (include)
  /__w/1/s/install/lib/cmake/ascent/AscentConfig.cmake:150 (include)
  CMakeLists.txt:38 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at /__w/1/s/install/lib/cmake/ascent/thirdparty/BLTSetupCUDA.cmake:25 (if):
  if given arguments:

    "CUDA" "IN_LIST" "_languages"

  Unknown arguments specified
Call Stack (most recent call first):
  /__w/1/s/install/lib/cmake/ascent/BLTSetupTargets.cmake:97 (include)
  /__w/1/s/install/lib/cmake/ascent/AscentConfig.cmake:150 (include)
  CMakeLists.txt:38 (find_package)
```

cyrush:
CMake version was old in my downstream example.

cyrush:
```
cmake_minimum_required(VERSION 3.14)
project(using_with_cmake)
cmake_policy(SET CMP0057 NEW)
```

cyrush:
I still can't get this to work, is there an example of a using-with-cmake that is using new BLT?

white238:
Axom's is tested in every PR:

https://github.com/LLNL/axom/blob/develop/src/examples/using-with-cmake/CMakeLists.txt

cyrush:
Thanks, I have to add the policy command in the ascent config - or else it doesn't take.

Axom is doing this as well:

https://github.com/LLNL/axom/blob/f495289b65f3c9b4ca74627460ed7e6bd5077638/src/cmake/axom-config.cmake.in#L216

white238:
[...]

PR #687, opened 2024-05-02
Support if(... IN_LIST ...) in downstream non-blt projects

white238:
Error case:

* BLT project, "Foo", installs BLT targets via `blt_install_tpl_setups()` macro
* Downstream Non-BLT project loads Foo's installed config which includes' BLT's targets
* Downstream project possibly fails when BLT uses an `if` with an `IN_LIST` operator


white238:
Fixes #686

rhornung67 (approved):
Thank you @white238

chapman39 (approved):

cyrush:
thanks @white238 !

PR #691, opened 2024-06-19
Remove GNUInstallDirs from Builtin TPLs

white238:
This was bleeding into user installs. I am not sure how this wasn't causing problems before.


davidbeckingsale (approved):
I think it has been for a while - must be where we started seeing lib64 come from (I assumed that was a Cmake version thing)

rhornung67 (approved):
Thank you @white238

kennyweiss (approved):

PR #696, opened 2024-08-26
Harden two cases where projects do not enable C as a language

### LINKED ISSUE #660: C test enabled without checking if C is an enabled language
https://github.com/LLNL/blt/blob/a7f0a6ecc4fdfa1724399b1454c3909b9ee02e81/tests/smoke/CMakeLists.txt#L211-L218

There should be a guard around this so that it is not turned on if `C` is not an enabled language.

white238:
Fixes #660

This PR does two things:

* Guards a smoke test against projects who don't enable C as a language
* Enables the C language, and emits a warning, when projects use GTest or GMock and don't have the required C language enabled.

The latter is required due to GTest|GMock turning on C, which in CMake enables the language globally but several needed CMake variables are directory scoped. This causes a CMake error like this:

```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_C_COMPILE_OBJECT
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_C_LINK_EXECUTABLE
-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

The omitted warning looks like this:

```
CMake Warning at blt/repo/thirdparty_builtin/CMakeLists.txt:95 (message):
  Enabling the C language due to Google Mock|Test requiring it.  To quiet
  this warning, add C to your project(...) call.
```

@rcarson3

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -13,6 +13,9 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 - Removed GoogleTest, GoogleMock, and GoogleBenchmarks calling CMake's `GNUInstallDirs`
   which was causing non-deterministic install variables depending on your combination of
   static/shared libraries and if any of the previously mentioned TPLs were enabled.
+- Enable C language support and emit a warning when GoogleTest or GoogleMock are enabled but
+  not the C language.
+- Guard HIP C smoketest against projects that don't enable C as a language in their projects.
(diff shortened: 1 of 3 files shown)


kennyweiss (approved):
Thanks @white238!

cyrush (approved):

PR #698, opened 2024-09-05
Prevent general tests from running while running benchmarks custom target

chapman39:
### Issue

It is expected only benchmarks should run when executing `make run_benchmarks`.

The configuration `-C` option in `add_custom_target` prevents benchmarks from running while executing `make test`. However, all tests and benchmarks run while executing `make run_benchmarks`.

### This PR

Adds a "Benchmark" label `-L`. This means now only benchmarks will run during `make run_benchmarks`. And `make test` remains to only run general tests, no benchmarks.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -19,6 +19,7 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 - Enable C language support and emit a warning when GoogleTest or GoogleMock are enabled but
   not the C language.
 - Guard HIP C smoketest against projects that don't enable C as a language in their projects.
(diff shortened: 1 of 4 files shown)


chapman39:
Also, users can still run general tests and exclude benchmarks with `ctest -LE Benchmarks`.

kennyweiss (approved):
Thanks @chapman39 !

Are there any docs to update? Or does this fall under general CMake usage?

kennyweiss (approved):
Thanks!

rhornung67 (approved):

white238 (approved):

white238:
Thanks @chapman39 !

PR #724, opened 2025-08-14
Process compile flag lists like link_flags

### LINKED ISSUE #723: `blt::openmp` INTERFACE_COMPILE_OPTIONS may not delimit lists correctly.
`blt_add_target_compile_flags()` does not prefix `SHELL:` the way `blt_add_target_link_flags()` does. This means that BLTSetupOpenMP.cmake gets multiple arguments in the `${OpenMP_CXX_FLAGS}` string, we get a single list element containing a space instead of a semicolon delimited list of arguments.

The OpenMP::OpenMP_CXX target deals with this by prefixing `SHELL:`, so that seems like a reasonable approach.

For a failure case, consider that the Apple Clang compilers with `libomp` from Homebrew require '-Xclang -fopenmp'. Configure the following CMake project with `-DOpenMP_ROOT=/opt/homebrew/opt/libomp`.

```
cmake_minimum_required(VERSION 3.15)
project(testbltopenmp)

set(ENABLE_OPENMP TRUE)

include(blt/SetupBLT.cmake)

blt_print_target_properties(TARGET blt::openmp)
blt_print_target_properties(TARGET OpenMP::OpenMP_CXX)
```

Result:
```
...

Problems were encountered while collecting compiler information:
	clang++: error: unknown argument: '-Xclang -fopenmp'

```

eirrgang:
Add logic to `blt_add_target_compile_flags()` to prepend `SHELL:` to possible argument lists in the way that `blt_add_target_link_flags()`.

Fixes #723


eirrgang:
Alternatively, this could be done in BLTSetupOpenMP.cmake by changing to
```
blt_import_library(NAME          openmp
                   COMPILE_FLAGS SHELL:${_compile_flags}
                   LINK_FLAGS    ${_link_flags}
                   EXPORTABLE    ${BLT_EXPORT_THIRDPARTY})
```

eirrgang:
> Alternatively, this could be done in BLTSetupOpenMP.cmake by changing to
> 
> ```
> blt_import_library(NAME          openmp
>                    COMPILE_FLAGS SHELL:${_compile_flags}
>                    LINK_FLAGS    ${_link_flags}
>                    EXPORTABLE    ${BLT_EXPORT_THIRDPARTY})
> ```

I moved the patch to `BLTSetupOpenMP.cmake` to be less invasive, since there don't seem to be widespread complaints about this in `BLTInstallableMacros.cmake`.

PR #725, opened 2025-08-22
CUDA clean up

adayton1:
- Since BLT now requires a minimum of CMake 3.15, logic for earlier versions can be removed
- Improve error message for CUDA versions that don't support the requested language standard
- Print out CMAKE_CUDA_STANDARD
- Print out CMAKE_CUDA_ARCHITECTURES if CMake version is greater than or equal to 3.18

### DIFF
--- a/cmake/thirdparty/BLTSetupCUDA.cmake
+++ b/cmake/thirdparty/BLTSetupCUDA.cmake
@@ -110,6 +110,7 @@ message(STATUS "CUDA Implicit Link Directories: ${CMAKE_CUDA_IMPLICIT_LINK_DIREC
 ############################################################
 # Check CUDA language standard is supported
 ############################################################
+# Note: change to use CUDAToolkit_VERSION after moving away from find_package(CUDA)
 if(CMAKE_CUDA_STANDARD STREQUAL "17")
     if(NOT DEFINED CMAKE_CUDA_COMPILE_FEATURES OR (NOT "cuda_std_17" IN_LIST CMAKE_CUDA_COMPILE_FEATURES))
         message(FATAL_ERROR "CUDA ${CUDA_VERSION_STRING} does not support C++17.")


cmake/thirdparty/BLTSetupCUDA.cmake
```diff
@@ -67,18 +50,6 @@ if (NOT BLT_ENABLE_CLANG_CUDA)
     enable_language(CUDA)
 endif()
 
-if(CMAKE_CUDA_STANDARD STREQUAL "17")
```
adayton1, line 70: I just moved these checks to after the call to find_package(CUDA) so that we can use CUDA_VERSION_STRING.

white238: We really need to move away from `find_package(CUDA)` because it is deprecated since 3.10 and also in 4.0 if i remember correctly throws a error or a loud warning.

https://cmake.org/cmake/help/latest/module/FindCUDA.html

cmake/thirdparty/BLTSetupCUDA.cmake
```diff
@@ -11,44 +11,27 @@
...
 endif()
 
 # CUDA_HOST_COMPILER was changed in 3.9.0 to CMAKE_CUDA_HOST_COMPILER and
 # needs to be set prior to enabling the CUDA language
 get_property(_languages GLOBAL PROPERTY ENABLED_LANGUAGES)
-if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.9.0" )
-    if ( NOT CMAKE_CUDA_HOST_COMPILER )
-        if("CUDA" IN_LIST _languages )
-            message( FATAL_ERROR 
+if ( NOT CMAKE_CUDA_HOST_COMPILER )
```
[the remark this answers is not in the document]
adayton1: I modified the comment to what I think it should be now, but I'm really not sure if CMAKE_CUDA_HOST_COMPILER still needs to be set before enabling the CUDA language.

cmake/thirdparty/BLTSetupCUDA.cmake
```diff
@@ -128,6 +107,21 @@ message(STATUS "CUDA Link with NVCC:            ${CUDA_LINK_WITH_NVCC}")
 message(STATUS "CUDA Implicit Link Libraries:   ${CMAKE_CUDA_IMPLICIT_LINK_LIBRARIES}")
 message(STATUS "CUDA Implicit Link Directories: ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}")
 
+############################################################
+# Check CUDA language standard is supported
+############################################################
+if(CMAKE_CUDA_STANDARD STREQUAL "17")
```
white238, line 113:
```suggestion
# Note: change to use CUDAToolkit_VERSION after moving away from find_package(CUDA)
if(CMAKE_CUDA_STANDARD STREQUAL "17")
```

white238 (approved):
Thanks @adayton1

adayton1:
I don't have permission to merge, and I will be unavailable Monday and Tuesday. So feel free to merge whenever you are ready, @white238.

PR #739, opened 2025-10-22
Expand detection of MSVC

eirrgang:
COMPILER_FAMILY_IS_MSVC was set according to detected build tools, but may have missed some (such as Ninja, which is available for Windows). CMake sets MSVC=1 when an MSVC compatible compiler is detected. It is possible that some versions of CMake might not set MSVC=1 in all cases when it should, but it is hard to imagine a false positive, so using two checks for COMPILER_FAMILY_IS_MSVC seems fair.


white238 (approved):
Thanks @eirrgang ! Waiting on @kennyweiss to weigh in because he does the Axom Windows builds.

cmake/SetupCompilerOptions.cmake
```diff
@@ -16,7 +16,7 @@
 # use CMAKE_BUILD_TOOL to identify visual studio
 # and CMAKE_CXX_COMPILER_ID for all other cases
 
-if("${CMAKE_BUILD_TOOL}" MATCHES "(msdev|devenv|nmake|MSBuild)")
+if(MSVC OR ("${CMAKE_BUILD_TOOL}" MATCHES "(msdev|devenv|nmake|MSBuild)"))
```
kennyweiss, line 22: Please add a comment here about why this is necessary. (e.g. copied/adapted from your PR comment)

kennyweiss (approved):
Thanks @eirrgang !

cmake/SetupCompilerOptions.cmake
```diff
@@ -16,7 +16,7 @@
 # use CMAKE_BUILD_TOOL to identify visual studio
 # and CMAKE_CXX_COMPILER_ID for all other cases
```
eirrgang:
```suggestion
# Use CMake variable MSVC (or CMAKE_BUILD_TOOL) to identify msvc style compilers
# and CMAKE_CXX_COMPILER_ID for all other cases.
# Some build tools (such as Ninja) are available for both Windows and non-Windows platforms,
# but we can also use native Windows build tools to detect a Visual Studio environment
# in case CMake fails to detect and set MSVC=1 for some reason.
```

cmake/SetupCompilerOptions.cmake
eirrgang: Let me know if you are unable to commit the above suggestion (if you approve) or please suggest (or directly commit) edits.

cmake/SetupCompilerOptions.cmake
kennyweiss: Thanks @eirrgang !

PR #744, opened 2025-12-03
Robustify blt_add_test in non mpi configs

kennyweiss:
When the user provides `blt_add_test` with `NUM_MPI_RANKS`, we prepend the test command with the `MPIEXEC` and the number of ranks (e.g. `srun -n4`).

In non-mpi configurations, we should not add `MPIEXEC`, which is likely not a defined variable in CMake.

This PR fixes this, but also allows the `NUM_MPI_RANKS` to be 1 in non- mpi configurations.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -12,6 +12,11 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 ### Changed
 - Removed default use of `/bigobj` flag for Visual Studio builds. Projects should add this flag explicitly if needed for windows builds.
 
+### Fixed
+- In non-mpi configurations, `blt_add_test` will now throw a `FATAL_ERROR` if the user provides `NUM_MPI_RANKS`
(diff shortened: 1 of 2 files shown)


white238 (approved):
Thanks @kennyweiss

RELEASE-NOTES.md
```diff
@@ -12,6 +12,11 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 ### Changed
 - Removed default use of `/bigobj` flag for Visual Studio builds. Projects should add this flag explicitly if needed for windows builds.
 
+### Fixed
+- In non-mpi configurations, `blt_add_test` will now throw a `FATAL_ERROR` if the user provides `NUM_MPI_RANKS`
+  and the requested number of ranks is greater than 1. Previously, if `MPI_EXECUTABLE` was not defined
```
cyrush, line 17: Good fix!

PR #745, opened 2025-12-08
Oneapi 2025 introduces a flag name change impacting gtest

adrienbernede:
Report of the issue on the intel forum:

https://community.intel.com/t5/Intel-oneAPI-DPC-C-Compiler/oneAPI-2025-2-warning-10430-Wno-implicit-float-size-conversion/m-p/1699832/highlight/true#M4492

Pull request to fix the issue in google test:

https://github.com/google/googletest/pull/4798

Example of failures do to the issue:

https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/3168882

WARNING:
This PR is not necessarily meant to be merged... instead, we could wait for the fix to be merged in googletest and then update our copy.
In the meantime, I intend to use this PR as reference for a patch in the local copy of BLT in radiuss-spack-configs.


white238 (approved):
I think this should be merged.. Thanks @adrienbernede!

thirdparty_builtin/googletest-1.16.0/googletest/cmake/internal_utils.cmake
```diff
@@ -100,7 +100,12 @@ macro(config_compiler_and_linker)
       set(cxx_strict_flags "${cxx_strict_flags} -Wchar-subscripts")
     endif()
     if (CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
-      set(cxx_base_flags "${cxx_base_flags} -Wno-implicit-float-size-conversion -ffp-model=precise")
+      string(APPEND cxx_base_flags " -ffp-model=precise")
+      if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 2025.2.0)
+        string(APPEND cxx_base_flags " -Wno-implicit-float-size-conversion")
+      else()
+        string(APPEND cxx_base_flags " -Wsycl-implicit-float-size-conversion")
```
rhornung67: Does this require building with SYCL enabled?

adrienbernede:
I think there's a typo. I'll check and address Rich's comment.

davidbeckingsale:
Based on the Intel discussion forum the naming is correct, but I think it needs to be `-Wno-sycl-implicit-float-size-conversion` (the "no" is missing)

adrienbernede:
@davidbeckingsale Indeed! that's the typo I was thinking of...

kennyweiss (approved):
Thanks @adrienbernede -- can/should we add a host-config for this compiler on an LC system so we can test this change in our CI?

rhornung67:
> Thanks @adrienbernede -- can/should we add a host-config for this compiler on an LC system so we can test this change in our CI?

That's probably wise since code teams are using intel2025 for testing on LC. We're testing this in RAJA so Axom should add it.

thirdparty_builtin/googletest-1.16.0/googletest/cmake/internal_utils.cmake
adrienbernede: This works with Umpire that has SYCL support off by default.
Per the comment on the Intel forum, it's only to respect a naming convention.

adrienbernede:
> Thanks @adrienbernede -- can/should we add a host-config for this compiler on an LC system so we can test this change in our CI?

Oneapi 2025 is now in radiuss-spack-configs, so we test it in umpire for example. I can see to add it to BLT host-configs as well.

white238:
I added a host-config and test for the new compiler. I got the following warning and fixed it:

```
[  4%] Building CXX object blt/thirdparty_builtin/googletest-1.16.0/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o
icpx: command line warning #10430: Unsupported command line options encountered
These options as listed are not supported.
For more information, use '-qnextgen-diag'.
option list: 
	-Winline
```

according to this page (https://clang.llvm.org/docs/DiagnosticsReference.html#winline) `-Winline` is a noop anyways.

adrienbernede:
@white238 I can't hit the merge button. This looks good to me.

PR #749, opened 2025-12-22
Adds PROCESSORS property to tests based on number of ranks and threads

kennyweiss:
This PR add the `PROCESSORS` property to tests in the `blt_add_test` macro to give CTest more context in scheduling its tests.

The number of processors is based on the number of mpi ranks -- provided through `NUM_MPI_TASKS` -- and the number of OpenMP threads -- provided through `NUM_OMP_THREADS`.

We also throw a FATAL_ERROR if the number of threads is not an integer or is less than 1.

**Update:** 
* I tested this out in Axom (https://github.com/llnl/axom/pull/1749) and the results are not the slam dunk I thought they were. Adding the `PROCESSORS` property to tests appears to slow the testing down somewhat (~3% for the Axom tests).
* So, we should probably hold off on merging this, or, at the very least, making it opt-in. I'm going to mark this PR as DRAFT.

### DIFF
--- a/cmake/BLTMacros.cmake
+++ b/cmake/BLTMacros.cmake
@@ -490,9 +490,36 @@ macro(blt_add_test)
 
     # Handle OpenMP
     if( arg_NUM_OMP_THREADS )
+        # Ensure the number of OpenMP threads is a positive integer
+        string(REGEX MATCH "^-?[0-9]+$" _is_integer "${arg_NUM_OMP_THREADS}")
+        if( NOT _is_integer OR arg_NUM_OMP_THREADS LESS 1)
+            message(FATAL_ERROR "In blt_add_test(), attempted to run a test with invalid NUM_OMP_THREADS: ${arg_NUM_OMP_THREADS}")
+        endif()
+
         set_property(TEST ${arg_NAME}
                      APPEND PROPERTY ENVIRONMENT OMP_NUM_THREADS=${arg_NUM_OMP_THREADS})
     endif()
+
+    # Inform CTest how many processor slots this test needs for scheduling.
+    # When tests are parallelized with `ctest --parallel/-j`, CTest uses this
+    # property to avoid oversubscribing the available resources.
+    set(_blt_num_mpi_tasks 1)
+    if( arg_NUM_MPI_TASKS )
+        set(_blt_num_mpi_tasks ${arg_NUM_MPI_TASKS})
+    endif()
+
+    set(_blt_num_omp_threads 1)
+    if( arg_NUM_OMP_THREADS )
+        set(_blt_num_omp_threads ${arg_NUM_OMP_THREADS})
+    endif()
+
+    math(EXPR _blt_test_processors "${_blt_num_mpi_tasks} * ${_blt_num_omp_threads}")
+    if(_blt_test_processors LESS 1)
+        set(_blt_test_processors 1)
+    endif()
+
+    set_tests_properties(${arg_NAME}
+                         PROPERTIES PROCESSORS ${_blt_test_processors})
 
 endmacro(blt_add_test)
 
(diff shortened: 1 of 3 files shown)


cmake/BLTMacros.cmake
```diff
@@ -490,10 +490,37 @@ macro(blt_add_test)
...
+
         set_property(TEST ${arg_NAME}
                      APPEND PROPERTY ENVIRONMENT OMP_NUM_THREADS=${arg_NUM_OMP_THREADS})
     endif()
 
+    # Inform CTest how many processor slots this test needs for scheduling.
+    # When tests are parallelized with `ctest --parallel/-j`, CTest uses this
+    # property to avoid oversubscribing the available resources.
+    set(_blt_num_mpi_tasks 1)
+    if( arg_NUM_MPI_TASKS )
```
adayton1, line 511: Would it make sense to also check that if this is set, it should be 1 or greater?

kennyweiss: Thanks. We check that about 40 lines up -- https://github.com/llnl/blt/pull/749/changes#diff-feccb804465e3037f5f4ef6779f5f994221b63f8f8b6a034e77c016fd227f633R462-R467

cmake/BLTMacros.cmake
```diff
@@ -490,10 +490,37 @@ macro(blt_add_test)
 
     # Handle OpenMP
     if( arg_NUM_OMP_THREADS )
+        # Ensure the number of OpenMP threads is a positive integer
+        string(REGEX MATCH "^-?[0-9]+$" _is_integer "${arg_NUM_OMP_THREADS}")
```
adayton1: It might be good to make sure leading and following whitespace has been stripped from arg_NUM_OMP_THREADS. Otherwise, this check could fail and the user would have no idea what is wrong if they put " 1" instead of "1", for example.

cmake/BLTMacros.cmake
adayton1: Perfect! Then the check at line 517 shouldn't be necessary since you have verified both numbers being multiplied are 1 or greater: 

`if(_blt_test_processors LESS 1)`

kennyweiss: That's fair. I figured it helps readability, but I can remove it.

kennyweiss: Done.

cmake/BLTMacros.cmake
kennyweiss: Done.

adayton1 (approved):
Thanks, @kennyweiss !

white238 (approved):
Thanks @kennyweiss ! Did you notice a speed up after this?

bartlettroscoe:
FYI: We have been using the `PROCESSORS` property for 15+ years with TriBITS/Trilinos and it works okay.  Without this, CTest can't effectively schedule the running of tests that fill the cores but not overload them.  You can sese how effective this is by going to:

* https://github.com/orgs/trilinos/discussions/15303#discussioncomment-17140029

and downloading and viewing the plots under the section "System RAM and Load Fracs Test Plot".

aaroncblack:
@kennyweiss @white238 I tried adding PROCESSORS property on our tests in radar ( about 45 tests, mostly small, some multirank ), and it sped things up quite a bit for me.  I tried running with 'ctest --parallel 112' on one of our nodes with 112 cores.

I'm adding the property myself after blt creates the test but maybe we can look at the behavior later.  These tests aren't multi-threaded however.

One initial hiccup I had was I didn't make sure the srun jobs were launched with '--exclusive' so they were oversubscribed CPUs.  (I was running ctest in an salloc).

After I added:
```
set(BLT_MPI_COMMAND_APPEND "--exclusive" CACHE STRING "")                                                                                                                                                                                        
```
to my cmake host config everything worked well.

I'll see if I can test your branch here.

PR #754, opened 2026-02-04
add comm size check to blt mpi smoke

### LINKED ISSUE #753: blt mpi smoke test may need more smoke
I am hitting a case where the blt smoke test passes, but I believe the MPI environment is actually bad.

I suggest we enhance the blt smoke test to check the size of the communicator to make sure it matches the expected number of MPI ranks (which I believe is 4)

The MPI reduce in blt's mpi smoke does not fail, b/c I expect it is self consistent even if there is only one MPI rank.

My more complex (non blt) tests that use mpi send / recv fail with a message like:

```
Invalid rank has value 1 but must be nonnegative and less than 1
```

(this means the only valid rank id is `0`)

These tests should be definitely be launched with more than 1 mpi task :-(.

It is also possible that my MPI tests are launched differently than the BLT MPI test and that is where the surprise is.

Either way, having the comm size check in blt mpi smoke would provide more info and confidence about the underlying issue.

cyrush:
resolves #753, helps detect mpi env puzzles earlier

### DIFF
--- a/tests/smoke/blt_mpi_smoke.cpp
+++ b/tests/smoke/blt_mpi_smoke.cpp
@@ -25,7 +25,12 @@ int main(int argc, char** argv)
   int commSize = -1;
   MPI_Comm_size(MPI_COMM_WORLD, &commSize);
 
-  // this test should be using 4 mpi tasks, check the commSize
+  // Check for expected commSize.
+  //
+  // We expect to have 4 MPI Tasks.
+  //
+  // The commSize check must stay in sync with the NUM_MPI_TASKS
+  // value passed to the corresponding blt_add_test call in CMakeLists.txt.
   if(commSize != 4)
   {
       std::cout << "Comm Size should equal 4" << std::endl;
--- a/tests/smoke/blt_mpi_smoke.cpp
+++ b/tests/smoke/blt_mpi_smoke.cpp
@@ -25,6 +25,15 @@ int main(int argc, char** argv)
   int commSize = -1;
   MPI_Comm_size(MPI_COMM_WORLD, &commSize);
 
+  // this test should be using 4 mpi tasks, check the commSize
+  if(commSize != 4)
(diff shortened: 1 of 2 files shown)


tests/smoke/blt_mpi_smoke.cpp
```diff
@@ -25,6 +25,15 @@ int main(int argc, char** argv)
   int commSize = -1;
   MPI_Comm_size(MPI_COMM_WORLD, &commSize);
 
+  // this test should be using 4 mpi tasks, check the commSize
+  if(commSize != 4)
```
white238, line 34: Can you add a comment here and at the `blt_add_test` that says these numbers should be in sync?

white238 (approved):
Thanks @cyrush ! Merge away!

tests/smoke/blt_mpi_smoke.cpp
```diff
@@ -25,6 +25,20 @@ int main(int argc, char** argv)
   int commSize = -1;
   MPI_Comm_size(MPI_COMM_WORLD, &commSize);
 
+  // Check for expected commSize.
+  //
+  // We expect to have 4 MPI Tasks.
+  //
+  // The commSize check must stay in sync with the NUM_MPI_TASKS
+  // value passed to the corresponding blt_add_test call in CMakeLists.txt.
```
kennyweiss, line 33: Would it make sense to pass it in w/ an environment variable instead of hard-coding it?
Or is it better to be explicit?

kennyweiss (approved):

tests/smoke/blt_mpi_smoke.cpp
cyrush: Hard code is better.

In recent case that this would have helped me - the mpi launch environment was very broken -- I would not trust that it could pass an env var.

PR #755, opened 2026-02-19
Adds `BLT_ALLOW_MISSING_MPI_WRAPPER` option to allow Fortran configs with MPI to not have an MPI_Fortran_COMPILER

kennyweiss:
This PR adds the `BLT_ALLOW_MISSING_MPI_WRAPPER ` option to enable users to configure MPI without providing some MPI_<lang>_COMPILER variables for some enabled languages.

This came up in an msan config where we need Fortran to be enabled (`ENABLE_FORTRAN=TRUE`) as well as MPI (`ENABLE_MPI=TRUE`), but we don't need an MPI wrapper for Fortran -- and more importantly, we don't have an MPI wrapper for Fortran that is compatible with our C and C++ MPI wrappers.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -16,6 +16,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 - Update CUDA runtime smoketest to be compatible with CUDA 13
 - Adds `BLT_ALLOW_MISSING_MPI_WRAPPER` (default `OFF`) to allow setting up MPI without providing `MPI_<lang>_COMPILER` for some enabled languages.
   This supports niche cases (e.g. Fortran enabled without a compatible `MPI_Fortran_COMPILER`) while preserving the default behavior of requiring wrappers.
+- Checks for missing `MPI_<lang>_COMPILER` flags for each enabled language when `ENABLE_MPI` evaluates to `ON`
+  and `BLT_ALLOW_MISSING_MPI_WRAPPER` evaluates to `OFF`.
 
 ### Fixed
 - In non-mpi configurations, `blt_add_test` will now throw a `FATAL_ERROR` if the user provides `NUM_MPI_RANKS`
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -14,8 +14,8 @@ The project release numbers follow [Semantic Versioning](http://semver.org/spec/
 - Removed `-Winline` from being added to GoogleTest for Clang/Intel as it is a noop and caused a warning on Intel.
 - Toggle GoogleTest adding `-Wno-implicit-float-size-conversion` or `-Wno-sycl-implicit-float-size-conversion` based on Intel version.
 - Update CUDA runtime smoketest to be compatible with CUDA 13
-- Allows setting up MPI in configs with `C`, `CXX` and/or `Fortran` but without a corresponding `MPI_<lang>_COMPILER`. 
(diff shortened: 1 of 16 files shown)


tomstitt (approved):
thanks @kennyweiss !

rhornung67 (approved):

kennyweiss:
Fun wrinkle from the `openmp_mpi_fortran_tests` CI:
```
CMake Error at /usr/tce/backend/installations/linux-rhel8-x86_64/gcc-10.3.1/cmake-3.23.1-mdfqd2l7c33zg7xcvqizwz25vqmp7jfw/share/cmake-3.23/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find MPI (missing: MPI_C_FOUND C) (found version "3.1")
      Reason given by package: MPI component 'C' was requested, but language C is not enabled.  
```

Good test!

I was wondering if I'd need extra logic about checking for a valid C and/or C++ MPI compiler/wrapper. 
Should hopefully be an easy fix.

white238:
I chatted with @kennyweiss directly but putting summary here:

I am concerned this is changing a long standing behavior that all enabled languages needed an MPI compiler wrapper due to not listing components on the `find(mpi)` call prior to this. I think it failed before if a user didn't set all the wrappers (or possibly defaulted to a system one).

My recommendation is to add a sanity check for all wrappers being provided in the normal case (improvement over what is there today), then the niche case requiring the user to set a new variable, something like `BLT_ALLOW_MISSING_MPI_WRAPPER` (or some better name), to skip the sanity check.

Thoughts?

cyrush:
I think implicitly needing all wrappers was a mistake.

That was what led to issues with fortran flags being fused with c flags and undermining clang builds. 

Also you don't need compiler wrappers -- you need the knowledge of how to compile and link (often comes from the wrappers, but not always). 

FindMPI will work if you provide flags as well. This check for fortran wrapper my undermine that. 

It would be better if BLT had and option for which languages to use with MPI.  

That is what CMake's MPI has evolved to.

white238:
@cyrush that makes sense. We can always pivot later. shall we merge then?

Adding some documentation to the MPI section that we will use information only from the given compiler wrappers would be good though.

cyrush:
I think current fix is needed, and we can improve more in the future.

kennyweiss:
I updated the branch based on @white238's suggestion.  Specifically, by default, you need to provide a `MPI_<lang>_COMPILER` for each enabled language when `ENABLE_MPI=ON`.
Not providing these is a config error.

However, you can override that by setting `BLT_ALLOW_MISSING_MPI_WRAPPER=ON`.

Based on @cyrush's comment, we might not want to *require* users to provide a `MPI_<lang>_COMPILER` for each language.

Tag: @cyrush @white238 

**Edit:** In the meantime, I have patched blt in my local build, so this fix is not blocking me. We should get the behavior correct before merging.

ldowen:
I started to make a PR that would do something like this because the clang 19.1.3 compiler on LC systems causes all kinds of issues. Lucky me that it already exists. However, could it be extended to disable OpenMP for Fortran as well? Or should I make a separate PR for that?

ldowen:
Actually, I have another idea: what if instead of using the `MPI_<LANG>_COMPILER` to determine which MPI to bring in, what if there was a list that BLT used that by default was populated with the enabled languages but could be overwritten by only certain languages that would tell it what languages to enable MPI and OpenMP for. Seems more direct way to control this.

PR #764, opened 2026-05-29
Allow Staging Sphinx Documentation

chapman39:
* adds new arguments to blt_add_sphinx_target to set sphinx source and conf.py dirs, as well as an explicit DEPENDS argument
* adds documentation for the motivation and how to use this new feature
* tested minimal, advanced examples, also one using conf.py.in file


cmake/SetupDocs.cmake
```diff
@@ -44,53 +44,69 @@ endmacro(blt_add_doxygen_target)
...
-## blt_add_sphinx_target(sphinx_target_name)
+## blt_add_sphinx_target(sphinx_target_name
+##                       SOURCE_DIR   [source_dir]
+##                       CONF_DIR     [conf_dir]
+##                       DEPENDS      [dep1 ...])
 ##
 ## Creates a build target for invoking sphinx to generate docs
 ##------------------------------------------------------------------------------
-macro(blt_add_sphinx_target sphinx_target_name )
+macro(blt_add_sphinx_target sphinx_target_name)
```
kennyweiss: Is this left as a parameter (w/o keywords) for backwards compatibility?

docs/api/documentation.rst
```diff
@@ -30,18 +30,58 @@ blt_add_sphinx_target
...
 out of source. For ``make install``, this will place the resulting docs in
 ``docs/sphinx/<sphinx_target_name>``.
+  
+Staging Example
+-----------------
+
+A stage directory is useful when the Sphinx input tree must be assembled from
+multiple source locations, such as when one documentation project is a subset
+of another. In this case, the stage step copies shared files and local docs
+into a temporary tree, which is then passed to Sphinx as ``SOURCE_DIR``.
```
kennyweiss, line 65: It looks like staging is *enabled* by this PR via the new `DEPENDS` keyword, rather than a new feature being added. Is that right?

kennyweiss (approved):
Thanks @chapman39 

The PR description says 
```
tested minimal, advanced examples, also one using conf.py.in file
```
but I didn't see a test. 

Also, please see my comment about staging

docs/api/documentation.rst
chapman39: yes good point - theres some work that needs to be done by the user in order for staging to work. i could try to implement staging within the cmake macro, so it'd be easier to setup for users.

cmake/SetupDocs.cmake
chapman39: yes compatibility and also it'd be inconsistent with the doxygen macro counterpart. i could make a shim macro - maybe that'd be better.

docs/api/documentation.rst
kennyweiss: Your call. Since you already have it working for your use case, it might make sense to wait until there's an established need from others before generalizing.  

Maybe create an issue for it and see if anyone expresses interest?

chapman39: https://github.com/llnl/blt/issues/765

cmake/SetupDocs.cmake
chapman39: i decided to add the TARGET keyword

cmake/SetupDocs.cmake
```diff
@@ -20,88 +20,118 @@ endif()
 
 
 ##------------------------------------------------------------------------------
-## blt_add_doxygen_target(doxygen_target_name)
+## blt_add_doxygen_target(TARGET [doxygen_target_name])
```
kennyweiss: Is this a breaking change? If so, please document in the RELEASE-NOTES so people know how to upgrade (or why upgrading broke their documentation).

cmake/SetupDocs.cmake
```diff
@@ -20,88 +20,118 @@ endif()
...
     install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/html" 
-            DESTINATION docs/doxygen/${doxygen_target_name} OPTIONAL)
+            DESTINATION docs/doxygen/${DOXYGEN_TARGET} OPTIONAL)
 
 endmacro(blt_add_doxygen_target)
 
 
 ##------------------------------------------------------------------------------
-## blt_add_sphinx_target(sphinx_target_name)
+## blt_add_sphinx_target(TARGET       [sphinx_target_name]
```
kennyweiss: Similarly for the TARGET here.  If this is a breaking change, please document it in the RELEASE-NOTES so people know how to upgrade (or why upgrading broke their documentation).

kennyweiss: Also:
Is the target name optional? It looks like it is required, so perhaps this should lose the square brackets?

```suggestion
## blt_add_sphinx_target(TARGET       sphinx_target_name
```

cmake/SetupDocs.cmake
chapman39: yes. thank you for pointing this out

cmake/SetupDocs.cmake
chapman39: good point. i didn't realize that was the pattern. ill change that and also check if target isn't set

chapman39:
thank you for the feedback @kennyweiss .

about testing, i meant i tested the cmake macro locally for the 3 use-cases mentioned. if we want documentation testing in blt let me know. we might need to modify the host configs to include SPHINX_DIR, or something. or i can create an issue on it and we can come back to it.

kennyweiss:
I think it's in good shape after your latest changes (i.e. highlighting in the code/user docs and RELEASE_NOTES that `TARGET` is now required)

cmake/SetupDocs.cmake
```diff
@@ -20,88 +20,128 @@ endif()
...
+## blt_add_doxygen_target(TARGET doxygen_target_name)
 ##
 ## Creates a build target for invoking doxygen to generate docs
 ##------------------------------------------------------------------------------
-macro(blt_add_doxygen_target doxygen_target_name)
+macro(blt_add_doxygen_target)
+    set(_doxygen_options)
+    set(_doxygen_single_value_args TARGET)
+    set(_doxygen_multi_value_args)
+    cmake_parse_arguments(DOXYGEN
```
white238, line 31: To be consistent with the rest of BLT, this should have been:

```
    cmake_parse_arguments(arg
```

Like:

https://github.com/llnl/blt/blob/899963297b270407774a6b907091250c0c7e8b7d/cmake/BLTMacros.cmake#L20-L26

This makes it so all the arguments stand out, for example `DOXYGEN_TARGET` would have been `arg_TARGET`.

This isnt user facing so is easy to fix in a follow up PR. Would you mind doing one?

cmake/SetupDocs.cmake
```diff
@@ -20,88 +20,128 @@ endif()
...
+                          "${_doxygen_single_value_args}"
+                          "${_doxygen_multi_value_args}"
+                          ${ARGN})
+
+    if(DOXYGEN_UNPARSED_ARGUMENTS)
+        message(FATAL_ERROR
+            "blt_add_doxygen_target received unexpected arguments: ${DOXYGEN_UNPARSED_ARGUMENTS}")
+    endif()
+
+    if(NOT DOXYGEN_TARGET)
```
white238, line 42: The specific check should probably go first as well.

docs/api/documentation.rst
```diff
@@ -30,18 +30,58 @@ blt_add_sphinx_target
...
+
+Let's say project A depends on project B.
+
+.. code-block:: cmake
+
+    set(PROJECT_A_STAGE_DIR "${CMAKE_CURRENT_BINARY_DIR}/_stage")
+    set(PROJECT_A_SOURCE_DIR "${PROJECT_A_STAGE_DIR}/src/docs")
```
white238, line 72: I think you should set `PROJECT_B_SOURCE_DIR` here as well and just call it `SPHINX_STAGE_DIR` not `PROJECT_A_STAGE_DIR`

chapman39: https://github.com/llnl/blt/pull/766

Issue #697, opened 2024-09-05
fortran vs mpi logic


cyrush:
Use case:

Fortran language support is on - but MPI Fortran support is not needed. 

I think because BLT calls `find_package(MPI REQUIRED)` we will always hit an error:

```
  49    -- Found MPI_C: /lib64/libxpmem.so (found version "3.1")
     50    -- Found MPI_CXX: /lib64/libxpmem.so (found version "3.1")
     51    -- Could NOT find MPI_Fortran (missing: MPI_Fortran_WORKS)
  >> 52    CMake Error at /usr/tce/backend/installations/linux-rhel8-x86_64/gcc
           -8.3.1/cmake-3.24.2-ywx52e32uh6gkxzuyubpwkulzgdvxyh6/share/cmake-3.2
           4/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
     53      Could NOT find MPI (missing: MPI_Fortran_FOUND) (found version "3.
           1")
```

For this case, the following would work:
```
find_package(MPI COMPONENTS C)
```

But we need to find the right way to wire as a blt option.

Issue #355, opened 2020-04-16
Linking Fortran MPI to C++ project


greene30:
I have a project that is written in just C++ (I set the cmake LANGUAGE to CXX) that uses BLT. I recently added a third party library that has Fortran MPI calls. When I tried to link my code, I was getting undefined references to Fortran MPI. I was able to fix this by adding Fortran to the LANGUAGE list. This solution may be confusing to future developers on my team since we have no Fortran code in our source. Could you add a BLT function or variable to make it clearer that we're only adding the Fortran MPI libraries?

PR #771, opened 2026-07-16
Update BLT to apply MPI flags for the respective compilers to only the respective compiler lines

aaroncblack:
I encountered issues building a C++/Fortran code using a mixed vendor compiler clang/gfortran set up.  The BLT mpi target added the C++ mpi paths to the general compiler flags, resulting in the gfortran compiles failing.

Description of changes:
Update BLT to apply MPI flags for the respective compilers to only the respective compiler lines.

Add test that builds a mixed language executable and confirms the respective MPI flags and include paths for C, C++, and Fortran are only added to language compile lines.


aaroncblack:
I made heavy use of codex on this, so take it with a grain of salt.

cmake/thirdparty/BLTSetupMPI.cmake
```diff
@@ -30,6 +30,31 @@ set(_mpi_compile_flags )
 set(_mpi_includes )
 set(_mpi_libraries )
 set(_mpi_link_flags )
+set(_mpi_fortran_includes )
```
aaroncblack, line 30: @white238 I told the agent that our C++ codes make heavy use of the C MPI API, so to focus on splitting out the C/C++ vs the Fortran paths and flags.

It said in many MPI installs the MPI_C_INCLUDE_DIRS and MPI_CXX_INCLUDE_DIRS are identical, but if its a concern, another option is to have the C++ get both C and CXX mpi flags and paths, de-duplicated.

cyrush:
there are actually several issues related to this

https://github.com/llnl/blt/issues/697

https://github.com/llnl/blt/issues/355

https://github.com/llnl/axom/issues/1853

we have discussed solutions, I dont think we have really captured those discussions well.

What I recall:
Ideally we move to use the bona fide cmake mpi per language targets, as well as we avoid having to even search for mpi fortran in cases where it is not needed (I realize that is not your case)

aaroncblack:
> What I recall: Ideally we move to use the bona fide cmake mpi per language targets, as well as we avoid having to even search for mpi fortran in cases where it is not needed (I realize that is not your case)

Yeah, I'd agree moving to the cmake mpi per language targets is a good idea.  I'll try updating this PR to make use of those.

